PR #59, opened 2026-06-06
fix(docker): add restart: unless-stopped so stacks survive crashes/reboots

JBailes:
Deeper install dig: the Docker compose stacks had **no `restart:` policy**, yet README (line 485) and MANUAL (1286/1359) state the containers are *"supervised by Docker's restart policy."* So a container crash or a **host reboot left the whole stack down** until a manual `docker compose up` — contrary to the docs.

Add `restart: unless-stopped` to every long-running service — `postgres`, `embedder`, `aimee-kb`, `aimee-server`, `aimee-server-kb`, `llm` — across `compose.combined.yaml`, `compose.server.yaml`, `compose.yaml`, and `compose.server-standalone.yaml`. The recommended Docker deployment now auto-recovers from crashes and reboots, as documented.

Verified the policy is applied (`docker inspect` → `RestartPolicy.Name=unless-stopped`); restart-on-unexpected-exit / on-daemon-restart is standard Docker behavior. Pairs with the server pid-file restart fix (#57) so a restarted server also starts cleanly.


docs/WHATS_NEW.md
```diff
@@ -1,31 +1,235 @@
...
+- Watched-proposal triggers can create runs. Trigger mode is recorded, but current Go scheduling does
+  not change between `autonomous` and `interactive`.
+- A human gate always parks. Autonomous mode cannot approve one.
+- Roundtable findings feed the next author pass instead of disappearing at a boolean verdict.
+- A complete pre-supplied proposal can advance without being rewritten; a failed attempt with no
+  proposal still fails.
+- Agent admission limits apply globally and per workflow. Saturated agents are routed around.
+- A merge conflict, missing commit, lost replay, or exhausted gate returns a named terminal or
+  parked state instead of silently advancing.
+- The build workflow's final PR now targets `trunk`.
```
gooTan: **[chairman · nit]** Three of nine seats flagged the ASCII apostrophe in "workflow's" as a mismatch against the acceptance criterion's typographic "workflow’s". Worktree check: docs/WHATS_NEW.md contains 0 typographic apostrophes and 10 ASCII apostrophes (e.g. line 1 '# What's new in 0.3.0', line 91 'a roundtable's findings'), so the added bullet follows the file's established convention; the curly quote in the packet is a text-entry artifact, and the bullet's wording is otherwise verbatim. Not a defect.

Suggestion: No change needed. If a byte-exact typographic apostrophe is genuinely intended, that should be stated as an explicit requirement — and it would be inconsistent with the rest of the file.

PR #123, opened 2026-06-08
docs(proposal): agent roundtable — round-robin collaborative drafting and review

JBailes:
## Summary

A **docs + tooling** PR (no product C changes). It does three things:

1. **Adds the Agent Roundtable proposal** — generalize the single-shot
   Mixture-of-Agents ensemble (`src/server/delegate_ensemble.c`,
   `aimee delegate aggregate`) into a bounded multi-round roundtable (draft /
   review modes, parallel / sequential turns, deterministic convergence,
   keep-best, preflight cost limits). The proposal first repairs the shipped
   engine, whose defects are verified in-tree:
   - **Unreachable (§0.2):** no shipped binary calls `delegate_ensemble_run` —
     its only non-test caller, `cmd_agent_delegate.c`, is lint-only — so
     `aimee delegate aggregate` runs an ordinary single-agent delegate, not the
     ensemble. Still true on `testing` @ #128.
   - **Mis-routed (§0.1):** every fan-out task is built `role = NULL`, so all N
     "reference models" run the one default agent; `agent_task_t.temperature` is
     dead on the parallel path; a shared-`agent_t` mutation races under
     `agent_run_parallel`; `srand(time(NULL))` is reseeded per call.
   - P0 (wire the entry point + fix routing) is **standalone-shippable**, but per
     review is being carried as a **separate follow-up PR**, not bundled here.

2. **Builds the §0.4 index-honesty control as a real gate.**
   `scripts/check-proposal-links.py` (wired into `make lint`) asserts every
   `proposals/*.md` link resolves; a delisted proposal must be plain text, not a
   dead link. It immediately surfaced **63 dead links** in `PROPOSALS.md` (the
   rot that hid the missing MoA done-file) — all delisted in this PR so the gate
   is green.

3. **Refreshes two stale proposals to match shipped reality** (`testing` @ #128):
   - `optimization-surface.md`: the reward loop is now **closed** in production
     (`kb_service_memory.c:98`, `kb.c:1760`), the phantom bandit export is fixed
     (data-driven), and `optimize points|baseline|replay|run|compare|promote`
     shipped (#117, #125–#128). Status banner added; only the
     `briefing_style`/`guardrail_strictness` tail remains.
   - `v1-dispatch-migration-finish.md`: the `cli_v1_rpc_local`→`cli_v1_dispatch_local`
     rename (#122) and the kb-direct surfacing gate (Part B) shipped; only the
     A.2 comment purge remains.

The branch is merged up to current `testing` (#128).

[...]

PR #131, opened 2026-06-09
feat(delegate): make the MoA ensemble actually work (P0 — routing + entry point)

JBailes:
## Summary

Makes the **Mixture-of-Agents ensemble actually work**. This is the P0 fix from the
agent-roundtable proposal (PR #123 §0.1/§0.2), carried as its own PR per review.

Before this, `aimee delegate aggregate "<prompt>"` **never ran the ensemble**:
`delegate_ensemble_run` had no caller in any shipped binary, and even its engine
routed every fan-out task to the one default agent. Two commits fix both layers.

### Commit 1 — engine (routes to distinct agents + temperature + srand)
- **Unrouted references:** every fan-out task was built `role = NULL`, so all N
  "reference models" ran the single default agent. Added a per-task `agent`
  selector to `agent_task_t`; `parallel_worker` resolves it by name (like
  `aux_router`) via a new `agent_run_named` helper, and the engine now sets
  `tasks[i].agent = ensemble_reference_models[i]`. The references are real,
  distinct participants for the first time.
- **Dead per-task temperature:** `agent_run` had no temperature param and
  `parallel_worker` dropped `task->temperature`. Added `agent_run_ex` (carries
  temperature); `agent_run` is now a thin `0.3` wrapper so its ~28 call sites are
  byte-unchanged. This also makes `agent_coord`'s `0.3 + 0.1*i` vote perturbation
  take effect (it was a no-op for the same reason).
- **Shared-`agent_t` race:** the named path **clones** the selected `agent_t`
  before mutating runtime config/ablation/write_enforce, so concurrent workers
  never write the shared `cfg`-owned struct. Rate-limit/health accounting stays on
  the process-wide `provider_catalog_*` tables (keyed by name).
- **`srand` reseed:** `shuffle_indices` now seeds a thread-local `rand_r` from the
  monotonic clock instead of `srand(time(NULL))` per call.
- **Un-stubbed the routing blind spot:** the test stub ignored `tasks` entirely
  (why the bug shipped unseen); it now captures the per-task selectors and a new
  test asserts three references → three distinct, ordered, non-NULL agents.

### Commit 2 — entry point (`POST /v1/delegate/aggregate`)
- New dispatch method `delegate.aggregate` → `handle_delegate_aggregate`, exposed
  as `POST /v1/delegate/aggregate` via `rh_dispatch_op_async` — the sanctioned
  async pattern for every LLM-heavy op (`kb.build`, `eval.run`, `memory.benchmark`):
  returns a queued run handle and finalizes through `/v1/runs/{id}`, so the
  buffered listener never runs the fan-out inline. Explicit `CAP_DELEGATE`.
- Thin client: a specific `delegate aggregate` route + marshaller mapping
  `positional[0] → prompt` (not role); forward auto-posts + polls the run.

### Design note (correcting the proposal)
The proposal mandated a *native `op == NULL`* handler and "no `delegate.*` dispatch
method," claiming `rh_dispatch_op_async` "isn't reusable." Reading the code, it **is**
reusable and is the codebase-wide async convention; a dispatch method that owns a
first-class `/v1` route is fully coverage-clean (**0 excluded**) and is not a
`/v1/rpc` fallback. So this uses the dispatch-backed async path (the proposal's
Option A) — lower risk, consistent with `eval.run` et al. Proposal §0.2 should be
updated to match (follow-up to PR #123).

### Scope deferred (proposal P1+, not this PR)
The multi-round roundtable, draft/review modes, heap-grown artifact, and the
generic role-path unconditional clone (the residual benign `agent_vote` race) are
strictly additive and out of scope here.

## Verification
- `make` server + thin client link clean (`-Werror`).
- `unit-test-delegate-ensemble` passes incl. the new routing test.
- Green: `check-cli-v1-routes`, `check-v1-method-coverage` (0 excluded),
  `server-api-conformance`, `api-conformance`, `sdk-parity`, `dispatch-caps`.

### Note: pre-existing unrelated `docs-gen-check` drift
`docs/gen/api-v1.md` is stale on `testing` (the optimize/bandit PRs #125–#128 added
endpoints to `api/openapi-v1.yaml` without regenerating it). That gate is red
independent of this PR — left untouched to keep this change focused on MoA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -578,6 +578,7 @@ set(DATA_SRCS
 
 set(AGENT_SRCS
     ${AIMEE_SRC_DIR}/server/agent_runtime.c
+    ${AIMEE_SRC_DIR}/server/agent_parallel.c
     ${AIMEE_SRC_DIR}/server/agent_runtime_liveness.c
     ${AIMEE_SRC_DIR}/server/agent_fallback.c
     ${AIMEE_SRC_DIR}/server/agent_context_budget.c
@@ -744,6 +745,7 @@ set(CLI_SRCS
     ${AIMEE_SRC_DIR}/client_integrations.c
     ${AIMEE_SRC_DIR}/cmd_profile.c
     ${AIMEE_SRC_DIR}/cmd_manuscript.c
+    ${AIMEE_SRC_DIR}/cmd_optimize.c
     ${AIMEE_SRC_DIR}/manuscript.c
     ${AIMEE_SRC_DIR}/cli_persona.c
     ${AIMEE_SRC_DIR}/cli_roles.c
@@ -1127,6 +1129,13 @@ add_executable(aimee
 target_include_directories(aimee PRIVATE ${AIMEE_SRC_DIR} ${AIMEE_SRC_DIR}/headers ${AIMEE_SRC_DIR}/vendor/headers ${AIMEE_SRC_DIR}/db1 ${AIMEE_SRC_DIR}/db2)
 target_compile_options(aimee PRIVATE ${AIMEE_COMMON_WARNINGS})
 target_link_libraries(aimee PRIVATE Threads::Threads)
+if(NOT WIN32)
+    # cmd_optimize.c (bandit math) and cli_attention_guard.c use sqrt/pow/log;
+    # libm is implicit in libSystem on Apple but a separate link on Linux. The
+    # make build links -lm unconditionally (L_CLIENT); mirror that here. Windows
+    # math lives in the CRT, so no separate lib.
+    target_link_libraries(aimee PRIVATE m)
+endif()
 if(WIN32)
     target_compile_definitions(aimee PRIVATE AIMEE_WINDOWS _WIN32_WINNT=0x0A00)
     target_link_libraries(aimee PRIVATE ws2_32)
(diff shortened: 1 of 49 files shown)


JBailes:
**Live reachability validated** (scratch `aimee-server`, local UDS, no providers needed):

```
POST /v1/delegate/aggregate {"prompt":"Explain optimistic vs pessimistic locking tradeoffs."}
  -> {"id":"oprun_...","method":"delegate.aggregate","status":"queued"}   # async handle; listener not blocked

GET /v1/runs/oprun_...
  -> {"status":"completed","result":{"status":"error",
       "message":"Mixture-of-Agents ensemble disabled (set ensemble.enabled=true)"}}

aimee delegate aggregate "Explain optimistic vs pessimistic locking tradeoffs."
  -> aimee: Mixture-of-Agents ensemble disabled (set ensemble.enabled=true)
```

That `delegate.aggregate`-specific message is emitted only by the new
`handle_delegate_aggregate`, confirming end-to-end: route -> method -> async
dispatch via `rh_dispatch_op_async` -> `/v1/runs/{id}` -> handler. The thin-client
run also proves `delegate aggregate` maps `positional[0] -> prompt` (the prompt
cleared the >=20-char guard and reached the `ensemble_enabled` check) rather than
the old catch-all `role = aggregate`. The disabled-by-default branch is the
expected terminal state without an enabled ensemble; per-task agent routing past
that point is covered by the un-stubbed `test_ensemble_routes_to_distinct_agents`
unit test.

PR #159, opened 2026-06-09
Finish recall economy proposal

JBailes:
## Summary
- move recall economy progressive disclosure proposal from pending to done
- mark the proposal state as done
- update the proposal index count and retrieval entry

## Tests
- docs-only change; verified proposal file moved locally


JBailes:
Closing: this was only a proposal-state/docs move, but the request is to implement the proposal. Replacing with an implementation PR.

PR #177, opened 2026-06-10
curator: deep-ify all four vector tables (entity/claim/narrative/code-unit)

JBailes:
Extends the 4B deep tier to **all four curator vector tables**, wiring each as an
active, deep-confirmed consumer. Builds on the memory deep tier (#174) + halfvec
unification (#175). Everything is **gated by `memory_deep_embedding_enabled`
(default off)**, reuses the 4B `/embed_deep` command, and degrades gracefully on
partial coverage. Deep is always a precision guard — additive or merge-preventing,
never destructive.

### Entity resolution — merge guard
`embedding_deep` + `resolve_entities`: a confident live merge (>= 0.85) is rejected
when the 4B deep space disagrees (< 0.60), so the live embedding can't collapse
two distinct canonical entities. Deep can only prevent a merge, never force one.

### Claim contradictions — fuzzy, deep-confirmed mining (new capability)
`subj_attr_deep_vec` + a fuzzy phase in `detect_contradictions`: finds **paraphrased**
contradictions (semantically-same subject+attribute, different value) the exact
self-join misses, and **requires** a 4B-deep confirmation (>= 0.75) before linking.

### Narrative + code-unit — deep-confirmed similarity links (new capability)
`embedding_deep` / `body_deep_vec` + index-time linking: near-duplicate narratives
and code clones get an **additive `similar_to` edge** — search live (>= 0.85),
require deep confirmation (>= 0.85), never drop content.

### Safety
Deep confirmation mandatory everywhere; conservative thresholds; gated default-off;
additive/merge-preventing only (no destructive merges or drops); idempotent links;
bounded batches. The contradiction/similarity thresholds are deliberate guesses —
tune on a real corpus before enabling. All curator unit tests pass (guards no-op
with deep off); real-pgvector e2e covers the deep-on paths.

### DIFF
--- a/deploy/migrations/2026-embed-deep-halfvec.sql
+++ b/deploy/migrations/2026-embed-deep-halfvec.sql
@@ -47,4 +47,10 @@ ALTER TABLE memory_embeddings ADD COLUMN IF NOT EXISTS embedding_deep halfvec(25
 CREATE INDEX IF NOT EXISTS idx_memory_embeddings_deep_hnsw
     ON memory_embeddings USING hnsw (embedding_deep halfvec_cosine_ops);
 
+-- Curator entity resolution shares the deep tier: the same precision column on
+-- curator_entity_vectors (written/compared by resolve_entities).
+ALTER TABLE curator_entity_vectors ADD COLUMN IF NOT EXISTS embedding_deep halfvec(2560);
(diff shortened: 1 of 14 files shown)


JBailes:
Superseded: moving to a single-embedder model (0.6B or 4B per deployment) instead of the two-tier deep architecture.

PR #188, opened 2026-06-11
Run local-CLI agents (Claude) on the thin client, with streaming

JBailes:
## Problem

A "Claude agent" in aimee is a **local-CLI-exec** agent: `aimee-server` fork/execs the `claude` binary (`claude -p --output-format stream-json …`) and reads its output (`server/provider_cli_adapter.c` → `claude_spawn`). The same holds for the other provider-CLI / tmux-cli backends. That exec runs **server-side** — so on a remote / containerized `aimee-server` (e.g. the NAS combined plugin), which has **no `claude` binary, no tmux, and no Claude login**, a Claude primary or delegate fails (`exec(claude)` → 127). Those things live on the user's **thin-client** machine, where the working tree also already is.

This is distinct from the HTTP providers (minimax/mistral/mimo/codex), where only the *credential* had to move to the client — here the *execution itself* must move.

## Approach

Reuse the existing detached-workspace runner reverse-channel (`/v1/runner/poll` + `/v1/runner/respond`), which is already serviced by a background thread during a chat turn. Add **one generic client capability** — *spawn a command, stream stdout back* — and keep all CLI-specific logic on the server, so the client stays generic and the same mechanism serves every local-CLI provider.

- **`exec_stream`** workspace-provider op: the `shared` provider runs argv+stdin locally with streamed stdout (this is what the **client** uses to run `claude`); the `detached` provider marshals it over the reverse channel.
- The runner queue's single response slot becomes a **FIFO + streaming transport**, so the client posts incremental `{partial,chunk}` responses then a final `{rc}`. Single-flight per workspace is preserved; `/v1/runner/respond` passes through unchanged.
- `provider_cli` routes to `exec_stream` when `workspace_provider_active()` is `detached` and the adapter exposes `build_argv` (claude). `claude_spawn`'s argv construction is factored into a shared `build_argv` hook.
- A **per-turn thread-local sink** (`cli_stream_sink`) carries each parsed stream-json text delta into the chat SSE, so a remote Claude turn **streams token-by-token**; the chat worker installs it and skips the buffered final emit when deltas streamed.

**No Claude credential is ever sent to or stored on the server** (claude uses the client's own `~/.claude`). Co-located deployments are unchanged (local fork/exec). Works for the primary chat turn and `aimee delegate … --via claude`.

## Tests

- `exec_stream` round-trip over the runner queue: multi-chunk reassembly, stdin delivery (claude reads its prompt on stdin), and non-zero exit propagation.
- `provider_cli` detached routing: claude argv marshalled (with `stream-json`), prompt fed on stdin, stream-json deltas reassembled into the result.
- `make unit-tests` + `make lint` green (clang-format-19, all conformance/coverage/cli-v1 gates ok).

## Not yet validated

Live end-to-end on the NAS thin client (no `claude`/tmux on the server) — to be run next.


JBailes:
**Live e2e validated on the NAS thin-client deployment** (combined image `0.0.7-cliexec`, `remote_writes: full`; the server container has no `claude` and no tmux).

- `aimee config set provider claude` + `aimee chat` in the `~/gow` **detached** workspace returned a coherent answer — claude ran **on the client** over the reverse channel, not on the server.
- Tree-access proof: asked claude to `ls` the cwd; it returned the client's **actual** `~/gow` entries (`GamesOnWhales.Wolf.Net.Bindings, LXC2Docker, Sunshine, …, wolfmanager`), confirming it executed against this client's real working tree with the client's own `~/.claude` login. No Claude credential reached the server.
- Output streamed back through the server into the chat turn (delta→SSE sink).
- Pre-fix, the same turn failed with `claude exited 126: chdir(<client path>) failed` (server had no such path / no claude) — exactly the gap this PR closes.

All 11 CI checks green.

PR #225, opened 2026-06-13
Default delegate tools_enabled from the backing model's capability

JBailes:
## Problem

Delegate tasks that require tool use (e.g. any `review`-role delegate) fail with:

```
no configured model supports required capabilities (caps=tools, min_context=…)
```

even though every configured delegate is backed by a tool-capable model
(mistral, minimax, openai, anthropic).

## Root cause

`tools_enabled` is **both** a capability signal and an execution policy, and the
two were conflated:

1. `agent_config.c` parsed an **absent** `tools_enabled` key as a blanket `0`.
2. The routing capability filter (`delegate_filter_route_capabilities`,
   `server/delegate_routing.c`) treats that per-agent flag as authoritative for
   the tools bit — when it is off it **clears** the model's intrinsic
   `MODEL_CAP_TOOLS` (`effective_flags &= ~MODEL_CAP_TOOLS`).

So a perfectly tool-capable model looked tool-*incapable* to the filter purely
because nobody set the opt-in flag, and tool-requiring roles found **zero**
candidates. (`min_context` in the error is the job's demand — the prompt token
count — not a model limit; the unmet capability was `tools`.)

## Fix

When the `tools_enabled` key is **absent**, derive its default from the backing
model's intrinsic tool capability (`model_capability_get`) instead of a blanket
`0`:

- tool-capable model + no key → tools **on** (routable to, and executed with,
  tools for tool roles);
- model with no known tool capability → stays **off**;
- explicit `"tools_enabled": true|false` → honoured verbatim (operator wins).

This makes capability and policy consistent in one place at config load, so both
routing **and** execution agree.

## Test-link wiring

`agent_load_config` now references `model_capability_get`, so unit-test binaries
that link the real `agent_config.o` must also link `model_registry.o` +
`models_dev.o` + `models_dev_cache.o` (previously gc-sections-dropped because the
symbol was unreachable). Added to the four affected rules: `cmd-doctor`,
`cmd-onboard`, `server-http`, `agent-http`.

## Tests

- New `test_tools_enabled_capability_default` (in `unit-test-agent`): a
  tool-capable model with no `tools_enabled` key loads `tools_enabled == 1`;
  explicit `false` stays `0`; explicit `true` stays `1`.
- Full `unit-tests` build links green tree-wide (0 undefined/multiple-definition);
  the four touched binaries build and pass; `clang-format-19` clean.

### DIFF
--- a/src/server/agent_config.c
+++ b/src/server/agent_config.c
@@ -508,7 +508,30 @@ int agent_load_config(agent_config_t *cfg)
          ag->enabled = (!v || !cJSON_IsBool(v)) ? 1 : cJSON_IsTrue(v);
 
          v = cJSON_GetObjectItem(a, "tools_enabled");
-         ag->tools_enabled = (v && cJSON_IsBool(v)) ? cJSON_IsTrue(v) : 0;
+         if (v && cJSON_IsBool(v))
+         {
+            /* Explicit operator setting always wins (force on or off). */
+            ag->tools_enabled = cJSON_IsTrue(v);
+         }
+         else
+         {
+            /* Absent: derive the default from the backing model's intrinsic
+             * tool capability. `tools_enabled` is both a capability signal and
+             * an execution policy; defaulting it to a blanket 0 made every
+             * delegate that omits the key look tool-INCAPABLE to the routing
+             * filter (delegate_filter_route_capabilities), so tool-requiring
+             * roles (e.g. `review`) found zero candidates even though the
+             * underlying model (mistral / minimax / openai / anthropic / …)
+             * fully supports tool calls. Deriving from the model's real
+             * capability makes a tool-capable delegate usable for tool roles by
+             * default; a model with no known tool capability still defaults off;
+             * and an explicit "tools_enabled": false above still opts out. */
+            model_capability_t mc;
+            ag->tools_enabled =
+                (model_capability_get(ag->provider, ag->model, &mc) && (mc.flags & MODEL_CAP_TOOLS))
+                    ? 1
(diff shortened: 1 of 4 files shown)


JBailes:
### Delegate roundtable review — converged (R2 clean)

Ran the change through an aimee delegate roundtable (reviewer / architect / security seats, via mistral + mimo-2.5).

**R1** surfaced three concerns, all addressed in `d98909b`:
1. *Offline-safety / startup* — guarded the lookup on a non-empty model and documented that `model_capability_get` is total + offline (on-disk caches + a pure in-code heuristic, never the network) and returns 0 for unknown/empty models, so it cannot fail or block startup and fails **safe** to tools-off.
2. *Capability-vs-policy conflation* — made the invariant explicit: an explicit `"tools_enabled"` is honoured verbatim; only the **unspecified** default changes, so no operator intent is overridden.
3. *Test-link coupling* — the `model_registry.o`/`models_dev*.o` additions reflect a genuine `agent_config → model_registry` dependency (previously gc-sections-dropped) and follow the repo's per-binary link convention.

**R2** (re-review of the revised diff): **reviewer CLEAN, architect CLEAN, security CLEAN** — no blocking issues. Security explicitly noted the default-tools-ON surface is acceptably gated (explicit `false` opts out; unknown models fail safe to off) with no privilege escalation.

Verification: full `unit-tests` build links green tree-wide (0 undefined/multiple-definition); the four touched test binaries build and pass; new `test_tools_enabled_capability_default` passes; `clang-format-19` clean.

PR #232, opened 2026-06-13
WP-C.2c(3): chat-initiated delegate reaches the webchat user's vault

JBailes:
The final WP-C wiring. A webchat user's chat can have the model call the `delegate` tool mid-turn; that delegate must resolve the user's `webuser:` vaulted credential. But the server **agent loop is decoupled from the originating connection** (it runs on a pooled worker and carries per-request context via thread-locals like `agent_set_request_session`), so the chat-spawned delegate's `create_compute_ctx` saw **no** principal.

## Fix (mirrors the established per-request thread-local pattern)
- **`agent_set_request_vault_principal` / `_get`** (thread-local) — set by the chat worker from its `compute_ctx`'s principal **around** the agent loop and **cleared immediately after**, so it's bounded to that turn's execution and a later delegate on the same pooled thread can never inherit a prior turn's principal (cross-user-leak guard).
- **`create_compute_ctx`** falls back to the thread-local when its conn carries no principal; the **conn principal still wins** when present (the direct `/v1` path). The principal originates only from the attested conn — never a client value.

## Tests
`create_compute_ctx` falls back to the thread-local when the conn has no principal; the conn principal wins over the thread-local. `server_compute.c` kept at the 2000-line cap. Full server links; server-compute / dispatch / vault suites pass.

This closes the last WP-C.2 gap — webchat users now get their vaulted credentials both via the management UI **and** auto-injected into chat-initiated delegates. Reviewed-before-merge per protocol.

### DIFF
--- a/src/headers/agent_config.h
+++ b/src/headers/agent_config.h
@@ -47,6 +47,15 @@ void agent_set_request_codex_creds(const char *token, const char *account_id);
  * Thread-local. */
 void agent_set_request_session(const char *session_id);
 
+/* The attested vault principal (WP-C) for the in-flight chat turn, thread-local.
+ * The chat worker sets it from its compute_ctx around the agent loop and clears
+ * it after, so a delegate the chat spawns in-process (decoupled from the
+ * originating connection) can still reach the same user's vault:
+ * create_compute_ctx falls back to this when its conn carries no principal. Set
+ * strictly for the agent-loop duration; empty otherwise. NEVER a client value. */
+void agent_set_request_vault_principal(const char *principal);
+const char *agent_get_request_vault_principal(void);
+
 /* Snapshot of the per-turn, thread-local credential context (session id + Codex
  * creds). agent_set_request_session / agent_set_request_codex_creds bind these
  * on the dispatching thread, but a parallel fan-out (agent_run_parallel) runs
--- a/src/headers/agent_config.h
+++ b/src/headers/agent_config.h
@@ -3,6 +3,7 @@
 
 #include "agent_types.h"
 #include "config.h"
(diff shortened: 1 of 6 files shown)


JBailes:
## Pre-merge roundtable review → fixed in 4ffda56

Ran the delegate roundtable (reviewer / security / architect personas) over the diff; all three converged on **block**. I verified each load-bearing claim against the code before acting:

**Fixed (real correctness gap):** the per-turn vault principal was added as a thread-local but **not** added to `agent_request_creds_t`. `agent_run_parallel` (the ensemble/coord fan-out) restores creds on a fresh worker thread that doesn't inherit thread-locals — so a parallel/ensemble delegate silently couldn't reach the user's vault, while a same-thread delegate could. Added `vault_principal` to the snapshot struct + snapshot/restore, mirroring session id + codex creds. Added a snapshot/restore round-trip test.

**Verified as a non-issue (panel overstated to blocker):** "`attested_transport` left zero on the thread-local fallback." The only consumer of `cctx->vault_principal` in the compute path is `vault_service_inject_api_key` → `vault_service_get`, which is **principal-only** (no transport). The transport+principal pairing is used only at the separate `server_vault.c` unlock RPC, not here. Real code-smell, not a current bug — left as-is.

**Verified as already-safe (panel concern doesn't materialize):** the per-turn clear and the leak guard. `chat_stream_worker` has no longjmp/pthread_exit/cancellation in its tree, so the post-call clear always runs; and the set is unconditional at the top of every turn. Clarified the `create_compute_ctx` comment — the fallback is the chat-spawned delegate's **loopback conn** (no restored identity on the pooled agent-loop thread), not a NULL conn — so the "anonymous request picks up a stale principal" path isn't reachable.

`verify-local: ok` (full suite, incl. unit-test-agent + unit-test-server-compute).

PR #244, opened 2026-06-13
Server-sealed dual-access credential vault (WP-C.4)

JBailes:
## Why

Delegate/provider credentials must stay on aimee-server, encrypted at rest, and be usable by delegate runs autonomously — without a client re-unlocking after every restart. The existing per-user vault (uid:/webuser:) already stores credentials encrypted at rest, but it is **zero-knowledge**: the KEK is derived per-session from a client root key / login password and cached RAM-only with a 15-min TTL. After a restart or TTL expiry the server hits `VAULT_ERR_LOCKED` and **cannot decrypt** a delegate's key — so it can't drive delegates on its own ("otherwise we can't handle them").

## What

Adds a **dual-access wrap** (WP-C.4): every vaulted credential's per-credential DEK is wrapped a second time under a **server-sealed KEK**, so the server can decrypt autonomously while the user's zero-knowledge path is unchanged.

- **`vault_server_key`** — a 32-byte master key minted on first use, persisted `0600` at `<aimee_home>/.vault/.server-master.key`; the server KEK is HKDF-derived from it (domain-separated from the user KEKs) and cached process-wide. Fail-closed on any entropy/IO/derivation error.
- **`vault_store`** — `vault_store_set_dual` writes a second `wrapped_dek_server`; `vault_store_get_server` decrypts with the server KEK **alone** (no user KEK, no unlock, survives restart); `vault_store_add_server_wraps` backfills the server wrap for pre-existing user-only entries.
- **`vault_service`** — `set` dual-wraps every push (fail-closed if the server KEK is unavailable — never store a credential the server can't read); `inject_api_key` is **server-first**, falling back to the user-KEK path only for legacy creds; `unlock` backfills server wraps now that the user KEK is in hand.

The secret stays AES-256-GCM encrypted at rest under a random per-credential DEK; the server KEK only unwraps the DEK. The delegate run-path (`delegate_resolve_credentials` → `vault_service_inject_api_key`) now resolves credentials after a restart with no unlock.

## Trust boundary (deliberate)

The master key lives beside the vault on the data volume, so this defeats backup/disk theft and plaintext-in-config — **not** a root-level host compromise (root can read both). Same boundary aimee-server already runs under. Structured so an operator passphrase can wrap the master key file later with no data migration.

## Tests

- `test_vault_store`: dual-wrap server get, wrong-server-key fail-closed, legacy→backfill, at-rest is ciphertext-only (incl. `wrapped_dek_server`).
- `test_vault_service`: **autonomous inject after a simulated restart** (user vault locked) + missing-cred NO_ENTRY leaves api_key untouched.
- `test_vault_server_key`: master-key persistence, `0600` + 32 bytes, stable KEK across a cache reset ("restart").

Local: full `make unit-tests` green, `make server` (LTO) builds, `make lint` + `make build-integrity` pass.

## Follow-ups (scoped out, by decision)

- **Wire `agent add --key` to push into the vault** (and stop serializing literal `api_key` into `agents.json`). The push route (`/v1/vault/set`, now dual-wrapping) already exists; the open design points are unlock-at-creation UX and TCP (no per-user vault by design, D17) — worth deciding in review.
- Fold the OAuth tokens (`oauth_tokens.c` via `db1/secrets`, today plaintext-at-rest in a container) into the server vault.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -375,7 +375,7 @@ SERVER_SRCS = server/server_main.c server/server.c server/server_http.c server/s
               server/compute_concurrency.c server/provider_catalog.c \
               server/dashboard_server.c hud.c cmd_identity.c prompts.c persona.c working_profile.c \
               server/delegate_role.c server/delegate_depth.c server/delegate_prompt.c server/delegate_run_phases.c server/delegate_routing.c server/delegate_launch.c server/delegate_source_authority.c server/delegate_economics.c server/delegate_credentials.c server/delegate_credential_retry.c server/delegate_patch_coordinator.c server/delegate_ensemble.c server/delegate_checkout.c cmd_init.c \
-              server/trigger_scheduler.c server/server_trigger.c server/server_cron.c server/server_trajectory.c server/server_config.c server/vault_principal.c server/vault_crypto.c server/vault_kek_cache.c server/vault_store.c server/vault_service.c server/server_vault.c
+              server/trigger_scheduler.c server/server_trigger.c server/server_cron.c server/server_trajectory.c server/server_config.c server/vault_principal.c server/vault_crypto.c server/vault_kek_cache.c server/vault_server_key.c server/vault_store.c server/vault_service.c server/server_vault.c
 SERVER_OBJS = $(SERVER_SRCS:%.c=$(OBJDIR)/%.o)
 SERVER_DATA_SRCS = $(filter-out $(SERVER_SRCS) guardrails_orchestrator.c,$(DATA_SRCS))
 SERVER_DATA_OBJS = $(SERVER_DATA_SRCS:%.c=$(OBJDIR)/server/%.o) \
--- a/src/headers/vault_server_key.h
+++ b/src/headers/vault_server_key.h
@@ -1,0 +1,47 @@
+#ifndef DEC_VAULT_SERVER_KEY_H
+#define DEC_VAULT_SERVER_KEY_H 1
+
+#include <stdint.h>
+#include "vault_crypto.h" /* VAULT_KEK_LEN */
+
+/* vault_server_key: the SERVER-sealed key-encryption-key for dual-access
+ * credentials (WP-C.4).
+ *
+ * The per-user vaults (vault_store/vault_service) are zero-knowledge: a
+ * credential's data-encryption key (DEK) is AES-KW-wrapped under a KEK derived
+ * from the client root key (uid:) or login password (webuser:), so the server
+ * cannot read a user's secret without that client unlocking. That makes a
+ * delegate UNRUNNABLE after a restart / KEK-cache TTL expiry — the very thing we
+ * must fix: aimee-server has to drive delegates autonomously.
+ *
+ * Dual-access wrap: every vaulted credential's DEK is wrapped a SECOND time under
+ * this server KEK (in addition to the user KEK). The server can then decrypt the
+ * credential on its own — across restarts, with no client — while the user's
+ * zero-knowledge path is unchanged. The secret is still encrypted at rest (AES-
+ * 256-GCM under a random DEK); the server KEK only unwraps the DEK.
+ *
+ * Trust boundary (deliberate, per the credential-storage directive): the master
+ * key lives 0600 on the server data volume beside the vault. This defeats backup
+ * / disk theft and plaintext-in-config, NOT a root-level host compromise (root
+ * can read both key and data) — the same boundary the rest of aimee-server runs
+ * under. The module is structured so an operator passphrase could wrap the master
+ * key file later without a data migration.
+ *
+ * The 32-byte master key is generated with vault_crypto_random on first use and
+ * persisted at <config_default_dir>/.vault/.server-master.key. The server KEK is
+ * HKDF-derived from it (domain-separated from the user KEKs) and cached process-
+ * wide. Fail-closed: any entropy/IO/derivation failure returns -1 and the caller
+ * must abort the vault op rather than store/read an unprotected credential. */
+
+/* Load-or-create the master key file, derive the server KEK, and write it to
+ * `kek` (VAULT_KEK_LEN bytes). The derived KEK is cached after the first call.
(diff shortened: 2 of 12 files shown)


JBailes:
## Roundtable review (security persona, `.254`)

Ran a roundtable review of this PR. The cryptographic **design was confirmed sound**: one random per-credential DEK, AES-KW-wrapped under two independent KEKs (no IV/nonce, so the wraps can't interact), with AES-256-GCM over the secret bound by AAD = `principal|agent|cred` (blocks cross-credential substitution). The deliberate trust-boundary change is honestly documented.

> Note: the diff fed to the panel was partially truncated server-side, so `vault_store.c` / most of `vault_service.c` were flagged as "couldn't verify" (Q1–Q5). I've verified those against the actual code (below).

### Findings fixed (commit 7202523)
- **F1 (Medium):** `derive_and_cache` minted a fresh master key on *any* read failure → a transient `EACCES`/`EIO` or wrong-size file on an existing key would overwrite it and orphan every `wrapped_dek_server`. Now `master_key_read` distinguishes `ENOENT` (absent → safe to mint) from present-but-bad (**fail closed, never overwrite**).
- **F2 (Low):** cleanse the master buffer on the read-back-failure return.
- **F4 (Low):** `fsync` the parent dir after `rename` for crash durability.
- New test: a present-but-corrupt master key fails closed and is left untouched.

### "Couldn't verify" items — confirmed against code
- **Q1 (backfill atomicity):** `vault_store_add_server_wraps` is all-or-nothing — it breaks with `rc = -1` on any unwrap/wrap failure and only writes when `rc == 0 && changed`. ✓
- **Q2 (fail-closed on server-KEK unavailable):** `vault_service_set` returns `VAULT_ERR_CRYPTO` and stores nothing if `vault_server_kek()` fails; `inject` falls back to the user path only on `NO_ENTRY`, not on `CRYPTO` — no silent single-wrap write. ✓
- **Q3 (domain separation):** the server KEK uses a fixed 16-byte `SERVER_KEK_SALT`; user KEKs use random per-principal salts — collision is ~2⁻¹²⁸. ✓
- **Q4 (`out` cleansing):** `vault_store_get_server` cleanses `out` on every non-success return. ✓
- **F5 / Q5 (trust boundary, master-key loss):** documented in the header; the master key must be backed up — losing it orphans server wraps (user wraps still recover via unlock). Will mirror this in release notes / data-handling docs.

JBailes:
## Roundtable review — wiring (agent-add → vault push)

Ran a second roundtable against the wiring commits. This time the review delegate's evidence bundle came back with **0 changed files** — the diff was already committed/pushed, and the delegate (running on `.254`) can't see this local branch, so the stricter persona declined to grade from the in-prompt diff alone. It instead produced a precise verification checklist; I've closed each item against the code:

- **AUTH/AUTHZ (Q1):** `/v1/agent/add` is bearer-gated (cap `0`) — the *same* gate that already let `agent.add` write `agents.json`. The server-vault write adds no new privilege; it's bounded by the deploy bearer's trust domain. ✓
- **Tenancy of the shared `server` principal (Q2):** real and now **documented** (87dadf4). aimee-server is one person's agent; the TCP path is a single trust domain; webchat users get isolated `webuser:` vaults and never use this principal. If aimee-server ever goes multi-tenant over TCP, the `server` principal must be namespaced by attested identity — called out in the header.
- **D15 / no silent downgrade (Q6):** confirmed in code — `vault_service_get` returns `VAULT_ERR_LOCKED` (not `NO_ENTRY`) when a user cred exists but the vault is locked, and `inject` only falls back to the server principal on `NO_ENTRY`. A locked user cred is never downgraded to the server vault. ✓
- **Fail-closed (Q4):** every non-`VAULT_OK` status returns before `ag->api_key` is cleared/saved; the secret is never written to `agents.json`. ✓ The `$VAR` reference branch is unchanged (pre-existing behaviour; a `$VAR` ref is not a secret).
- **`set_server`/`get_server_principal` hygiene (Q5):** `OPENSSL_cleanse` on `kek`/`salt` on all paths; the get path uses `vault_store_get` directly (no `get_or_create_salt`), so a missing vault returns `NO_ENTRY` without a side-effecting file create. ✓
- **Orphan entry on save failure (Q3):** if `agent_save_config` fails after the vault push, the vault entry is an upsert that a retry overwrites; a key with no agent def is never resolved. Harmless; noted.

Net: no code defects required; one documentation hardening landed.

JBailes:
## Roundtable review — wiring (graded pass)

Got a real graded run this time (the earlier passes were blocked because the `.254` review delegate runs in a sandbox that can't see the local worktree, so its auto-gathered evidence bundle was always empty — fed it the full diff inline with an explicit "grade this, the empty bundle is expected" instruction).

Findings + resolutions (commit b5b9b48):

- **F2 — medium (fixed).** The `$VAR` branch called `agent_expand_env` into `ag->api_key`, resolving the env var to **plaintext**, which `agent_save_config` then writes to `agents.json` — reopening the very leak the literal-key branch closes, and contradicting its own "kept as a reference" comment. Now the `$VAR` reference is stored **unexpanded**; `agent_load_config` expands it from the environment at run time, so `agents.json` holds `"$VAR"`, never the secret.
- **F3 — low (fixed).** Renamed the per-principal server-wrap reader `vault_service_get_server` → `vault_service_get_server_wrap`, so it's no longer confusable with `vault_service_get_server_principal` (the server-vault reader).
- **F1 — low (addressed).** Tenancy of the shared `server` principal is acceptable under the single-deploy/single-bearer model and is documented; added a test pinning `VAULT_SERVER_PRINCIPAL == "server"` so a future multi-tenant refactor that needs per-identity namespacing is forced to be deliberate.
- **D15** (a LOCKED user cred is never downgraded to the server vault) — the review **confirmed correct**; no change.

Findings 4–6 (set_server/get hygiene, ordering, test) were cut off by the result-size cap, but those were verified in the prior pass (cleansing on all paths, no side-effecting create on the read path, orphan-on-save-failure is a harmless upsert). Local `make unit-tests` / server build / lint green.

PR #249, opened 2026-06-13
Move OAuth token storage into the server-sealed vault (WP-C.4)

JBailes:
## Why

OAuth access/refresh/expiry tokens (codex/gemini) were persisted via `db1/secrets`, whose only container-viable backend is a **plaintext 0600 file** (libsecret has no Secret Service daemon in a container). So provider OAuth tokens sat **unencrypted at rest** — the last plaintext credential path after the vault landed (#244).

## What

Route `oauth_tokens.{store,load,remove,refresh,get}` through the **server-owned vault** (`VAULT_SERVER_PRINCIPAL`, AES-256-GCM at rest). The server still reads/refreshes tokens autonomously (the server KEK needs no unlock), but they're now encrypted.

- New static helpers `oauth_secret_{store,load,remove}` wrap `vault_service_set_server` / `vault_service_get_server_principal` / `vault_service_delete`, keyed by `(server, client_name, oauth_*_cred)`.
- **Lazy migration:** on first read, a legacy `db1/secrets` plaintext value is moved into the vault and the plaintext is scrubbed.
- Exposed `vault_service_get_server_principal` in the header (was static).
- **KB unaffected:** `oauth_tokens.o` is in `AGENT_SRCS` (server/full binaries only, which already link the vault), not the kb binary. Added the vault objects to the `test_oauth_pkce` link.

## Tests

`test_oauth_pkce`: existing store/load/remove round-trip now runs through the vault; new case plants a legacy plaintext token, confirms `oauth_token_load` migrates it, scrubs the old key, and serves from the vault thereafter.

Local: full `make unit-tests`, server + kb (LTO) builds, `make lint`, `make build-integrity` all green.

## Note

Server-vault trust boundary applies (master key beside data; see `vault_server_key.h`). The Codex CLI's own `~/.codex/auth.json` is a separate file owned by the Codex subprocess and is out of scope here.

### DIFF
--- a/src/headers/vault_service.h
+++ b/src/headers/vault_service.h
@@ -85,6 +85,12 @@ vault_status_t vault_service_set(const char *principal, const char *agent, const
  * VAULT_OK, or VAULT_ERR_CRYPTO/IO on a fail-closed error. */
 vault_status_t vault_service_set_server(const char *agent, const char *cred, const char *secret);
 
+/* Read (agent, cred) from the server principal's vault under the server master KEK
+ * — no client, no unlock. VAULT_OK (plaintext written), VAULT_NO_ENTRY (no file or
+ * no such entry), or VAULT_ERR_* (fail closed). `out` is cleansed on non-OK. */
+vault_status_t vault_service_get_server_principal(const char *agent, const char *cred, char *out,
+                                                  size_t out_len);
+
 /* The use-path: resolve (agent, cred) for `principal` into `out`. Returns:
  *   VAULT_OK         + plaintext written;
  *   VAULT_NO_ENTRY   when there is no such credential (or no/blank principal) —
--- a/src/server/oauth_tokens.c
+++ b/src/server/oauth_tokens.c
@@ -14,6 +14,7 @@
 #include "db1.h"
 #include "oauth_flow.h"
 #include "oauth_pkce.h"
+#include "vault_service.h" /* server-sealed vault: encrypted-at-rest token storage */
 #include "cJSON.h"
 #include "log.h"
 #include "platform_process.h"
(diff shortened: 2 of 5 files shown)


JBailes:
## Roundtable review — graded

Findings + dispositions (fixes in da084e3):

**Fixed**
- **Critical — token loss in migration (real bug I introduced).** `oauth_secret_load` scrubbed the legacy plaintext via `(void)vault_service_set_server(...); (void)db1_secret_remove(...)` — so if the vault write failed (e.g. no master key), the only copy of the token was deleted. Now the plaintext is removed **only after** `vault_service_set_server` returns `VAULT_OK`; otherwise it's kept and migration retries next read. The value is still returned to the caller either way.
- **Logging (was flagged High).** `oauth_secret_store` now logs a vault-store failure. The error already propagated via `rc`; this just makes it observable.
- **Test coverage (Medium).** The migration test now asserts **both** legacy keys that `load` touches (access + expires) are scrubbed.

**Assessed, no change (verified against code the panel couldn't see)**
- **Stale buffer on `VAULT_NO_ENTRY` (flagged Medium):** N/A — `vault_service_get_server_principal` zeroes `out` at entry, so `buf[0]` is always cleared before the `&& buf[0]` check.
- **Concurrency double-migration (flagged High):** `vault_store_set` is internally mutex-guarded and migration writes are idempotent (same value); with the scrub-after-write fix there's no loss risk — it reduces to harmless redundant work for two concurrent first-readers.
- **Buffer not cleansed on migration failure (flagged High):** N/A — on a failed vault write the value is legitimately returned in `buf` for the caller to use as the token; cleansing it would break the contract (same as the pre-existing `db1`-backed behaviour).
- **`refresh_token` migration not tested (Medium):** the refresh token is only read during a network refresh, so it migrates lazily then, not on a plain `load` — can't unit-test without the network path; noted in the test comment.
- **`expires_at` in the vault (Low):** intentional — keeps the triplet together; it's non-secret but harmless.

Local: `make unit-tests`, server+kb LTO builds, lint, build-integrity all green.

PR #251, opened 2026-06-13
agents.json: never re-serialize a resolved $VAR api_key as plaintext

JBailes:
## Why

`agent_load_config` expands a `$VAR` `api_key` into `agent_t.api_key` for runtime use, and `agent_save_config` then wrote that **resolved** value back to `agents.json`. So any agent-config mutation (`agent add`/`enable`/`remove`) re-serialized the secret into the file as **plaintext** — the last remaining `agents.json` plaintext path after the credential vault work (#244/#249).

## What

Add `agent_t.api_key_disk` holding the **verbatim on-disk form** (a `$VAR` reference, or a legacy literal):
- **Load** keeps the reference in `api_key_disk` and resolves a separate runtime copy into `api_key` (unchanged runtime behaviour — the auth-header builders read the resolved `api_key`).
- **Save** writes `api_key_disk`, falling back to `api_key` for in-memory agents (e.g. `agent add $VAR`, which already stores the unexpanded reference there). A resolved secret is never written.

## Test

`test_agent_apikey.c` (split out of the 2000-line-capped `test_agent.c`, mirroring its link line): a `$VAR` api_key round-trips `load → save` with the reference preserved and the secret **never** present in the file, then reloads to the resolved value.

Local: `make unit-tests`, server+kb (LTO) builds, `lint`, `build-integrity` all green.

## Note

This closes the `$VAR`-reexpansion leak. A legacy *literal* secret already sitting in `agents.json` stays as-is (not made worse); the recommended path for new secrets is the vault (`agent add` routes literals there). Migrating pre-existing literals out of `agents.json` would be a separate change.

### DIFF
--- a/src/cmd_agent.c
+++ b/src/cmd_agent.c
@@ -744,7 +744,14 @@ static void ag_add(app_ctx_t *ctx, int argc, char **argv)
    for (int i = 3; i < argc; i++)
    {
       if (strcmp(argv[i], "--key") == 0 && i + 1 < argc)
-         agent_expand_env(argv[++i], ag->api_key, MAX_API_KEY_LEN);
+      {
+         /* Preserve the verbatim reference (e.g. "$VAR") for save so a resolved
+          * secret is never written to agents.json; resolve a runtime copy.
+          * Mirrors agent_load_config. */
+         const char *raw_key = argv[++i];
+         snprintf(ag->api_key_disk, MAX_API_KEY_LEN, "%s", raw_key);
+         agent_expand_env(raw_key, ag->api_key, MAX_API_KEY_LEN);
+      }
       else if (strcmp(argv[i], "--auth-cmd") == 0 && i + 1 < argc)
          snprintf(ag->auth_cmd, MAX_AUTH_CMD_LEN, "%s", argv[++i]);
       else if (strcmp(argv[i], "--auth-type") == 0 && i + 1 < argc)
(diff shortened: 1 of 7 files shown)


JBailes:
## Roundtable review — graded

The panel correctly flagged that the save fallback trusted an unproven invariant. Resolutions (commit 513a851):

**Fixed**
- **Q1 — High, real leak.** The fallback (`api_key_disk[0] ? api_key_disk : api_key`) assumed `api_key` was an unexpanded reference when `api_key_disk` was empty. True for `server_agent.c handle_agent_add`, but **`cmd_agent.c`'s `agent add --key`** did `agent_expand_env(... ag->api_key)` then saved — writing the **resolved secret** to `agents.json`. Now `cmd_agent.c` stores the verbatim reference in `api_key_disk` and resolves a runtime copy into `api_key` (mirrors load). The leak is closed at the writer.
- **Q4 — Medium, vault interaction.** `handle_agent_add` now sets `api_key_disk` in the `$VAR` branch and **clears it alongside `api_key` on the vault path**, so a vaulted literal never lands in `agents.json`. Also cleared `api_key_disk` at the `cmd_agent_setup.c` clear sites. The invariant ("`api_key_disk` tracks the on-disk form for any config-persisted agent") now holds at every writer.
- **Q5 — Low, test gap.** Added an in-memory fallback test (`api_key="$VAR"`, `api_key_disk` empty) asserting the reference, not a secret, is written.

**Assessed, no change**
- **Q2 (load population):** the panel confirmed it correct (both fields set atomically).
- **Q3 (struct size / binary serialization):** `agent_t` is only (de)serialized via JSON (`agent_save_config`/`agent_load_config`); no `sizeof(agent_t)` wire/binary path. `memset`/`memcpy` users are safe (new field zeroed/copied). Low, no action.
- Runtime delegate paths that set `api_key` from env/vault (`delegate_credential_retry.c`, `delegate_ensemble.c`, `provider_cli_adapter.c`) are never followed by `agent_save_config`, so they can't leak to disk — left as-is.

Local: `make unit-tests`, server+kb LTO builds, lint, build-integrity all green.

PR #291, opened 2026-06-14
Credential-vault consolidation P1: codex/OAuth into the server vault + setup-against-remote

JBailes:
## Credential-vault consolidation — P1: codex/OAuth into the server vault + setup-against-remote

First packet of the USER-approved [credential-vault consolidation proposal](docs/proposals/accepted/cred-vault-consolidation.md) (rev. 3). Builds on the already-merged server-sealed vault (WP-C). Scoped as one coherent, reviewable slice: **codex credentials are stored in the server-owned vault and provisionable from a thin client.**

### What this does
- **Codex → server vault (write gap, G2):** the codex device-setup now stores the access token + `ChatGPT-Account-ID` under the server principal (`vault_service_set_server`), which decrypts autonomously. The disk `codex-auth.json` stays as a migration-era fallback until WP-4.
- **Use-path server-principal fallback:** `codex_oauth_apply_vault_override` now falls back to `vault_service_get_server_principal` when the per-turn (attested) principal misses — the only principal a TCP thin client has. A *locked* user cred stays a hard miss, never a silent downgrade (D15).
- **D7 — `agent setup` targets the configured remote:** setup used `cli_v1_dispatch_local` (always the local UDS) and unconditionally required a local server, so on a thin-client→remote deployment it failed `server unavailable` and could never provision codex onto the server that uses it. Adds a remote-aware `cli_v1_dispatch` and routes the `agent.setup`/`agent.setup_poll` device flow through it; only requires a local server when no remote is configured.

Together: the device flow now provisions the codex token into the **remote** server's vault, and delegates authenticate from it.

### Testing
- `make -j1 ../aimee ../aimee-server` — clean (client + server).
- `make -j1 verify-local` — **ok** (lint + check-linking + unit suite). Added the `vault_service_get_server_principal` keyless-miss stub for `unit-test-server-compute` (links `delegate_credential_retry.o` against stubs, not the real `vault_service.o`).
- **Live end-to-end on `.254` is deploy-gated** (new client + server) and not exercised here; the code paths are build-verified.

### Scope / follow-ups (not in this PR)
- **P1b:** codex token refresh (D6) + `codex reauth` operator command + atomic `codex-auth.json`→vault lazy-migration (D8). Codex authenticates from the vault without these; refresh adds durability past token expiry.
- **P2–P4:** capability core + client key-forward; migration state machine; `vault_only` + legacy retirement — separate PRs.

### DIFF
--- a/docs/proposals/accepted/cred-vault-consolidation.md
+++ b/docs/proposals/accepted/cred-vault-consolidation.md
@@ -1,0 +1,344 @@
+# Proposal: complete credential-vault consolidation
+
+- **State:** rev. 3 (R1+R2 reviewed → revised; fleet can only reliably review a doc this size
+  via one model — see §Review history; recommending USER proposal-approval as gate 1)
+- **Status refreshed:** 2026-06-14
+- **Author:** JBailes
+- **Date:** 2026-06-14
+- **Builds on:** [delegate-refactor-async-and-credential-vault](../done/delegate-refactor-async-and-credential-vault.md)
+  (the envelope-vault + server-sealed dual-access core, WP-A..WP-C.4, already merged
+  to `testing` and **deployed live on `.254` in `aimee-server v0.2.68`**).
+- **Charter roles:** Secure (one encrypted-at-rest home for every credential,
+  fail-closed), Delegate (every delegate authenticates from the vault), Gate-Promote
+  (default-safe migration with coexistence + rollback).
+
+> **Revision note (R1).** Reviewed by a diverse delegate panel; security returned
+> **BLOCKED**, reviewer **CONCERNS** (architect/engineer lenses were lost to a provider
+> 429 + a reasoning-model stall and are owed in R2). Three convergent findings drove
+> rev. 2: (1) **server-principal writes were under-protected** — bare `/v1` bearer over
+> plaintext `0.0.0.0:8740` could mint server-owned creds (trust-boundary collapse); rev. 2
+> gates them behind a dedicated capability + per-write audit + a transport requirement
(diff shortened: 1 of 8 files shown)


JBailes:
## Roundtable review (round 1) — resolution

Diverse-lens review (security + engineer) returned CONCERNS; the reviewers ran sandbox-blind (no worktree access), so several findings were "unverified contracts." Resolution:

**Fixed (commit 36d69b7):**
- **F2 (medium, correctness):** the `agent setup` local-server guard used `!cli_rpc_remote_endpoint_is_tcp()`, which would wrongly require a local server for a *unix-socket* remote. Switched to `!cli_rpc_has_remote_endpoint()` — true for any configured remote (tcp **or** unix), matching exactly when `cli_v1_dispatch` routes off-box.
- **F7 (test gap):** added `test_codex_oauth_vault_server_principal_fallback` driving the real `delegate_resolve_credentials` — asserts the server-principal fallback applies the token with no per-turn principal, a LOCKED per-turn cred is a hard miss (D15, server principal not consulted), and all-miss applies nothing.

**Cleared by code (no change needed):**
- `cli_v1_send(NULL, …)` **is** the local-UDS path — it's exactly what `cli_v1_dispatch_local` calls; no local-case regression.
- Writer `vault_service_set_server("codex", …)` matches the reader `target_agent->name` — the codex agent is named `"codex"` by the setup convention.
- `cli_v1_dispatch` owns and frees `body`; the callee does not take ownership.

**Deferred to P1b (refresh) with rationale:** server-principal-write audit/observability log lines and intermediate-decrypt zeroing hygiene — they live in the same code the refresh work touches.

verify-local: ok (incl. the new test); CI green pre-fix except the historically-flaky e2e-docker.

PR #294, opened 2026-06-14
Credential-vault consolidation P2a: capability-gated server-principal writes

JBailes:
## Credential-vault consolidation P2a: capability-gated server-principal writes

Second packet of the [consolidation proposal](docs/proposals/accepted/cred-vault-consolidation.md) (after P1 #291). Implements WP-1 / D2b / D2c — the security core that gates who may land a credential in the server-owned (autonomously-decryptable) vault.

### What this does
- **Closes an ungated write path:** `agent add --key` over an unattested (plaintext-TCP) connection previously *silently* wrote the key to the server principal. Now it requires an **attested** (UDS/webchat) connection — where it writes the **caller's own** per-user vault — or is refused with a pointer to `vault set-server`.
- **`vault.set_server`** (new): writes a client-supplied secret under the **server principal**, gated on (a) an attested transport (`UDS_PEERCRED`/`WEBCHAT_TRUSTED`, never a TCP bearer — D2b) **and** (b) the **`vault:write:server` capability** for the caller's principal (D2c). Audited with a SHA-256 key **fingerprint** (never the key).
- **`vault.capability`** (new, UDS-only): operator-minted allow-list (`grant`/`revoke`/`list`) backed by `vault_capability` — a 0600 atomic grants store under `.vault/`.
- **CLI:** `aimee vault set-server <agent> <cred> <secret>` and `aimee vault capability grant|revoke|list [principal]`.
- **Scope note:** only **client-supplied** server-principal writes are gated. **Server-obtained** secrets (the codex/OAuth tokens the *server* fetched in P1) are unaffected — gating them on the originating client's transport would wrongly break codex consolidation over TCP.

### Operational consequence (by design, per approved rev.3 D2b)
A plaintext-TCP thin client can **no longer** provision shared/delegate keys directly — server-principal writes require an attested transport (UDS on the server box, or webchat) holding the capability. Provisioning moves to the box (or add TLS). P3 (migration) runs over UDS accordingly.

### Testing
- `make -j1 verify-local` — **ok** (build + v1-route-coverage + api-conformance + lint + unit suite).
- New `unit-test-vault-capability` (grant/revoke/has/list, whole-line match, input validation) — passes.
- Added dispatch-table + capability-test link stubs (`vault_handlers_stub.c`, `test_server_dispatch`).
- Live end-to-end is deploy-gated; code paths are build- + unit-verified.

### Follow-ups
- **P2b:** client `agent add --key` forwarding over attested transport (the client side of provisioning).
- **P3:** migration state machine (`agent key import`). **P4:** `vault_only` + legacy retirement.


JBailes:

Security + engineer reviews returned CONCERNS (both ran sandbox-blind — they couldn't open the worktree, so most findings were "unverified"). The two convergent, real findings are fixed in `8a89b15`:

- **Audit mislabel (both, medium):** `handle_vault_set_server` logged `transport=uds` even when the gate accepts webchat → now logs the actual transport.
- **No gate test (both):** extracted the gate into `vault_capability_server_write_allowed(transport, principal)` (attested transport AND capability) as the single source of truth, used it in the handler, and unit-tested it directly — TCP/NONE refused, UDS/webchat-with-cap allowed, attested-without-cap refused, empty principal refused.
- **Durability nit:** `cap_write_atomic` now fsyncs the parent dir after rename.

Cleared by code (their no-worktree limitation): `vault_capability.c` *was* in the diff (atomicity/0600/validation/idempotency all present and now gate-tested); the webchat `vault_principal` provenance is set by the existing WP-C.2 `server_http_identity.c` (server.token-gated `X-Aimee-Webuser`), not a self-asserted header introduced here. TOCTOU on capability check is a tiny, non-revocable-mid-handler window — noted, not blocking.

verify-local: ok (incl. the new gate tests); CI was fully green pre-fix (all 11).

PR #298, opened 2026-06-14
Credential-vault consolidation P3: agent key import (migrate keyring into the vault)

JBailes:
## Credential-vault consolidation P3: `aimee agent key import`

Third packet (after P1 #291, P2a #294). WP-3 — the migration that moves existing client-held delegate keys into the server vault.

### What this does
`aimee agent key import [--scrub]` (client-orchestrated): reads the local `agent-keys.json` and pushes each entry into the **server vault** under the server principal via the P2a `vault.set_server` route, reporting per agent:
- **vaulted** — stored;
- **SKIPPED (…)** — refused (over a plaintext-TCP remote, since P2a/D2b requires an attested transport + the `vault:write:server` capability) or errored; the entry is **left intact**.
- `--scrub` (opt-in, default off): removes an entry from the local keyring **only after** a confirmed store — keys otherwise stay as a backup until P4 retires the legacy path.
- Idempotent / re-runnable.

### Notes
- For the plaintext-TCP thin-client → `.254` deployment, this is **run on the server host over UDS** (or with the capability granted) — consistent with P2a's security model. The command reports clearly when refused and how to proceed.
- The handler lives in `cli_agent_keys.c` (which already has keyring + RPC access) rather than `cli_main.c`, to stay under the 2000-line file cap; `cli_agent_keys_load` is exposed for it.

### Testing
- `make -j1 verify-local` — **ok**. Pure client change: no new server route / openapi / dispatch surface, so no route-coverage or stub changes.
- The migration loop is RPC-orchestration (integration-level, exercised live); the keyring I/O it builds on is pre-existing and covered by the session-push path.

### Follow-up
- **P4:** `vault_only` flag + retire the legacy `agent-keys.json` push + `codex-auth.json` read (this is where `--scrub`/retirement becomes the default).

### DIFF
--- a/src/cli_agent_keys.c
+++ b/src/cli_agent_keys.c
@@ -68,6 +68,11 @@ static cJSON *load_keys(void)
    return root;
 }
 
+cJSON *cli_agent_keys_load(void)
+{
+   return load_keys();
+}
+
 int cli_agent_key_set(const char *agent_name, const char *api_key)
 {
    if (!agent_name || !agent_name[0])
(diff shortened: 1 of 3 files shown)


JBailes:

Security review was **BLOCKED purely on sandbox access** (it couldn't reach the worktree — no findings). Engineer review = CONCERNS. Disposition:

- **#1 (HIGH, "ArrayForEach over an object iterates nothing → migrates zero keys") — FALSE ALARM.** `cJSON_ArrayForEach` walks an object's members via `->child`/`->next` (with `entry->string` = key); the *same* keyring object is already counted with `cJSON_GetArraySize` in `cli_session_creds_prime`, and the session-push works. No change needed.
- **#2 (MEDIUM, `--scrub` relies on `ok` ⇒ durably persisted) — confirmed safe.** The `vault.set_server` handler (P2a) calls `vault_service_set_server` synchronously and returns `ok` only after the write; `--scrub` fires only inside the `ok` branch.
- **#4 (`$ENV` refs uploaded verbatim) — N/A.** `agent-keys.json` holds *literal* keys (the client-held keyring); `$ENV` refs live in `agents.json` (`api_key_disk`), which this command does not read.
- **#5 (no --dry-run / suggestions) — added `--dry-run`** (`b2166cb`): previews entries without sending any secret or mutating the keyring. Default off.
- **#3 (refusal classified by free-text substring) — minor/cosmetic:** only affects the summary label and the exit code (refused⇒0, error⇒1); the keyring entry is left intact in both cases (no safety impact). Left as-is.

No dedicated unit test: the migration loop is RPC-orchestration (integration-level) and linking `cli_agent_keys.o` for a unit test pulls in the full client RPC stack (rabbit hole for low value); the keyring I/O it builds on is pre-existing and exercised by the session-push path.

verify-local: ok; CI was fully green pre-fix (all 11).

PR #300, opened 2026-06-14
Credential-vault consolidation P4: vault_only flag (retire legacy credential paths)

JBailes:
## Credential-vault consolidation P4: `vault_only` — retire the legacy paths

Final packet (after P1 #291, P2a #294, P3 #298). WP-4/D10–D12 — the cutover lever that makes the server vault the sole credential source and retires the legacy paths.

### What this does
A new `vault_only` config bool (default **off**). When set:
- **`/v1/session/credentials` (the legacy client RAM-push) is rejected** (D11) with a clear error pointing at `aimee vault set-server` / `aimee agent key import` — an un-upgraded thin client gets an explicit failure instead of silently relying on a push that no longer authenticates anything.
- **the on-disk `codex-auth.json` / `~/.codex/auth.json` read is skipped** (D12) — codex tokens come only from the vault.

Default **off** keeps today's vault-first-with-legacy-fallback behavior (the migration window). Flip to `1` only after `aimee agent key import` (P3) has moved every credential into the vault; **flipping it back to `0` is the rollback lever**.

### Wiring
- Config field (`config.h`), load parse (`config.c`), save (`config_save.c`), CLI surface (`config_fields.c`), regenerated `docs/gen/configuration.md`.
- Gates: `handle_session_credentials` (`server_runner_endpoints.inc`), `agent_read_codex_oauth_token` (`agent_config.c`).

### Testing
- `make -j1 verify-local` — **ok** (build + config-surface + docs-gen + lint + unit suite).
- `test_config`: `vault_only` default-off + a save/load round-trip (this test caught a real bug — the load path needed a `config.c` parse entry, not just the CLI `config_fields[]` surface).

### Scope / not in this PR (notable deferrals from D10/D13)
- The `vault_only --dry-run` blast-radius enumerator and the formal cutover gate (0 legacy reads for N days) — operational tooling; the flag + rollback are here.
- `.server-master.key` rotation + `agent add --key` upsert semantics (D13) — separable hardening.
- The env-lease (`agent.credentials` pool) path is not gated here (our delegates use the single-key session-push path, which *is* gated).

### After this merges
The consolidation is **code-complete** (P1–P4). It only takes effect live after a **`.254` deploy** of the merged `testing` build (user-gated).

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -12,7 +12,7 @@ aimee config set <key> <value>    # set one value
 
 Structured options (arrays, nested objects — e.g. `ensemble.reference_models`) are not CLI-settable; they are written into the config file under the sections listed at the end.
 
-## CLI-settable keys (108)
+## CLI-settable keys (109)
 
 | Key | Type |
 |-----|------|
@@ -120,6 +120,7 @@ Structured options (arrays, nested objects — e.g. `ensemble.reference_models`)
 | `provider` | string |
 | `reasoning_cap_enabled` | bool |
 | `typed_facts_enabled` | bool |
+| `vault_only` | bool |
 | `verify_cross_project` | bool |
 | `verify_enabled` | bool |
 | `virtual_context_assembly_budget` | int |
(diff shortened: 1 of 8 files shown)


JBailes:
Superseded: dropping the vault_only flag per direction — the vault becomes the permanent, sole credential store (no cutover toggle), client-held agent-keys.json retired entirely, and provisioning made automatic over an attested path. Reworking.

PR #303, opened 2026-06-14
Native TLS phase 1b: in-process TLS listener

JBailes:
## Native TLS — phase 1b: in-process TLS listener

Builds on phase 1a's conn-io shim (#302). aimee-server now terminates TLS itself — no proxy.

### What this does
- New optional **`aimee.api.tls_port`** (`server_api_tls_port`): binds a TLS listener alongside the UDS + plaintext-TCP ones. Cert+key load from `<config>/tls/server.crt` + `server.key` (plain server TLS — **no client cert**; a TLS+bearer connection is the attested write path for the credential vault).
- **`server_tls.{c,h}`**: the server `SSL_CTX` (TLS 1.2+, cert+key, no `SSL_VERIFY_PEER`), the handshake, and per-conn `begin`/`end` that register/clear the `SSL` on the phase-1a conn-io shim.
- Each TLS connection runs in its **own `conn_worker`**, which owns the `SSL` end to end (handshake in the worker so the accept loop never blocks; teardown = unregister + shutdown + free + close).
- **SSE-offload is refused over TLS** (it `dup()`s the fd to a second thread that can't share the conn's `SSL` → would emit plaintext) — the caller returns 503. Inline streaming + unary requests over TLS work transparently via the conn-io shim. (SSE-over-TLS is phase 1c.)

### Notes
- `conn_worker`/`conn_offload` moved to `server_http_conn_worker.inc` to keep `server_http.c` under the 2000-line cap.
- Not added to CMakeLists (CMake's server build doesn't compile `server_http.c`; phase 1a confirmed this — macos/windows-cmake pass).

### Testing
- `make -j1 verify-local` — ok (build + line-check + config-surface + lint + unit suite). Default off (`tls_port` unset) ⇒ no behavior change.
- Live TLS handshake is exercised at deploy time (cert on `.254`); the client already speaks https (`aimee_tls.c`).

### Next
- **Phase 3:** a TLS+bearer conn becomes an attested transport (`vault_principal_resolve`) → `agent add --key` over TLS auto-provisions the server vault.
- **Phase 4:** retire the client-held `agent-keys.json` path entirely (vault permanent, sole source).


JBailes:

Both seats: **CONCERNS, no blockers** — "the SSL lifecycle, concurrency, and accept-loop correctness as written are sound." Hardening applied (`5b91785`):
- **Fail-loud on misconfig (both, medium):** a configured `tls_port` whose cert/key won't load now `LOG_ERROR`s (was WARN) + surfaces the OpenSSL error; the UDS listener still comes up (don't kill the whole server for a cert typo).
- **Private-key perms (both, medium):** warn if the key file is group/world-readable.
- **Handshake/idle timeout (171, medium):** `SO_RCVTIMEO`/`SO_SNDTIMEO` (30s) on the TLS fd so a stalled peer can't pin a per-conn worker.

Dispositioned (not changed): `SSL_ERROR_WANT_WRITE`/`WANT_READ` retry (171 #4) — **N/A, the fds are blocking** (no non-blocking SSL); `SSL_CTX_free` in stop — benign (process-lifetime ctx); cipher-suite floor / `SSL_MODE_RELEASE_BUFFERS` — OpenSSL 3.0 secure defaults, single-user; `pfds[3]` — correct for the 3 listener classes; handshake-failure `close()` without a TLS alert — cosmetic.

verify-local green; CI was fully green pre-fix (all 11).

PR #305, opened 2026-06-14
feat(vault): make the vault the universal credential source (P4a)

JBailes:
## What

Make the credential **vault** the universal source for *every* connection — the core of "all configured agents, all connections, in the vault, permanently."

Until now the vault was read **only** on the delegate path (`delegate_credential_retry` pre-injects into `agent->api_key`). The universal resolver `agent_resolve_auth` used `session-creds → literal → env` and never consulted the vault, so a **primary chat / webchat / in-model** agent could not be served from it.

## Change

`agent_resolve_auth` now resolves in this precedence:

1. **vault** — the turn's attested principal, then the autonomous **server principal** (the path that serves a thin-client TCP/TLS conn with no per-user identity)
2. client-pushed session key (legacy RAM keyring — **retired in P4b**)
3. `agent->api_key` literal / `$ENV` ref
4. provider env / `$ENV` fallback (**kept by design** — roundtable D11)

- New `agent_vault_get()` helper (codex token + account id); api_key uses `vault_service_inject_api_key` (handles the dual-wrap). Locked/miss falls through.
- `agent_build_extra_headers` + `agent_has_resolvable_credentials` consult the vault first too.

## Non-breaking by design

Nothing is removed here. The vault is **added as the top tier**; session-creds + env remain below, so any path not yet migrated still resolves. **P4b** then deletes the legacy client-held push path (`cli_session_creds_prime`, `/v1/session/credentials`, the `session_credentials` store).

## Tests

`make -j1 verify-local` green. Shared `tests/support/vault_service_stub.c` resolves the three new `vault_service` read symbols (NO_ENTRY miss) for the ~36 keyless test binaries that link `agent_config.o`; wired into `TEST_WORKSPACE_OBJS_EXTRA` + the `unit-test-server-http` link line. Binaries needing real/controllable vault behavior (`unit-test-vault-service`, `unit-test-server-compute`) keep their own objects.

## Deploy note

No live impact until deployed (user-gated). Once `.254` upgrades, agents resolve from the vault; populate it via `agent key import` (over UDS) or native-TLS provisioning (PR #304) first.

### DIFF
--- a/src/server/agent_config.c
+++ b/src/server/agent_config.c
@@ -3,6 +3,7 @@
 #include "util.h"
 #include "agent_config.h"
 #include "vault_principal.h" /* VAULT_PRINCIPAL_MAX for the per-turn vault principal */
+#include "vault_service.h"   /* vault_service_* : the permanent credential store (P4) */
 #include "session_credentials.h"
 #include "model_registry.h"
 #include "platform_path.h"
@@ -203,6 +204,8 @@ static int agent_provider_env_value(const char *provider, char *dst, size_t dst_
    return 0;
 }
 
+static int agent_vault_get(const char *agent_name, const char *cred, char *out, size_t out_len);
+
 int agent_has_resolvable_credentials(const agent_t *agent)
 {
    if (!agent)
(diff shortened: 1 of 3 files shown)


JBailes:
## Roundtable resolution (security + engineer, jobs 175/176)

Fixed in `ea6699f`:
- **ENG#2** (wasted decrypt): the `have_primary_key` vault read moved past the codex-oauth / auth_cmd branches — only key-bearing auth types pay for it.
- **SEC#3** (style): codex account header now uses an explicit vault-miss test, not the bare-`;` consume pattern.
- **SEC#1 (HIGH)** — server-principal fallback: this is **by design**, now documented. The server vault holds the operator's shared default keys; aimee-server is a **single-owner** personal agent; the server bearer already gates every `/v1` op; and a per-user (`webuser:`/`uid:`) entry **overrides** the fallback. So serving the server key on a per-user miss is the intended "all agents, all connections" behavior, reachable only by bearer-authenticated turns — not a cross-tenant leak. No behavior change.
- **ENG#5**: noted `agent_has_resolvable_credentials` probes the codex **token** (an account without a token is not usable).

Refuted by code the sandboxed reviewers could not read:
- **ENG#1/#8** (empty-principal guard): `vault_service_inject_api_key` is robust to an empty/NULL principal (server-wrap miss → get miss → server-principal hit) and enforces **D15** (locked stays a hard error; only `NO_ENTRY` falls to the server vault). No guard needed.
- **ENG#6**: `agent_t.auth_type` is a fixed array, never NULL.
- **SEC#5 / ENG#4** (stub duplicate-symbol): the green link is the proof — no binary links both the shared stub and the real `vault_service.o` / `test_server_compute.o` stubs (`unit-test-vault-service` and `unit-test-server-compute` use custom link lines, not `TEST_WORKSPACE_OBJS_EXTRA`).

Noted, not actioned: **SEC#2** (vault outranks a configured key — intended for "vault is the source"; a stale entry is the operator's to clear) and **SEC#4** stack-zeroization of the transient key buffer (pre-existing pattern; the secret is short-lived stack, the delegate path already `OPENSSL_cleanse`s its copy).

PR #306, opened 2026-06-14
feat(vault): retire the client-held credential path entirely (P4b)

JBailes:
## What

Retire the **client-held credential path** entirely. The credential vault is now the single, permanent credential store; with P4a (#305) the vault is the universal source in `agent_resolve_auth`, so the legacy per-session push is dead weight. This removes it end to end — the final packet of the consolidation.

## Removed — server
- `/v1/session/credentials` endpoint (`handle_session_credentials`) + route + dispatch entry + auth capability.
- `session_credentials.{c,h}` (the in-RAM session keyring) and every reader tier in `agent_config.c`. Resolution is now purely **vault → stored api_key/$ENV**.
- The request-body codex push (`codex_oauth_token`/`codex_account_id` in `server_compute`/`server_compute_async`), replaced with an explicit per-turn clear (the vault delegate path sets codex creds from the vault).

## Removed — client
- `cli_session_creds_prime` (the push) + its 3 call sites (chat TUI + delegate/roundtable marshallers).
- `cli_attach_codex_creds` (client codex push; was already callerless).
- `cli_agent_add_localize_key`: `agent add --key` against a remote now **forwards** the key so the server vaults it over an attested (native-TLS #304 / UDS) connection, or **refuses** over plaintext (D2b) — never localized to `agent-keys.json`.

## Kept
- `aimee agent key import` (+ `cli_agent_keys_load/set`): the migration tool that reads any leftover `agent-keys.json` into the vault.
- `$ENV` / provider-env resolution as a documented fallback (P4a decision).
- `agent_set_request_codex_creds` + `g_request_session_id` (vault delegate path / fan-out creds snapshot).

## Wiring
Dropped `session_credentials.c` from Makefile + CMakeLists + test object lists; removed the `unit-test-session-credentials` target + test; removed the openapi `/session/credentials` path and regenerated `cli_v1_routes_gen.inc`; dropped the dead test stubs.

## Test
`make -j1 verify-local` green (full build + unit suite + v1 method-coverage + schema/charter checks).

## Deploy note
No live impact until deployed (user-gated). After this, a deployed `.254` resolves credentials **only** from the vault (+ `$ENV`). Sequencing: deploy server+client, then populate the vault — `agent key import` over UDS, or native-TLS provisioning (#304) — before relying on agents. The legacy push is gone, so an un-migrated key will not resolve.


JBailes:
## Roundtable resolution (security 177 + engineer 178)

Both reviewers were sandbox-blind (couldn't read the worktree; truncated diff) — all findings were "unverified, confirm X." Verified against the real tree:

**Confirmed already-correct (no change):**
- **SEC#1 (HIGH-if-absent)** — `agent add --key` over plaintext TCP: the server *does* refuse. `handle_agent_add` routes the server-vault write through `vault_capability_server_write_allowed`, which refuses `ATTEST_TCP_BEARER`, and returns an error *before* any config save — so the key is never persisted to `agents.json`. Over TLS/UDS it vaults. (P2a/P4a.)
- **SEC#2 / cli_tui** — codex chat now resolves from the vault (then the server's on-disk `~/.codex/auth.json` fallback); the client no longer pushes. Intended + documented in the deploy note; codex was consolidated server-side in P1 (setup-against-remote).
- **SEC#4 / ENG dangling-symbols** — ran the requested tree-wide grep: **no code references** to any removed symbol remain (only comment text), and the green link proves no undefined references.
- **ENG#2** — `agent_build_extra_headers` `(void)agent_vault_get(...)`: the helper guarantees `acct` empty on miss and the `if (acct[0])` guard below protects; behavior unchanged.
- **ENG#3b / ENG#4** — OpenAPI spec drift: checked — `codex_oauth_token`/`codex_account_id`/`cred_session_id` were **never** documented on the chat/delegate schemas, so there is nothing to remove. The only documented path (`/session/credentials`) is deleted.

**Fixed (`a21dc80`):**
- **ENG#1** — `agent_config.h` comments still described the retired client-push keyring; refreshed to state the vault is the single source and `g_request_session_id` now only carries session identity for fan-out workers (the variable is intentionally retained for the creds snapshot).

**ENG#3a** confirmed: `delegate_credential_retry` (from the vault) is the only remaining writer of the per-turn codex creds; the per-turn `agent_set_request_codex_creds(NULL,NULL)` clear prevents any stale-thread carryover.

PR #307, opened 2026-06-14
fix(delegate): inferred modality caps are soft, not fleet-wide hard fails

JBailes:
## Problem

Delegate capability requirements are inferred from the **prompt text** (`delegate_infer_capability_requirements`): a prompt that merely *mentions* `.png`/`screenshot`/`image_url`/`.pdf`/an audio extension was made to **hard-require** vision/pdf/audio. Since the prompt is the only signal (no real image payload is attached to these text delegates), it over-triggers constantly — and because no configured model advertises `vision`, the **whole fleet** got rejected:

```
no configured model supports required capabilities (caps=tools,vision, min_context=7773)
```

Observed live: jobs 179–186 (a "document all config options" review) all hard-failed this way.

## Fix — inferred modality caps are best-effort

- New `MODEL_CAP_MODALITY_SOFT` = vision|pdf|audio.
- `delegate_filter_route_capabilities`: if the full caps would disable every role candidate but the **hard** caps (tools) + `min_context` alone keep some, relax the modality caps and route on the hard set (logs a warning). The per-agent predicate is factored into `agent_meets_filter()` so the relaxation dry-runs before mutating `ag->enabled`.
- `agent_route_with_caps`: same fallback (retry without soft caps); body renamed `_inner`, public fn wraps it — keeps both routing gates in agreement.
- Drop the markdown-image (`![`) vision trigger: it appears in any doc/diff under review and never means the model must decode an image.

`tools` + `min_context` stay **hard** (real constraints); genuine modality routing still works (and is preferred) when a capable model is configured.

## Tests
`make -j1 verify-local` green. New: relaxation keeps text models when an unmet VISION is required alongside a satisfied TOOLS; a truly-unmet hard cap (tools) still fails closed. Existing filter/inference tests unchanged.

### DIFF
--- a/src/headers/model_registry.h
+++ b/src/headers/model_registry.h
@@ -30,6 +30,14 @@ enum
    MODEL_CAP_STREAMING = 1 << 5,
 };
 
+/* Modality capabilities (vision/pdf/audio) are INFERRED from prompt TEXT
+ * (delegate_infer_capability_requirements) — a heuristic that over-triggers on a
+ * text task merely mentioning an image/pdf/audio filename. They are therefore
+ * BEST-EFFORT routing preferences: when requiring them would disable every
+ * candidate, the router relaxes them and routes on the hard caps (tools) +
+ * min_context instead, rather than hard-failing the whole task fleet-wide. */
+#define MODEL_CAP_MODALITY_SOFT (MODEL_CAP_VISION | MODEL_CAP_PDF | MODEL_CAP_AUDIO)
+
 typedef struct
 {
    char provider[MODEL_PROVIDER_MAX];
--- a/src/server/agent_config.c
+++ b/src/server/agent_config.c
@@ -1354,8 +1354,9 @@ static int agent_satisfies_required_caps(const agent_t *ag, unsigned required_ca
 /* Route to the cheapest capable agent, filtering by required capability flags and minimum context
  * window when sys_cfg->model_meta_capability_routing is enabled.  Falls back to plain agent_route
  * when capability routing is disabled. */
-agent_t *agent_route_with_caps(agent_config_t *cfg, const char *role, const config_t *sys_cfg,
-                               unsigned required_caps, int min_context)
+static agent_t *agent_route_with_caps_inner(agent_config_t *cfg, const char *role,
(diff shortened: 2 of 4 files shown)


JBailes:
## Roundtable resolution (job 201, sanitized prose review)

> Note: the live fleet **could not review the raw diff** — jobs 197/198 failed with exactly `caps=tools,vision,pdf` because the diff text contains `.png`/`screenshot`/`image_url`/`.pdf` (the strings being edited), tripping the very bug this PR fixes on the old deployed code. So this was reviewed via a sanitized prose description.

**Addressed:**
- **(a) test the wrapper** — added `test_agent_route_with_caps_relaxes_soft_modality`: a mistral agent (TOOLS, no VISION in the registry) still routes when `TOOLS|VISION` is required, and the hard TOOLS cap is still enforced through relaxation. Plus the filter-level relaxation + hard-fail tests.
- **#1 predicate/mutator consistency (MEDIUM)** — by construction: both the dry-run count and the mutate pass call the **same** `agent_meets_filter()` with the same `enabled` + role checks. The existing drops-deprecated / min-context / tools tests confirm the non-relaxed path is unchanged.
- **#2 double side-effects (MEDIUM)** — `agent_route_with_caps_inner` is pure (selection + `clock_gettime` only; no log/metrics/DB), so the single relaxation retry is safe.
- **#4 retry state (LOW)** — the retry is a plain local `r` with one conditional call to `_inner` (never the wrapper); no recursion, no module/global state.
- **#3 observability (LOW)** — `delegate_filter_route_capabilities` logs a `WARN` naming the dropped soft caps on every relaxation (the dispatch path runs the filter before the router, so one log per request — no double-log).

**Verified fail-closed (#5/#6):** an unmet **hard** cap (no tool model / context too small) still errors in the filter and returns NULL in the router — only the inferred modality caps relax.

**Out of scope (#7, noted as follow-ups, not blockers):** JSON-mode/structured-output placement in the cap vector, and escalating vision to *hard* for OCR/transcribe prompts. With the current fleet (no vision model at all), even a true vision task can only be served by relaxing to text + the WARN; refining the heuristic is a separate change.

PR #312, opened 2026-06-15
fix(client): native-TLS /v1 dispatch on Linux (tls: endpoint)

JBailes:
## Problem

The Linux thin client can't speak **native TLS** to the server's `/v1` surface. `cli_v1_send` routes a configured remote through `cli_http_request` (plaintext `tcp:` or UDS) and only uses the TLS-capable `aimee_client_request` on **Windows** (`#else` branch). So `aimee agent key import` — and every `/v1` dispatch — over the server's native-TLS listener (#304) fails on Linux with:

```
could not reach the aimee-server /v1 endpoint
```

This broke the documented **TLS credential-provisioning flow** (the only network path allowed to write the server-principal vault). Discovered while deploying the vault consolidation to the live server, where I had to provision via raw `python+ssl` POSTs as a workaround.

## Fix

- **`cli_http_request` (posix) is TLS-aware**: a `tls:host:port` endpoint wraps the connected socket in `aimee_tls` (under `WITH_TLS`). Send/recv factored into `cli_http_txn` + `cli_conn_write_all`/`cli_conn_read` so one request/response loop serves both plaintext and TLS (mirrors `aimee_client.c`'s io abstraction).
- **`cli_http_connect`** accepts the `tls:` scheme (connects the same TCP socket; the caller does the handshake).
- **`cli_rpc_client_endpoint`** normalizes `https://` → `tls:host:port` (and `http://` → `tcp:`); the `--server`/`AIMEE_SERVER_URL` fallback selects `tls:` vs `tcp:` by scheme via the new `aimee_client_remote_active_scheme()`.
- Without `WITH_TLS`, a `tls:` endpoint fails cleanly.

## Tests / verification
- `make -j1 verify-local` green. Tests that link `posix/cli_client.o` now also link `aimee_tls.o` (it references TLS under `WITH_TLS`; `L_FULL` already provides `-lssl/-lcrypto`).
- **Live**: built `WITH_TLS`; `agent list` and `agent key import` over `tls:<server>:8742` (insecure-verify + bearer) reach the native-TLS listener and vault all 4 keys — the same calls previously failed on Linux.

## Follow-up
Native-TLS provisioning is now first-class on Linux: `AIMEE_API_ENDPOINT=tls:host:port` (or `https://`) + `AIMEE_API_BEARER` + `AIMEE_TLS_INSECURE` for a self-signed cert.

### DIFF
--- a/src/aimee_client.c
+++ b/src/aimee_client.c
@@ -306,8 +306,10 @@ static char *tcp_request(const char *url, const char *token, const char *method,
    return out;
 }
 
-int aimee_client_remote_active(char *desc_out, unsigned long desc_sz)
+int aimee_client_remote_active_scheme(char *desc_out, unsigned long desc_sz, int *is_https_out)
 {
+   if (is_https_out)
+      *is_https_out = 0;
    char url[512], tok[256];
    if (!resolve_remote(url, sizeof(url), tok, sizeof(tok)))
       return 0;
@@ -316,11 +318,20 @@ int aimee_client_remote_active(char *desc_out, unsigned long desc_sz)
       char host[256], port[16];
       int is_https = 0;
       if (parse_url(url, host, sizeof(host), port, sizeof(port), &is_https) == 0)
+      {
          snprintf(desc_out, desc_sz, "%s:%s", host, port);
+         if (is_https_out)
+            *is_https_out = is_https;
+      }
       else
          snprintf(desc_out, desc_sz, "%s", url);
    }
    return 1;
+}
+
+int aimee_client_remote_active(char *desc_out, unsigned long desc_sz)
+{
+   return aimee_client_remote_active_scheme(desc_out, desc_sz, NULL);
 }
 
 int aimee_client_remote_token(char *tok_out, unsigned long tok_sz)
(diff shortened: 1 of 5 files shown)


JBailes:
## Roundtable resolution (jobs 227/228)

Both reviewers were **sandbox-blind** — the 445-line diff exceeded the delegate's file-load budget (the `[file reference limit reached]` markers), so neither could read the actual change. (Notably both jobs *completed* rather than stalling — the deployed #309 heartbeat fix at work.) Verified the security notes against the real code:

**Fixed (`<this commit>`):**
- **#2 userinfo** — `cli_rpc_normalize_endpoint` kept `user:pass@host` in the host:port; now stripped (credentials belong in the bearer).

**Verified already-correct (no change):**
- **TLS verification is ON by default** (`aimee_tls.c`: `SSL_VERIFY_PEER` unless `AIMEE_TLS_INSECURE`); insecure is strictly opt-in.
- **close ordering** — `aimee_tls_free` (which does *not* close the fd) runs before `close(fd)`; no read-from-closed-fd race. `cli_http_txn` never closes the fd (centralized in the caller).
- **fail-closed** — a `tls:` endpoint without `WITH_TLS` returns NULL, never silently downgrades to plaintext.
- **scheme passthrough** (`tcp:`/`unix:`/bare `host:port`) is unchanged existing behavior.

**Coverage note:** the `tls:` branch is proven by a live end-to-end test (`agent list` + `agent key import` over `tls:<server>:8742` against the deployed v0.2.71 — both previously failed on Linux). A unit test for the branch would need a TLS echo server (follow-up).

PR #390, opened 2026-06-16
docs(proposals): correct ingress-cost delegate-pricing note (no external dep)

JBailes:
Per operator clarification: aimee's delegates are **subscription-based**, so their per-token price is **0** — which is exactly what the #339 known-zero rows encode (priced=1 "free", not "unknown"). There is **no missing price data** and **no external dependency**.

Delegate pricing is already **per-delegate configurable on its model** through the single pricing authority — `model_registry` `cost_in_per_mtok`/`cost_out_per_mtok` (operator/models.dev override) → the `token_tracker_registry.c` bridge → `token_estimate_cost`, where a **nonzero registry price overrides the static 0** (`token_tracker.c:187`). So a metered delegate's price is set without touching code, and there is **no second pricing source** (agents.json carries no price; price lives only in the registry, per §1's "one pricing source of truth").

This corrects the proposal's previously-noted "price data" item — it was a framing error, not a real dependency. Updated `## §8` + the State/status lines; docs-only, `make lint` ok.

Net: `ingress-cost-accounting` is now design-ready with **zero external dependencies**.

### DIFF
--- a/docs/proposals/pending/ingress-cost-accounting-and-optimizations.md
+++ b/docs/proposals/pending/ingress-cost-accounting-and-optimizations.md
@@ -2,9 +2,11 @@
 
 - **State:** reviewed — design-ready (2026-06-16). §1–§2 foundations landed
   (#185/#339); the design roundtable's 12 blockers on the remaining §2–§7 work are
-  resolved in §8. Implementation-ready; sole external dependency is real
-  local-delegate price data (operator). (Consolidated after PR #180 review + twelve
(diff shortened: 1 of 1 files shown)


JBailes:
Superseding: per operator decision, delegate/model cost is stored in a DB1 server-owned price table (not the model_registry). Folding the (still-valid) 'no external dependency / subscriptions=0' point into the DB1 implementation PR.

PR #436, opened 2026-06-17
feat(mtls): aimee cert CLI + /v1/cert/issue|list|revoke (slice 2b)

JBailes:
## mTLS slice 2b — operator cert lifecycle CLI + routes

Third slice of the [mTLS client-identity proposal](docs/proposals/pending/mtls-client-identity.md). Builds on slice 1 (`cert:<CN>` principal, #434) and slice 2 (PKI CA + issue/revoke/list, #435).

### What
Exposes the client-cert lifecycle as an operator-gated `/v1` surface and `aimee cert` CLI:

| Route | Body | Returns |
|---|---|---|
| `POST /v1/cert/issue` | `{cn, days?}` | `{cn, serial, cert, key}` — **key returned once** |
| `POST /v1/cert/list` | `{}` | `{certs:[{serial,cn,issued_at,expires_at,revoked}]}` |
| `POST /v1/cert/revoke` | `{serial}` | `ok` |

`server_cert.c` adapts the slice-2 `pki_*` primitives. `pki_list()` added to enumerate issued certs.

### Authorization (defence in depth)
- `cert_op_allowed()` admits **only** a kernel-attested local UDS peer (`ATTEST_UDS_PEERCRED`) or a confidential native-TLS+bearer connection (`ATTEST_TLS_BEARER`). A plaintext TCP bearer and an mTLS *client* are refused — clients must not mint/revoke certs.
- Route table additionally gates all three at `CAP_DELEGATE`.
- Private key is delivered exactly once and `OPENSSL_cleanse`'d from the handler buffer after send.

### Client
`aimee cert issue|list|revoke` registered in the help table (TIER_ADMIN) and the `/v1` route table; marshals to the new endpoints.

### Verification
- Builds clean: `aimee-server`, `aimee-kb`, `aimee` client.
- `unit-test-server-dispatch` links the new handlers (shared stub) and passes.
- `unit-test-pki` covers CA/issue/revoke/list.
- All sync/lint green: `dispatch-caps`, `cli-v1-routes`, `server-api-conformance` (249 routed), `v1-method-coverage`, `line-check`, `build-integrity`.
- Live server proven healthy locally (HTTP `/v1` listener comes up); full operator e2e is gated to the .254 testbed per the standing workflow (local harness cannot host a long-running server).


JBailes:
## Roundtable review (2 lenses: mistral + minimax-M3, uncapped)

Both lenses returned CHANGES-REQUIRED; every BLOCKER/MAJOR was adversarially verified against the code and found **false**, with two real MINOR hardenings applied.

**mistral** — 1 MAJOR + 3 MINOR, all false:
- "key not cleansed if `cJSON_CreateObject()` fails" → the cleanse is already present on that exact error path (`server_cert.c:50`) and the success path (`:58`).
- cert_list_cb leak / pki_list bounds / cert+serial cleanse → no leak path; `cJSON_AddStringToObject` strdup's safely; cert/serial are public.

**minimax-M3** — 2 MAJOR, both false (verified):
- *MAJOR: key leak on `pki_issue` failure path* — `pki_issue` zeroes `key_pem[0]` at entry (`pki.c:359`) and writes key material only inside the success block (`:391`), immediately before `rc=0`. On any failure the caller's `key` buffer holds only `'\0'`. (`char key[4096] = ""` also zero-inits the whole array, not just byte 0.)
- *MAJOR: `pki_list` missing `g_mu`* — `g_mu` guards the in-memory revocation snapshot `g_revoked[]`, **not** DB access: `pki_revoke` writes the DB outside the lock (`pki.c:436–447`). `pki_list` touches only the DB, so omitting the lock is correct and consistent.

**Applied (real, minor):**
- `handle_cert_issue`: reject `days` outside `[1,3650]` before the `double→int` cast (UB / absurd-validity guard).
- `handle_cert_list`: surface a `pki_list` DB error instead of masking it as an empty `{certs:[]}`.

Commit: d0af737. Build + `unit-test-server-dispatch` + `unit-test-pki` green locally.

PR #437, opened 2026-06-17
feat(mtls): client presents tls/client cert — OpenSSL/Linux leg (slice 3)

JBailes:
## mTLS slice 3 — client-cert presentation (WP-D), OpenSSL/Linux leg

Final functional slice of the [mTLS client-identity proposal](docs/proposals/pending/mtls-client-identity.md). Completes the loop: server verifies client certs (slice 2, #434/#435), operator issues them (slice 2b, #436), and now the **client presents** its cert so it's identified as a distinct `cert:<CN>` principal.

### What
`aimee_tls.c` (OpenSSL backend): before `SSL_connect`, if both `<aimee_home>/tls/client.crt` and `client.key` exist, load them as the client certificate (`SSL_CTX_use_certificate_chain_file` + `SSL_CTX_use_PrivateKey_file` + `check_private_key`). Absent → plain client TLS (unchanged behaviour).

### Security: fail closed
A group/world-readable private key (`& 077`) is **refused** and never presented — the key is this client's identity material. The eligibility decision (exists + `0600`) is factored into the pure, exported `aimee_tls_client_cert_eligible()` and unit-tested:

- `unit-test-aimee-tls-clientcert`: present-on-0600 (paths resolved), **refuse-on-0644**, none-when-absent, cert-without-key→none, bad/NULL home→none.

### Deferred to slice 3b (Windows/macOS)
Presenting a PEM client cert on **Schannel** (EC-key import into a CryptoAPI `CERT_CONTEXT`) and **Secure Transport** (`SecIdentityRef` from PEM) is intricate and must be validated on real hardware — exactly how the native-TLS backends themselves were validated. Writing it blind would be untestable here. Each native backend carries an in-code marker at the precise hook point; the Linux leg lands first because it is the testbed for the whole mTLS feature.

### Verification
- Linux client builds clean with `WITH_TLS=1` (new `aimee_tls.o`).
- `unit-test-aimee-tls-clientcert` passes; registered in the suite.
- `line-check`, `docs-gen-check` green. Proposal WP-D updated to reflect the 3/3b split.
- Full mTLS e2e (present → server identifies `cert:<CN>` → revoked cert rejected) is gated to the live testbed per the standing workflow.

Stacked logically on #436 (slice 2b) but textually independent (different files).

### DIFF
--- a/docs/proposals/pending/mtls-client-identity.md
+++ b/docs/proposals/pending/mtls-client-identity.md
@@ -95,13 +95,22 @@ extension of an existing listener, not a new one.
 ### WP-D — Client: present a client certificate
 The thin client loads its identity (`<aimee_home>/tls/client.{crt,key}`) and
 presents it via each `aimee_tls.h` backend:
-- OpenSSL (Linux): `SSL_CTX_use_certificate_file` + `SSL_CTX_use_PrivateKey_file`.
+- OpenSSL (Linux): `SSL_CTX_use_certificate_chain_file` +
+  `SSL_CTX_use_PrivateKey_file`. **DONE (slice 3).** Presented automatically when
+  both files exist; a group/world-readable key is refused (fail closed). Gate is
+  unit-tested (`unit-test-aimee-tls-clientcert`).
 - Schannel (Windows): supply the client cert in `SCHANNEL_CRED.paCred` (imported
-  from a PFX or the user cert store).
+  from a PFX or the user cert store). **Deferred → slice 3b** (EC-key import via
+  CryptoAPI must be validated on real Windows, as the native backend itself was).
 - Secure Transport (macOS): `SSLSetCertificate` with a `SecIdentityRef` from a
-  PKCS#12.
+  PKCS#12. **Deferred → slice 3b** (identity construction validated on real macOS).
+
 This is the piece the native-TLS proposal explicitly deferred (mTLS was out of
-scope there); it is additive, gated on the client actually having a cert.
+scope there); it is additive, gated on the client actually having a cert. The
+Linux leg lands first because it is the testbed for the whole mTLS loop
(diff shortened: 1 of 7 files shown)


JBailes:
## Roundtable review (mistral + minimax-M3, uncapped)

**minimax-M3: VERDICT APPROVE** — "None of the five categories contain a BLOCKER or MAJOR. The fail-closed security property holds." It explicitly confirmed the threat-model framing: TOCTOU/symlink are "only exploitable if an attacker can already plant files in `tls/`, so it's hardening rather than a true bypass."

**mistral: CHANGES-REQUIRED** — every BLOCKER/MAJOR adversarially verified and found false or already-handled:
- *BLOCKER TOCTOU / symlink* → out of threat model (writer to your own `0700 tls/` already reads the key directly). The symlink leg is nonetheless **hardened**: `lstat` + `S_ISREG` now refuses a symlinked/non-regular key (commit 9d2cb87) + test.
- *MAJOR `SSL_CTX` leak on `SSL_new()` failure* → **false**: the original path already `SSL_CTX_free`s.
- *MAJOR buffer overflow on long home* → **false**: `snprintf` is bounded; truncation degrades safely to "no cert".

**Applied from the round:**
- `lstat` + `S_ISREG` symlink/non-regular-file rejection on the key, with a dedicated test (9d2cb87).
- Pinned the exact `& 077` boundary at `0640` group-readable, not just `0644` (19cc124).

**Deliberately not applied (documented):** full TOCTOU-atomic load (out of threat model) and an `st_uid == getuid()` ownership check (would false-reject legitimate root-owned / shared-key deploys; a non-owned key OpenSSL can't read already fails safe).

Gate now pins: present-on-0600, refuse-0644, refuse-0640, refuse-symlink, none-when-absent / cert-without-key / bad-home.

PR #440, opened 2026-06-17
feat(mtls): native client-cert presentation — Windows Schannel + macOS (slice 3b)

JBailes:
## mTLS slice 3b — native-backend client-cert presentation (WIP)

Completes WP-D by presenting `<aimee_home>/tls/client.{crt,key}` on the native TLS backends, matching the OpenSSL/Linux leg (slice 3, #437).

### Windows (Schannel) — this commit
- PEM cert → DER → `CertCreateCertificateContext`.
- PKCS#8 EC key (what `PEM_write_bio_PrivateKey` emits) → `NCryptImportKey(NCRYPT_PKCS8_PRIVATE_KEY_BLOB)` into an ephemeral CNG store → bound to the cert via `CERT_KEY_CONTEXT_PROP_ID`.
- `SCHANNEL_CRED.cCreds/paCred` set when loaded; absent → plain client TLS; load failure → cert-less handshake (mTLS server rejects).
- Key buffers `SecureZeroMemory`'d; cert + CNG handles freed in `aimee_tls_free`. Links `ncrypt`. Defensive `_WIN32_WINNT 0x0A00` + `#ifndef` CNG-symbol fallbacks for older MinGW headers.

### macOS (Secure Transport) — follow-up commit in this PR
Pending (`SecPKCS12Import` path — see comments).

### Validation
- **Windows: runtime-validated on a real Windows VM** (mTLS handshake presents the cert → server identifies `cert:<CN>`), mirroring how the native backends were originally validated.
- **macOS: compile-only via CI `macos-build`** — no Mac available for a runtime mTLS handshake; flagged explicitly.

PR #442, opened 2026-06-17
feat(mtls): macOS Secure Transport client-cert presentation (slice 3c)

JBailes:
## mTLS slice 3c — macOS Secure Transport client-cert presentation

Final native leg of WP-D: the macOS thin client presents `<aimee_home>/tls/client.p12` so the server identifies it as a distinct `cert:<CN>` principal. Completes the mTLS feature across all three backends (OpenSSL/Linux #437, Windows/Schannel #440, macOS now).

### Implementation
`SecPKCS12Import` → `SSLSetCertificate`. Export the PEM `client.{crt,key}` (used by the other backends) to `client.p12` via `openssl pkcs12 -export`; empty passphrase unless `AIMEE_TLS_CLIENT_P12_PASS`.

### Addresses the slice-3b roundtable HIGH (key leak)
Plain `SecPKCS12Import` imports the identity into the user's **login keychain**, persisting the private key across runs. This impl imports into a **transient, per-process keychain** (`SecKeychainCreate` under `TMPDIR`, referenced via `kSecImportExportKeychain`), removed with `SecKeychainDelete` on `aimee_tls_free` — and every connect error path now routes through `aimee_tls_free`, so the keychain is cleaned on failure too. The key never touches the login keychain. PKCS#12 buffer wiped with `memset_s`.

### Validated on a real Mac (CI)
The `macos-build` job now runs an **mTLS loopback test**: a test EC-P256/PKCS#8 client cert → `client.p12`, `openssl s_server -Verify` (mandates a client cert), then `aimee --server https://… config show` with the identity present. It asserts:
- `s_server` logs **`CN=ci-macos-client`** (cert presented + chain-verified),
- no connection shows "peer did not return a certificate",
- **no transient keychain is left behind** (leak check).

This mirrors the runtime validation that caught the Windows ephemeral-key bug.

Net diff: `aimee_tls_securetransport.c`, the `macos-build` CI step, and a one-line generated-docs refresh for `AIMEE_TLS_CLIENT_P12_PASS`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,6 +321,10 @@ jobs:
             -CAcreateserial -days 1 -out "$H/tls/client.crt"
           openssl pkcs12 -export -inkey "$H/tls/client.key" -in "$H/tls/client.crt" \
             -out "$H/tls/client.p12" -passout pass:
+          echo "::group::diagnostics"
+          openssl version
+          openssl pkcs12 -info -in "$H/tls/client.p12" -passin pass: -noout -nokeys 2>&1 | head -20 || true
+          echo "::endgroup::"
           # s_server REQUIRES a client cert (-Verify) and serves a status page.
           openssl s_server -accept 14443 -cert "$D/srv.crt" -key "$D/srv.key" \
             -CAfile "$D/ca.crt" -Verify 1 -www -naccept 3 > "$D/s_server.log" 2>&1 &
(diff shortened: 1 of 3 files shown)


JBailes:
## Runtime validation (real Mac) + roundtable

**Runtime-validated on GitHub's macОS runner** — the `macos-build` job's mTLS loopback (`openssl s_server -Verify` ↔ `aimee`) prints:
```
depth=0 CN = ci-macos-client
verify return:1
...
PASS: Secure Transport presented client cert; server verified CN=ci-macos-client; no keychain leak
```
Getting there surfaced two real macOS-only issues that only a real Mac reveals:
1. `SecPKCS12Import` fails `errSecAuthFailed (-25293)` on an **empty passphrase** → the `.p12` must be passphrase-protected (documented; supplied via `AIMEE_TLS_CLIENT_P12_PASS`).
2. The transient-keychain `SecPKCS12Import` → `SSLSetCertificate` path then presented the cert correctly, and the leak check confirms no `.keychain` is left behind.

**Roundtable (mistral + minimax):** both returned CHANGES-REQUIRED for the **same** finding — a pid-only transient-keychain path races for concurrent connections in one process — and recommended exactly a pid+counter suffix. **Already fixed** in commit 370fb11 (`__sync_add_and_fetch`); the reviewers saw the pre-fix diff. Every other category (keychain lifecycle, CF ownership, memory safety, fail-closed security) was rated CLEAN by both.

PR #466, opened 2026-06-17
feat(webchat-git): WP-I per-webuser code-server supervisor

JBailes:
## WP-I — per-webuser code-server supervisor (webchat-git Phase 2)

aimee-server owns the editor lifecycle because it owns both the workspace files **and** the sealed-vault credential-injection environment (WP-C/WP-C2). This is the supervisor half of the in-browser VSCode feature; the reverse-proxy + tab follow in WP-J.

### What it does
- **`webuser_editor.{c,h}`** — per-principal registry (`principal → pid/port/last_active`) with `ensure` / `touch` / `stop` / `reap_idle` / `shutdown`. Each webuser gets at most one code-server, bound to `127.0.0.1:<ephemeral port>`, rooted at their scoped workspace (`ws_scope_user_root`), launched with their vault-backed git env (`git_cred_inject_build_env`: `GH_TOKEN` + `GIT_ASKPASS` shim + in-memory ssh-agent `SSH_AUTH_SOCK`). The integrated terminal and Git view authenticate **without ever exposing a credential to the browser**.
- **`POST /v1/workspace/editor`** (CAP_TOOL_EXECUTE, webuser-only) → `{ok, port}`; `503` when unavailable. The loopback port reaches only webchat (the trusted reverse-proxy), never the browser.
- **`server_shutdown()`** reaps editors so they never outlive the server.

### Hardening
Children run hardened: own session (`setsid`), no core dumps, `PR_SET_NO_NEW_PRIVS`, stdio → `/dev/null`, `HOME` pinned to the user root (no state in the server's home), optional UID drop to a dedicated service user (`AIMEE_WEBCHAT_EDITOR_UID`, only when running as root — defence in depth ahead of WP-K).

### Fail-closed
Disabled unless `AIMEE_WEBCHAT_EDITOR` is set **and** a code-server binary is present → a **no-op on the default `WITH_VSCODE=0` images**.

### Testing
`unit-test-webuser-editor` covers the gating, identity guard (webuser-only), and bookkeeping. The live code-server spawn/jail is **deploy-gated** (CI carries no code-server binary on the default build) and validated on .254; the bookkeeping and fail-closed paths are unit-tested. Local: server links clean, `server-api-conformance-check` / `dispatch-caps-check` / `docs-gen` all green.


JBailes:
## Roundtable review (security + reviewer lenses, via mistral)

Ran two adversarial review lenses on the supervisor diff. Net outcome: **2 real findings fixed, rest dismissed as fork()-semantics misunderstandings.**

**Fixed (commit fc4e5d7):**
1. **Editor env leak (security)** — code-server gives the user an integrated *terminal*, so inheriting the server's full `environ` would expose secrets like `AIMEE_DB2_URL` (Postgres password) / `AIMEE_SERVER_TOKEN`. The child env is now built from a minimal curated base (`PATH`/`LANG`/`TERM`) with only the vault-backed git vars layered on — never the server environ.
2. **Zombie on reap (lifecycle)** — `reap_locked()` SIGTERM'd the child then zeroed the slot, so `sweep_dead_locked()` could never `waitpid()` it → permanent zombie per stop/idle. Now waits for the child (bounded, SIGKILL escalation) before clearing the slot.

**Dismissed (verified false):**
- *"use-after-free: parent frees `base` after fork while child holds pointers"* — after `fork()` the processes have separate COW address spaces; the parent freeing its own heap cannot touch the child's copy, and the child's `envp` stays valid through `execvpe` (same idiom `git_ssh_agent.c` already relies on).
- *"home_env stack pointer"* — valid through exec (reviewer conceded "technically safe").
- *SO_REUSEADDR / port false-positive / ECHILD* — moot for an ephemeral port-0 loopback bind; readiness already checks `waitpid` for child death.
- *port race* — bounded (3 attempts), detected via child-death in the readiness loop, default-off feature; not a hang.

PR #467, opened 2026-06-18
feat(webchat-git): WP-J /vscode reverse-proxy + Editor tab

JBailes:
## WP-J — /vscode reverse-proxy + Editor tab (webchat-git Phase 2)

The browser-side half of the in-app VSCode. aimee-server owns the per-user code-server (WP-I) on a loopback port; webchat ensures it exists and reverse-proxies `/vscode/*` to it. The loopback port never reaches the browser, and the editor's git credentials stay server-side (vault-only) — the user only ever talks to webchat's own origin.

### What it does
- **`webchat/vscode.go`** — `handleVSCode` resolves the user's editor port via `POST /v1/workspace/editor` (cached 30s, re-ensured on dial failure) and reverse-proxies to `127.0.0.1:<port>`. `httputil.ReverseProxy` transparently tunnels the **WebSocket** upgrade code-server uses for the terminal/editor sync. The webchat **session cookie + bearer are stripped** before forwarding (code-server runs `--auth none` behind our auth gate); `X-Frame-Options` is forced to `SAMEORIGIN` so the same-origin iframe frames reliably.
- **`server.go`** — register `/vscode` + `/vscode/` (proxy) and the `/editor` SPA route. Also adds the **`/projects` SPA route** — WP-F2 only added *client-side* routing, so a hard refresh on `/projects` 404'd; this fixes that latent gap.
- **frontend** — `Editor.tsx` embeds `/vscode/` in a same-origin iframe; Editor nav item + `/editor` route in `App.tsx`.

### Security / isolation
- Browser never sees the loopback port or any git credential.
- Same-origin iframe → no cross-origin credential exposure; framing allowed.
- code-server's user-data/extensions live under `<userroot>/.aimee-editor/` (dot-prefixed, so `git_project_list` skips it — no bogus project).
- Editor is server-gated (`AIMEE_WEBCHAT_EDITOR` + a `WITH_VSCODE` build); when off, `/vscode/` returns 503 and the tab shows that.

### Deployment assumption
webchat reaches code-server at `127.0.0.1:<port>`, i.e. they share a network namespace (the combined / aimee-server image). The editor feature is scoped to those images by design.

### Testing
`go build` / `go vet` / `go test ./webchat/...` and frontend `tsc -b && vite build` all green. The live proxy, WebSocket tunneling, and iframe framing/base-path are **deploy-validated on .254** — CI carries no code-server binary or browser. Path is forwarded unchanged (code-server derives its asset base from the sub-path); if a deploy shows base-path issues it's a one-line proxy tweak.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -7,6 +7,7 @@ import Chat from './pages/Chat';
 import Dashboard from './pages/Dashboard';
 import Workflows from './pages/Workflows';
 import Projects from './pages/Projects';
+import Editor from './pages/Editor';
 
 // A render error in any page used to throw past the root and unmount the whole
 // app, leaving a blank screen (the AppShell, nav, and other pages vanished too).
@@ -60,6 +61,7 @@ const NAV_ITEMS: NavItem[] = [
   { label: 'Dashboard', icon: '📊', route: '/dashboard', section: 'Main' },
   { label: 'Workflows', icon: '🔀', route: '/workflows', section: 'Main' },
   { label: 'Projects', icon: '📁', route: '/projects', section: 'Main' },
+  { label: 'Editor', icon: '🖥️', route: '/editor', section: 'Main' },
 ];
 
 function LogoutButton() {
@@ -133,6 +135,7 @@ export default function App() {
             <Route path="/dashboard" element={<Dashboard />} />
             <Route path="/workflows" element={<Workflows />} />
             <Route path="/projects" element={<Projects />} />
+            <Route path="/editor" element={<Editor />} />
             <Route path="*" element={<Navigate to="/chat" replace />} />
           </Routes>
         </ErrorBoundary>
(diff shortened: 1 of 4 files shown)


JBailes:
## Roundtable review (security + reviewer lenses, via mistral)

No blockers. **3 real fixes applied (commit 5311b76); rest dismissed with rationale.**

**Fixed:**
- **Port leak via slash-less redirect** — bare `/vscode` could let code-server answer with a redirect to `/vscode/` built from the upstream `Host`, leaking the loopback port and bouncing the browser off our origin. Now webchat 308-redirects `/vscode`→`/vscode/` itself, and `ModifyResponse` rewrites any `Location` pointing at the upstream host to a relative path.
- **Forwarded headers** — added `X-Forwarded-Host` (original host) + `X-Forwarded-For` so code-server builds correct external URLs for the proxied sub-path.
- **ErrorHandler robustness** — evict the cached port only when it still holds the *failed* port (was re-deriving `currentUser(req)`, which can be `""` mid-session and could clobber a freshly-stored entry).

**Dismissed (verified):**
- *"remove iframe `allow-same-origin`"* — code-server **requires** same-origin to function; removing it breaks the editor. The session cookie is `HttpOnly`+`SameSite=Strict` (a same-origin frame can't read it via JS), and cross-user isolation is **server-side** (per-UID editor, WP-I), not browser-sandbox. This is the correct model.
- *cache check-then-act race* — benign (keyed by username, self-healing) and now bounded by the port-match eviction.
- *global `X-Frame-Options` on other pages* — pre-existing, out of WP-J scope (noted as a follow-up).
- *WS FlushInterval / rate-limit / username regex* — `ReverseProxy` auto-handles the WS upgrade; identity is already a server-validated PAM session.

PR #468, opened 2026-06-18
feat(webchat-git): WP-K editor-env leak gate + no on-disk credential cache

JBailes:
## WP-K — editor-env leak gate + extension-host containment (webchat-git Phase 2, final)

The Phase-2 containment gate. Refactors the code-server launch env into a single fully-owned helper and **proves in a unit test** that nothing the user runs in the editor — integrated terminal or an untrusted extension — can read the server's secrets or persist the git token to disk.

### What it does
- **`webuser_editor_build_env()` / `webuser_editor_free_env()`** — build the editor env from a minimal curated base (`PATH`/`LANG`/`TERM`) + `HOME=<userroot>` + the user's vault-backed git env, **never** the server's own `environ`. Deep-copied with secret-zeroing teardown (replaces the prior dual-ownership `envp`/`base` split — also simpler/safer).
- **No on-disk credential cache** — adds `GIT_CONFIG_COUNT=1` / `credential.helper=""` so neither the terminal nor an extension can `git config credential.helper store` the token to `~/.git-credentials`. Auth still flows through the `GIT_ASKPASS` shim each time (vault-only at rest).
- **WP-K leak test** — with `AIMEE_DB2_URL` / `AIMEE_SERVER_TOKEN` / `ANTHROPIC_API_KEY` set in the process env, asserts the built editor env contains the curated base + `HOME` + credential-helper-off and **none** of those names or secret values (no `AIMEE_*`, no provider key, no leaked password).

### Containment model (honest)
The askpass shim does `printf "$GH_TOKEN"` — it reads the token **from the env**, so the token is necessarily present for git to work. Hiding it from the user's *own* extensions would require the socket-askpass that WP-C **deliberately dropped** in favour of **per-UID isolation**: another *user* cannot read the env (`/proc/<pid>/environ` is UID-restricted, and the editor can drop to `AIMEE_WEBCHAT_EDITOR_UID`). WP-K's boundary is therefore: **no unrelated server secrets in the editor env, and no on-disk credential persistence** — both now enforced and tested.

### Testing
`unit-test-webuser-editor` extended with the leak gate (passes); server links clean; existing `unit-test-git-cred-inject` / `unit-test-server-http` still green. The live extension-host behaviour is deploy-validated on .254.

---
This is the final work-packet of the webchat git-projects + in-app VSCode feature (WP-0,A,B,C,C2,D,E,F,F2,G,H,I,J,K).

### DIFF
--- a/src/headers/webuser_editor.h
+++ b/src/headers/webuser_editor.h
@@ -47,4 +47,17 @@ void webuser_editor_shutdown(void);
  * code-server binary is resolvable. Exposed for the route layer / tests. */
 int webuser_editor_available(void);
 
+/* Build the sanitized, fully-owned environment a webuser's code-server is
+ * launched with (webchat-git WP-K): a minimal curated base (PATH/LANG/TERM) +
+ * HOME=<userroot> + the user's vault-backed git env (GH_TOKEN/GIT_ASKPASS/
+ * SSH_AUTH_SOCK when present) + git's on-disk credential cache disabled. It NEVER
+ * contains the server's own environment, so the editor's terminal/extensions
+ * cannot read server secrets (AIMEE_DB2_URL, AIMEE_SERVER_TOKEN, provider keys).
+ * Returns a malloc'd NULL-terminated array (free with webuser_editor_free_env),
+ * or NULL on error. Exposed for the WP-K editor-env leak tests. */
+char **webuser_editor_build_env(const char *principal, const char *userroot);
+
+/* Free an envp from webuser_editor_build_env (zeroes any GH_TOKEN first). */
+void webuser_editor_free_env(char **envp);
+
 #endif /* WEBUSER_EDITOR_H */
--- a/src/server/webuser_editor.c
+++ b/src/server/webuser_editor.c
@@ -173,6 +173,92 @@ static int resolve_drop_uid(drop_ids_t *out)
    return 0;
 }
 
+void webuser_editor_free_env(char **envp)
+{
+   if (!envp)
+      return;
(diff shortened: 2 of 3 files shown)


JBailes:
## Review

The roundtable delegates (security + reviewer lenses) hit the **known write-guardrail misfire** — a diff that adds a new file (`--- /dev/null` header) or carries `.c` filename headers is misread as create-intent (`"named file ... does not exist"`), even with a no-tools prompt. Same class as the WP-H Dockerfile-diff misfire. Per that precedent I did a rigorous **self-review** of the security-sensitive code:

**Verdict: APPROVE, no blockers.**
- **OOM / double-free** — `out` is calloc-zeroed so `free_env` stops at the first NULL (frees only successful strdups); `inner`'s strings are independent allocations (strdup'd, never aliased into `out`), so freeing `out` then `inner` is disjoint. `git_cred_inject_free_env(NULL)` is a no-op.
- **Secret hygiene** — `inner`'s `GH_TOKEN` is zeroed right after `out` is built; `out`'s copy is zeroed at teardown; the child keeps its own COW copy for `execvpe` (intended). The server `environ` is never a source.
- **credential.helper** — `GIT_CONFIG_*` is applied at `-c`-level precedence (above config files); an empty `credential.helper` clears the helper list → no on-disk persistence.
- **Refactor parity** — HOME-pinning + no-creds fallback preserved; `spawn_code_server` updated to the single-owner `free_env`.

Build + leak test + `git_cred_inject`/`server_http` tests all green. The live extension-host behaviour is deploy-validated on .254.

PR #527, opened 2026-06-19
docs(autonomous-dev): correct webchat GUI claims (API-first)

JBailes:
Corrects the autonomous-development guide which overclaimed the webchat GUI. Reality: the Workflows tab is a read-only run viewer; submitting a proposal and recording gate approvals are API/operator actions today (submit form + in-tab gate approve/reject are in progress). Approvals are HMAC-signed with the operator key (delegate-unforgeable).


JBailes:
Superseded by #530: the autonomous-dev GUI (submit + gate approve/reject) is now implemented, so the doc describes the shipped state rather than the interim API-first correction. The B+C plan doc is carried into #530.

PR #590, opened 2026-06-21
Merge testing into main

JBailes:
Promote testing to main. Tree-equal to testing after resolving a squash-divergence conflict in frontend/src/pages/Chat.tsx (main received #582/#583 content via the squash promote #584; testing carries them as real commits plus more, so testing is a content-superset — resolved toward testing, verified identical).

Carries:
- #586 perf(webchat): window the transcript + isolate the stream tail to bound CPU over a session
- #585 feat(ingress): typed Envelope IR for pre-injection
- #583 docs(proposal): unified aimee-llm container
- #582 perf(webchat): instant (not smooth) stream auto-scroll

Frontend tsc + build green; merged tree verified identical to testing.


JBailes:
Cannot merge: main's required check allow-only-testing-source requires head=testing. Reconciling main into testing first, then promoting testing→main directly.

PR #654, opened 2026-06-23
feat(ingress): default claude_proxy_parity to ON

JBailes:
Flips the Claude Code `/v1/messages` ingress to exact Anthropic passthrough **by default** (`claude_proxy_parity` was opt-in, now opt-out). The model-swap + pre-injection proxy is now the `claude_proxy_parity: false` opt-out. `config_save` persists only the off override, per the default-on convention (cf. `memory_routing_enabled`).

**Verified end-to-end** against a real aimee-server + mock upstream:
- Default (no config key) → inbound model honored, `anthropic-version`/`anthropic-beta` forwarded upstream, `count_tokens` proxied to the real endpoint, upstream `429` + `Retry-After` relayed verbatim.
- `claude_proxy_parity: false` → model-swap restored (upstream gets the configured primary), betas dropped.

`make unit-tests` green; docs regenerated.


JBailes:
Superseded: per follow-up design discussion, parity will be auto-derived from whether the primary speaks the Anthropic API (no manual flag), as part of a broader dual-API gateway design.

PR #727, opened 2026-06-25
chore: merge main into testing (reconcile divergence before promotion)

JBailes:
`testing` was 2 behind `main` (prior testing→main merge commits #686/#689). That split history made the testing→main promotion PR (#725) `BEHIND`, so GitHub's test-merge 3-way-blended the two divergent generated `docs/gen/configuration.md` files into a state matching neither a clean regen — failing `docs-gen-check`.

Merge `main` into `testing` to rejoin the histories (no unique file content from main — pure history join), and confirm `make docs-gen` is clean on the merged tree. After this, #725 promotes cleanly (testing ⊇ main, authoritative current docs/gen).


JBailes:
Redundant — testing already merged main (9888f81). The real fix is a docs/gen regen on current testing (separate PR).

PR #732, opened 2026-06-25
feat(code-index): source code indexing from the git default branch

JBailes:
## Index the canonical branch, not the working tree

Code indexing — the graph **and** the embeddings — now reads the repository's git **default branch** (e.g. `origin/main`), not the user's working tree, current checkout, or a feature branch they are mid-edit on. The kb's code view must be **stable and canonical**: a developer's WIP, an uncommitted edit, or a throwaway branch must never thrash the graph/embeddings or pollute what other sessions retrieve as "the code."

This implements §0.5 of `docs/proposals/pending/code-graph-intelligence.md` (added here), converged via a roundtable design review (R2).

### `code_collect_files_cb` source selection

- **Default-branch resolution** (`git_resolve_default_ref`): `symbolic-ref origin/HEAD` → repair once via `git remote set-head origin -a` → first existing of `origin/main`, `origin/master`, `main`, `master`. **No fall-through to current HEAD/working tree** — a git repo with no resolvable default branch is **skipped** with a diagnostic, never silently indexed from an unstable checkout.
- **Read mechanism**: one `git ls-tree -r -z` + one `git cat-file --batch` (oids fed from a temp file so our side only reads — no bidirectional-pipe deadlock), pairing content to path by sequence (newline-in-path safe). Same extension/skip-dir/size/binary filters as the worktree walk.
- **Opt-in** `AIMEE_CODE_INDEX_SOURCE=worktree` indexes WIP for a user who wants it (team-unsafe, documented); `default`/`auto` (default) use the chain. Non-git dirs → working-tree walk.
- **Code-vs-edit-tool split (invariant)**: indexing reads the default branch; blast-radius / staged-change tools read the working tree via their own reader and don't route through this collector.

### Test — `unit-test-code-collect`
Materializes throwaway git repos and asserts:
```
test_code_collect:
  test_default_branch_is_canonical: ok   # main indexed, feature-branch WIP + untracked NOT
  test_worktree_optin: ok                # AIMEE_CODE_INDEX_SOURCE=worktree sees WIP
  test_clone_resolves_origin_head: ok    # deleted origin/HEAD repaired -> indexes origin/master
  test_no_default_branch_skips: ok       # no default branch -> skip, count 0
  test_non_git_uses_worktree: ok
ALL PASS
```
`aimee-server` + `aimee-kb` build clean (`-Werror`), clang-format clean.

> Stacked on #731 (code-graph P1); the two touch disjoint files. Merge #731 first, or this rebases cleanly onto `testing` once it lands.

PR #733, opened 2026-06-25
feat(kb): Reciprocal Rank Fusion core for hybrid code retrieval (P2 §5)

JBailes:
## §5 hybrid retrieval — the scoring model

Implements the **RRF scoring model** from `docs/proposals/pending/code-graph-intelligence.md` §5 as a pure, self-contained module — the genuinely novel, load-bearing piece of hybrid graph+vector+memory retrieval.

The three retrieval signals (graph N-hop neighborhood, vector cosine similarity, memory recall) have **non-comparable** raw scores — hop counts vs cosine vs recency — so fusing by raw score is wrong. `kb_rrf_fuse()` fuses by **rank** (Reciprocal Rank Fusion):

```
fused(d) = Σ_signals  weight_s / (k + rank_s(d))      (rank 1-based, k≈60)
```

summed only over the signals that contain `d`. RRF needs no normalization/calibration, is robust to an absent signal (the candidate simply isn't in that list), and **rewards cross-signal consensus** — a candidate ranked modestly by two signals beats one ranked #1 by a single signal.

- Per-signal weights `w_s` scale each signal's contribution (config-tunable in the follow-up).
- **Deterministic** tie-break: structural-trust weight desc, then id asc — stable regardless of input signal order (proven in the test by permuting inputs).
- A fused candidate carries its max `structural_weight` across signals + a `signal_hits` count, for downstream display/threshold.
- Pure (no DB/network/caller-alloc), so it unit-tests standalone and is reused verbatim by the eventual `/v1/code/hybrid` route.

### Test — `unit-test-kb-rrf`
```
test_kb_rrf:
  test_single_signal_math: ok        # exact w/(k+rank)
  test_consensus_beats_single: ok
  test_weighting: ok
  test_absent_signal_robust: ok
  test_deterministic_tiebreak: ok    # permute inputs -> identical order
  test_structural_weight_max: ok
  test_truncation: ok
  test_bad_args: ok
ALL PASS
```
`aimee-kb` builds clean (`kb_rrf.c` added to `KB_SRCS`); `-Werror`, clang-format clean.

### Next increment (integration-tier, needs live embedder)
Embed the query → gather the three signal lists (`db2_entity_edge_*` graph / `pgvec_code_search` vector / memory recall) → `/v1/code/hybrid` route + MCP tool + config-tunable weights.

> Stacked on #732 → #731 (disjoint code files; only the proposal doc overlaps). Merges cleanly once the parents land.


JBailes:
Roundtable review (4-panel) run on the fusion core. Two real correctness issues fixed in `987de3f`: (1) the qsort comparator used an epsilon score band that violates strict-weak-ordering (UB) — now compares scores exactly (genuine ties are bit-identical and fall through to the deterministic structural/id tie-breaks); (2) NaN/Inf `k`/`weight` slipped past the `<= 0` checks — now guarded with `isfinite()`, plus a 64-bit candidate-count accumulator and the structural-weight comparison made overflow-safe. Test extended with NaN/Inf cases. Remaining 'blocking' panel items were misreads from a truncated brief (the module frees on all paths, initializes `found`, truncates to `min(nacc,max)`, uses snprintf not strcpy, and has its header guard).

PR #734, opened 2026-06-25
feat(kb): /v1/code/hybrid — RRF-fused hybrid code retrieval (P2 §5)

JBailes:
## §5 headline — a working hybrid retrieval endpoint

Wires the RRF fusion core (#733) into a real KB route. `GET /v1/code/hybrid?query&symbol&project` fuses independently-ranked retrieval signals into one ranking + a memory "why" context.

Two signals fuse in **file-path** space so cross-signal **consensus** is meaningful — a file that is *both* textually relevant to `query` AND structurally connected to `symbol` (calls it) rises to the top:
- **`code`** — lexical search over file contents (`canonical_index_code_search`)
- **`graph`** — callers of `symbol` from the structural call graph (`canonical_index_find_callers`), marked structural (RRF tie-break)

`kb_rrf_fuse()` blends them by `Σ_s w_s/(k+rank_s)`. Each result carries its contributing `signals`, source enrichment (code snippet / caller+line), `signal_hits`, and `structural_weight`.

Memory recall (`db2_memory_find_facts_like`) returns as a separate typed **`why`** array — the recorded reasoning behind the code, context rather than a fused file row (matching the §5 example *"+ the decision that explains X"*).

### Robustness
- 413 on oversized responses (never truncated/invalid JSON), 400 missing query, 503 uninitialized store, empty graph leg when no `symbol`.

### Test — `test_code_hybrid_*` in `unit-test-kb-http-routes` (shim-backed)
Both legs fuse + label + enrich; memory `why` surfaces; no-symbol path; missing-query 400. Adds a `db2_memory_find_facts_like` stub (full `memory_t` layout, `void*`-cast like the existing `canonical_index` stubs).

### Remaining for §5 (noted in the proposal)
Vector leg (`pgvec_code_search`, needs the query embedder — integration-tier) as a third fused signal; MCP tool + client bridge so the agent calls this before grepping; config-tunable per-signal weights.

> Stacks on #733 → #732 → #731. Merges cleanly once parents land.

### DIFF
--- a/docs/proposals/pending/code-graph-intelligence.md
+++ b/docs/proposals/pending/code-graph-intelligence.md
@@ -165,10 +165,21 @@ lists + per-signal weights + the RRF constant `k`, and returns one ranking by
 (structural-trust desc, then id) and a cross-signal-consensus boost. Fully unit
 tested (`unit-test-kb-rrf`): exact rank-blend math, weighting, consensus-beats-single,
 absent-signal robustness, determinism under input-order permutation, truncation.
-The remaining wiring — embedding the query, gathering the three signal lists
-(`db2_entity_edge_*` graph neighborhood, `pgvec_code_search` vector, memory recall),
-the `/v1/code/hybrid` route + MCP tool, and the config-tunable weights — is the next
-increment (needs the live embedder, so it is integration-tier rather than shim-tested).
+
+**Status — route shipped.** `GET /v1/code/hybrid?query&symbol&project` fuses two
+signals in **file-path** space through `kb_rrf_fuse` so consensus is meaningful — a
+file that is both textually relevant to `query` AND structurally connected to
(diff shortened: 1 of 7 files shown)


JBailes:
Roundtable review (4-panel) run on the handler. One real finding fixed in `f57eabe`: the two fused legs scoped inconsistently — Signal A passed `proj` (NULL→all projects) while Signal B passed `""` (not the all-projects sentinel); both now pass `proj`, matching /v1/code/callers. The rest were misreads from a brief the panel itself flagged as truncated 'at the for-loop': the result/why loops guard the array pointer, `file_path` is a fixed array (never NULL), the 413 guard prevents buffer overflow, cleanup frees all six buffers on every path, and `db2_memory_find_facts_like(query, limit=5, mems, max=25)` is correctly ordered (verified against the real prototype).

PR #747, opened 2026-06-26
docs(proposal): code-graph §5 vector leg shipped

JBailes:
Reconciles the code-graph-intelligence proposal: the §5 **vector leg** (embedding similarity over code_embeddings as a 3rd RRF signal in /v1/code/hybrid) is now shipped, so the route fuses **code+graph+vector**. Moves §5 vector leg from the deferred list to shipped. Doc-only.


JBailes:
Folded the proposal reconciliation into #749 (the §7 graph-informed-delegation PR) to keep the code-graph proposal updates in one place.

PR #753, opened 2026-06-26
feat(kb): read-only /v1/code/graph node-projection route (P8 §8)

JBailes:
## Summary

Adds the read-only `GET /v1/code/graph?project=&node=&max_results=` route — the backend for the webchat graph view (proposal §8). It returns a node's incident projection edges, each labeled with:

- `relation` (calls / contains / …)
- `direction` — `out` (node→neighbor) or `in` (neighbor→node, e.g. callers)
- `structural_weight`
- `provenance` — the §3 trust tag (`structural` for AST-derived projection edges)

Plus `neighbor_count` and a `truncated` flag when the edge scan hit the cap.

## Design
- Reuses `db2_code_projection_list_edges` (the published generation's edges, capped at `HUBS_MAX_EDGES`) and `kb_graph_edge_provenance` — no new DB query path.
- Read-only and off the agent hot path; sits beside the existing `/v1/code/graph/hubs` (§4) route in the KB sidecar dispatch.
- KB-internal route (like `/v1/code/graph/hubs`), so not part of the public OpenAPI surface.

## Tests
- `test_code_graph_node_ok` — `?project=proj-alpha&node=hub` over the projection stub (hub→a, hub→b, c→hub) asserts both out-neighbors (a, b), the in-neighbor (c), relation `calls`, `provenance:structural`, and `neighbor_count:3`.
- `test_code_graph_node_missing_params` — 400 on missing `node` / missing `project`.
- Full `unit-test-kb-http-routes` green; `aimee-kb` builds clean; clang-format-19 + docs-gen no-op.

### DIFF
--- a/docs/proposals/pending/code-graph-intelligence.md
+++ b/docs/proposals/pending/code-graph-intelligence.md
@@ -1,9 +1,10 @@
 # Proposal: Code-graph intelligence — a living, embedded, reasoning graph over code
 
-- **State:** IN PROGRESS — §0.5/§1/§3/§4/§5/§7 implemented (on the `testing` branch);
-  including the §5 code+graph+vector hybrid and §7 graph-informed delegation;
-  remainder (§2 tree-sitter, §4 surprising-links, §6 live fusion,
-  §8 webchat viz) still open, as build-/integration-/frontend-tier work. See the
+- **State:** IN PROGRESS — §0.5/§1/§3/§4/§5/§7 implemented (on the `testing` branch),
+  including the §5 code+graph+vector hybrid, §7 graph-informed delegation, the §4
+  surprising-links distance+selection core, and the §8 read-only `/v1/code/graph`
+  backend route; remainder (§2 tree-sitter, §4 surprising-links route, §6 live fusion,
+  §8 webchat frontend) still open, as build-/integration-/frontend-tier work. See the
   "Implementation status" section below.
 - **Thesis:** aimee should treat the codebase as a *living* graph that is (a) fully
   built without a manual step, (b) parsed broadly, (c) ranked by **graph structure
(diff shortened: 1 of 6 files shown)


JBailes:
## Review gate — roundtable (4/4, converged, not degraded) + local code-review

Both passes converged on two real issues; addressed in 6d68dc6:

1. **`truncated` conflated the DB scan window with the page cap** (flagged blocking by 3 panelists + code-review). The flag was `ne >= HUBS_MAX_EDGES` only, so a neighborhood cut by `max_results` reported `truncated:false`. Now the loop scans the full edge window to count every incident edge (`match_count`), emits up to `max_results`, and sets `truncated = (match_count > neighbor_count) || (ne >= HUBS_MAX_EDGES)`. New `match_count` field surfaces the pre-cap total.
2. **Self-loop semantics pinned.** A `source==target==node` edge (recursion) is now classified as direction `"self"` via a leading `is_source && is_target` branch rather than falling into `"out"`. (The "may emit twice" concern was already moot — it's a proper `else if` — but the guard makes it explicit + testable.)

Other roundtable "blocking" items were **refuted against the source** (the panel reviewed a prose brief, not the code):
- *Memory leaks on 400/503/413* — the 400 + db-not-init paths return **before** `calloc`; every post-alloc path frees `edges`; the serialize tail frees `edges`/`s`/`resp`. Clean on all paths.
- *Buffer overflow if max_results > HUBS_MAX_EDGES* — `max_r` is clamped ≤200, buffer is 10000, and `max_r` only bounds emission, never indexes the buffer.
- *No node filter → project-wide truncation* — real but matches the accepted sibling `/v1/code/graph/hubs`, and is exactly what `truncated` now signals.

Tests added: `test_code_graph_node_capped_truncates` (max_results=1 → truncated:true), `test_code_graph_node_self_loop` (rec→rec → single "self" entry). Full `unit-test-kb-http-routes` green, `aimee-kb` builds clean, clang-format-19 + docs-gen no-op.

Deferred to a follow-up (spans 3 handlers, out of scope here): extracting the duplicated serialize-and-cap tail block (hybrid/hubs/graph) into a `code_emit_capped` helper.

PR #757, opened 2026-06-26
feat(kb): /v1/code/graph/surprising — high-similarity, graph-distant link route (§4)

JBailes:
## Summary

Wires the merged §4 surprising-links **pure core** (`kb_graph_surprising` + `kb_graph_shortest_hops`, PR #752) into a read-only KB route. Surfaces code **file-node pairs that are semantically CLOSE** (high embedding cosine) **yet structurally FAR** (graph hop-distance ≥ `d_min`, or disconnected) — the "this module reinvented that one" signal.

`GET /v1/code/graph/surprising?project&max_results&k&d_min&percentile&min_cosine` →
`{ status, project, candidate_count, edge_count, d_min, percentile, links:[{a,b,cosine,hops|disconnected}], link_count, truncated }`.

## Design
- **New `pgvec_code_similar_pairs`** (`db2/pgvec_transport.c`): a project-scoped **lateral self-kNN** over `code_embeddings` — each anchor row rides the HNSW index for its top-`k` nearest others — cosine-floored, ordered closest-first. Pairs are deduped to canonical `a<b` **in C** (kNN is asymmetric, so an SQL `ak<bk` filter would silently drop one-directional pairs). A row's `node_key` **is byte-identical** to the projection file-node key (both from `db2_entity_node_key_file`), so candidate pairs join the projection graph with no remapping. The exact SQL was **validated against real pgvector 0.8.2 halfvec**.
- **`handle_get_code_graph_surprising`**: loads projection edges → `kb_graph_edge_t`, gathers candidates, runs the pure `kb_graph_surprising` (data-driven cosine percentile + hop distance), emits the links. Query-param tuned (`k`, `d_min`, `percentile`, `min_cosine`, `max_results`) — **no config churn**. A vector-store miss (no embeddings / embedder down) yields an empty well-formed result, not an error. `truncated` fires on output-cap OR either input scan window full.
- **Scope:** the §4 **LLM-judge / precision self-suppress** confirmation stage needs the LLM + a sampled corpus and remains a documented follow-up; this route returns the structural candidates.

## Tests
`test_code_graph_surprising_ok` — a disconnected high-cosine pair surfaces while a 1-hop pair is filtered by `d_min`; `kb_graph_surprising` runs **for real** (only the pgvec gather + projection edges are stubbed). `test_code_graph_surprising_missing_project` — 400. Full `unit-test-kb-http-routes` green; `aimee-kb` builds clean; clang-format-19 + docs-gen no-op.

### DIFF
--- a/docs/proposals/pending/code-graph-intelligence.md
+++ b/docs/proposals/pending/code-graph-intelligence.md
@@ -1,10 +1,11 @@
 # Proposal: Code-graph intelligence — a living, embedded, reasoning graph over code
 
-- **State:** IN PROGRESS — §0.5/§1/§3/§4/§5/§7 implemented (on the `testing` branch),
-  including the §5 code+graph+vector hybrid, §7 graph-informed delegation, the §4
-  surprising-links distance+selection core, and the §8 read-only `/v1/code/graph`
-  backend route; remainder (§2 tree-sitter, §4 surprising-links route, §6 live fusion,
-  §8 webchat frontend) still open, as build-/integration-/frontend-tier work. See the
+- **State:** IN PROGRESS — §0.5/§1/§3/§4/§5/§7/§8-backend implemented (on the
+  `testing` branch), including the §5 code+graph+vector hybrid, §7 graph-informed
+  delegation, the §4 surprising-links route (`/v1/code/graph/surprising`), and the §8
+  read-only `/v1/code/graph` backend route; remainder (§2 tree-sitter, §4
+  surprising-links LLM-judge confirmation, §6 live fusion, §8 webchat frontend) still
+  open, as build-/integration-/LLM-/frontend-tier work. See the
   "Implementation status" section below.
 - **Thesis:** aimee should treat the codebase as a *living* graph that is (a) fully
   built without a manual step, (b) parsed broadly, (c) ranked by **graph structure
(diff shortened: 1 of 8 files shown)


JBailes:
## Review gate — roundtable (4/4, converged) + local code-review

Both passes ran on the code. **Four** real issues found and fixed in `3bb515b`:

1. **Degenerate hop-distance (critical, code-review):** the projection has a `project --contains--> <every file>` edge, so over the full graph any two same-project files are exactly **2 hops** apart (file←project→file) — the default `d_min=4` filtered out *all* connected pairs, returning `[]` on every synced project. Fix: build the distance graph **excluding project-incident edges**, so hop distance reflects real code coupling (file→symbol→…→file). `edge_count` now reports the coupling-graph size. New test `test_code_graph_surprising_hub_excluded` (edge_count 0 + disconnected proves the hub was dropped).
2. **Unbounded anchor scan (code-review):** the lateral self-kNN ran one HNSW probe per file (the final `LIMIT` can't push past the global `ORDER BY cosine`). Added `anchor_cap` (5000) → O(anchors·k) regardless of project size.
3. **Asymmetric error contract (roundtable 4/4 consensus):** a real pgvec failure (`<0`) was coerced to an empty 200, indistinguishable from "no surprising links." Now **503 "vector store unavailable"**; `np==0` (not embedded yet / genuinely none) stays 200. New test `test_code_graph_surprising_vecstore_down`.
4. **`truncated` over-report + schema stability (roundtable):** ask `kb_graph_surprising` for `max_r+1` and flag truncated only when that extra slot fills (precise, mirrors the §8 fix). Every link now carries **both** `hops` (-1 = unreachable) and a `disconnected` bool — no per-row JSON key swap.

**Refuted against source** (the panel reviewed the brief, not the code): the "buffer overflow (1000 vs 4096)" — the C copy loop is guarded by `if (n >= max) break`, so SQL row count never drives the write; memory across all 7 allocations is freed on every path (code-review traced each); "O(N²) without an index" — `code_embeddings` has an HNSW cosine index and anchors are now capped. Auth/rate-limit/metrics findings are architectural constants shared with every sibling `/v1/code/*` route (auth at the mTLS layer), out of scope here.

SQL re-validated against real pgvector 0.8.2 halfvec; `aimee-kb` builds clean; full `unit-test-kb-http-routes` green.

PR #758, opened 2026-06-26
feat(mcp): expose /v1/code/graph/surprising as index({command:'surprising'})

JBailes:
## Summary

Makes the §4 **surprising-links** route (PR #757) **agent-callable**, completing its actuation surface. Mirrors the `index_graph_hubs` / `index_hybrid` exposure pattern (PR #737):

- **`kb_client_code_graph_surprising`** (`server/kb_client_code_graph.c`) — URL-escapes `project`, builds `/v1/code/graph/surprising?project=&max_results=`, forwards the route JSON **verbatim** via `kb_client_v1_get_json`; frees on every path.
- **`mcph_index_graph_surprising`** (`server_mcp_call_table.inc`) — requires `project`, bounds `max_results` (default 20, max 200), passes through `code_graph_passthrough`.
- **ext_tool schema + command-map** (`mcp_tools_extended.c`) — `{"surprising" → index_graph_surprising}` under the `index` family.

Agents can now ask *"where did we reinvent logic?"* — file pairs that are semantically close (high embedding similarity) yet structurally far apart or disconnected.

## Surface impact
The tool-surface **golden is unchanged** (top-level count stays **47**): the sub-command's props (`project`, `max_results`) are already in the `index` family's merged prop union from `index_graph_hubs`, and the sub-command is folded into the existing multiplexed `index` tool — no new top-level tool. Aligns with the MCP-presentation-program's minimal-surface goal (only the genuinely-new `surprising` capability is exposed; the §8 node-projection route stays frontend-only as it overlaps `find_callers`/`blast_radius`).

## Verification
`make all` builds clean, `line-check` clean, registry golden test passes, no docs drift, `check-kb-v1-coverage` ok. The forwarded route itself is already tested (`test_code_graph_surprising_*`, PR #757); the MCP layer is thin passthrough (same as `index_graph_hubs`, which ships without a dedicated handler test).

### DIFF
--- a/src/headers/kb_client.h
+++ b/src/headers/kb_client.h
@@ -1101,6 +1101,9 @@ char *kb_client_pdf_inspect_structure(const char *project, const char *document_
 char *kb_client_code_hybrid(const char *query, const char *symbol, const char *project,
                             int max_results, int *status_out);
 char *kb_client_code_graph_hubs(const char *project, int max_results, int *status_out);
+/* /v1/code/graph/surprising: file pairs that are semantically close yet
+ * structurally far (project required). */
+char *kb_client_code_graph_surprising(const char *project, int max_results, int *status_out);
 
 /* Counts of indexed files / definition-kind terms for one project.
  * Uses /v1 over remote HTTP or local UDS.
--- a/src/mcp_tools_extended.c
+++ b/src/mcp_tools_extended.c
@@ -119,6 +119,16 @@ void mcp_add_extended_tools(cJSON *tools)
    ext_prop(t, "max_results", "integer", "Max hubs to return (default 20, max 200).");
    ext_require(t, "project");
 
+   t = ext_tool(
+       tools, "index_graph_surprising",
+       "Find 'surprising links': file pairs that are semantically close (high embedding "
+       "similarity) yet structurally far apart in the call/dependency graph (or "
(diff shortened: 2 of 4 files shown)


JBailes:
## Review gate — code-review (clean) + roundtable (4/4, not degraded)

**code-review: []** — verified the diff is a byte-for-byte mirror of the merged `index_graph_hubs` path (leak/escape/bound parity exact; project URL-escaped; snprintf cap correct; all 3 registration names match).

**Roundtable** raised two "blocking" items, both refuted against the source:
- *URL injection via concatenation* — single `snprintf` with `%s` (escaped project) + `%d` (int); no other user params. Same pattern as the merged hubs forwarder.
- *max_results upper bound bypassed by direct callers* — the kb_client fn has the **identical** `max_results < 1 -> 1` clamp as `kb_client_code_graph_hubs` (neither upper-clamps); the upper 200 is enforced at the MCP layer **and** the route handler clamps `max_r > 200` as the backstop, so there is no unbounded path.

The suggestions (`code_graph_passthrough` shape assumptions, response-leak) are also refuted: the passthrough forwards JSON **verbatim** (no shape assumption), and `kb_client_v1_get_json` returns NULL on transport failure (nothing to free). One fair maintainability note — three near-identical kb_client forwarders — is left as-is; generalizing 3 tiny mirrors would be over-engineering.

Golden tool-surface unchanged (count stays 47); `make all` + line-check + coverage green.

PR #760, opened 2026-06-26
feat(kb): §4 surprising-links LLM judge — shared-symbol cross-check + batched confirmation

JBailes:
## Summary

Completes the **§4 relevance gate** — the last quality stage of surprising-links. `GET /v1/code/graph/surprising?judge=true` now confirms the top structural candidates instead of returning them raw.

For each candidate it: resolves both file-node keys → paths (`pgvec_code_node_path`), gathers each file's symbol outline (`canonical_index_structure`), computes a cheap **shared-symbol cross-check**, then runs **ONE batched Tier-B LLM-judge call** (`kb_curator_llm_run` — the curator's existing in-process provider path) asking which pairs are genuine parallel/duplicated logic vs coincidental similarity. Each judged link gains `shared_symbols` and, when the model returns a verdict, `confirmed` + `reason`; a top-level `judged` counts them.

## Design
- **`kb_surprising_judge.{c,h}`** — modeled 1:1 on `kb_curator_judge.c`: cJSON request from the pairs' paths+symbols, dispatch via `KB_CURATOR_STAGE_JUDGE` (**Tier-B**, so it reuses the configured `kb_curator_tier_b_*` provider — *no new config endpoint*), parse `{verdicts:[{i,surprising,reason}]}` by index with the `strchr('{')` prose-tolerance. Bounded: `SURPRISING_JUDGE_MAX=12` links × 40 symbols/file, **one** call per request.
- **Opt-in** via the `judge` query param (no config key → no docs churn). Degrades gracefully: no Tier-B LLM → `judged=0`, links returned as unconfirmed; unresolved node keys skipped (no LLM call); unparseable response → candidates still returned.
- **`pgvec_code_node_path`** — resolves a node_key back to its file_path (same `code_embeddings` row).

## Tests
- **`unit-test-kb-surprising-judge`** (new) — exercises the **real** request-build + verdict-parse + cross-check, DB accessors stubbed and the LLM faked via the curator **sidecar seam** (`cfg=NULL` + `printf` judge_cmd, the same seam `test_curator_judge.c` uses). Covers: per-pair verdict mapping, shared-symbol count, unparseable→-1, no-verdicts→0, unresolved-keys→skip.
- Route shim test asserts `judge=true` annotates `confirmed`/`reason`/`shared_symbols` + top-level `judged`.
- `aimee-kb` builds clean; line-check + docs-gen green.

## Validation note
The judge runs **inside aimee-kb** against its configured Tier-B LLM (on the split .254 deploy, that's `aimee-llm`). The `.254` LLM port isn't LAN-reachable from the build host, so end-to-end against the live model is a deploy-time check; here the parse/build logic is validated hermetically and the prompt follows the production `kb_curator_judge` pattern verbatim.


JBailes:
## Review gate — code-review (2 real → fixed) + roundtable (4/4, not degraded)

**code-review** ran on the source (2 agents): memory came back **clean** (all allocations balanced on every path; `strchr('{')` NULL-guarded; `out[idx]` bounds-checked; `char[40][128]` bounded by the `kept<40`+`snprintf(...,128)` caps). It found **two real logic issues, both fixed in `7f89f70`**:
1. **Fabricated verdict on a skipped pair** — the parse accepted any in-range `i`, so a hallucinated/duplicate echo could set `confirmed` on an unresolved (never-sent) pair. Now a per-pair `sent` flag is set when the pair is added to the request, and verdicts for un-sent (or already-judged) indices are rejected. New test `test_judge_rejects_unsent_verdict`.
2. **`shared_symbols:0` ambiguity** — conflated 'cross-check found none' with 'never computed'. The route now emits `shared_symbols` only for `sent` pairs.

**Roundtable** reviewed the brief only (the panel said so) and its 'blocking' items are all refuted against the code: the NULL-check and bounds-check exist; the stack arrays are capped; the request/response/verdicts/out frees are balanced (verdicts is route-owned and freed on every path); `err` *is* set on the unparseable path; `jn≤ns≤max_r`. Fair non-blocking notes (synchronous LLM latency, no caching, prompt-injection over the user's own symbols) are acknowledged: judging is **opt-in**, **bounded** (one batched call, ≤12 pairs), off the hot path; caching is a future optimization and the injection threat model is the user's own indexed code.

Both unit suites green (`unit-test-kb-surprising-judge` exercises the real parse/build via the sidecar seam); `aimee-kb` builds clean.

PR #763, opened 2026-06-26
feat(kb): §6 cross-session memory fusion — knowledge-graph leg in /v1/code/hybrid

JBailes:
## Summary

Realizes the **fusion half of §6**: folds the cross-session **knowledge graph** into `/v1/code/hybrid` as a 4th *ranked* signal (`memory`), not just the post-fusion `why` annotation. So a single hybrid query now ranks by **code + call-graph + vector + cross-session memory** — the proposal's thesis.

Symbol-anchored like the graph-callers leg: it seeds the symbol's entity node (`db2_entity_node_key_symbol`), walks its incident curator-built entity edges (`db2_entity_edge_explain_by_entity`), and resolves each neighbor entity to a `file_path` (`db2_entity_node_get`). Those files — which the recorded reasoning associates with the symbol but the code/graph/vector legs never see — fuse in via `kb_rrf_fuse` in file-path space. **This is the substrate a regenerated code-only snapshot can never hold.**

## Design
- 4th `kb_rrf_signal_t` leg (`fused` widened to `HYBRID_PER_SIGNAL*4`); `memory_items` deduped by file_path, ranked by edge order (weight-desc). Files it carries get `"memory"` in their `signals` array.
- Config-tunable **`code_hybrid_weight_memory`** (default 1.0; `<=0` disables — 6-touch: struct / default / parse / flat-field / docs-desc / configuration.md).
- **Gated** on a symbol + `w_memory>0`; **empty without an entity graph**, so zero behavior change on projects the curator hasn't synthesized yet. Memory cleanup added to all three free paths (alloc-oom / resp-null / final).

## Tests
`test_code_hybrid_memory_leg` — a symbol whose entity graph links to a memory-extracted file surfaces that file with `signals:["memory"]`. (Entity structs are mirrored in the test via void* cast, since `db2/entity_*.h` pull `memory.h`'s `edge_t`, which clashes with the route test's simplified `memory_t` — same pattern as the projection-edge stub.) `unit-test-config`/`unit-test-cmd-config` green; `aimee-kb` builds clean; line-check + docs-gen clean.

## Scope
Query-only (no-symbol) memory fusion and the §6 **live-reindex hook** (git post-merge/fetch) remain follow-ups; this lands the symbol-anchored fusion leg.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -22,7 +22,7 @@ aimee config set <key> <value>    # set one value
 
 Structured options (arrays, nested objects — e.g. `ensemble.reference_models`) are not CLI-settable; they are written into the config file under the sections listed at the end.
 
-## CLI-settable keys (122)
+## CLI-settable keys (123)
 
 | Key | Type | Description |
 |-----|------|-------------|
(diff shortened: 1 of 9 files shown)


JBailes:
## Review gate — code-review (clean) + roundtable (degraded 2/4)

**code-review** (reads the source) returned **[]** — all concerns verified safe: `eedges` freed exactly once on every path (calloc-fail/key-fail/success); `nmem` bounded by `< HYBRID_PER_SIGNAL`; the `file_path[512]→id[256]` copy is a **bounded snprintf**, and it's *less* truncation than the sibling code/graph legs already do (`MAX_PATH_LEN[4096]→id[256]`) — the existing RRF key-size behavior, not a new defect; `fused` correctly widened to `*4`; empty-entity-graph is a clean no-op.

**Roundtable** was degraded (2/4; glm/minimax dropped) and reviewed the brief — its blocking items are refuted by the code: 'unbounded copy' (it's `snprintf`), 'nmem overflow' (guarded), 'eedges leak' (freed outside the `if`), 'dedup uses ==' (it uses `strcmp`).

**Acted on one item** (`b231521`): the **self-edge** — a self-loop on the symbol entity made `neighbor==skey`, so the symbol's own file could appear as its own memory neighbor. Added a one-line skip; benign per code-review but semantically cleaner and flagged by 2 panelists.

`aimee-kb` builds clean, route test green.

PR #764, opened 2026-06-26
feat(kb): §6 live change-detection — skip code re-scan when default branch unchanged

JBailes:
## Summary

Lands the **detection half of §6 'live'**: notice that the canonical code moved, so a re-scan only happens when it actually did. (§1's idempotent drain already finishes the rebuild once `files` is refreshed; the gap was git-awareness.)

- **`git_resolve_default_sha`** (`code_collect.c`) — the default ref's **tree SHA** (resolves the §0.5 ref chain, then `git rev-parse <ref>^{tree}`). **`code_default_branch_changed`** — the pure gate (current-known-AND-differs, or never-indexed → reindex; unresolved → don't thrash).
- **`/v1/code/scan`** — on the local-scan path, a non-`force` scan now **no-ops the expensive git re-walk** when the default-branch SHA matches the last-indexed one (stored in `kb_runtime_state` under `code_scan_sha:<project>`) → `{"skipped":true}`. A successful scan persists the new SHA. `force` overrides. This is exactly the cheap gate a future **post-merge/fetch hook** reuses so it fires a rebuild only when the branch moved.

## Tests
- **`unit-test-code-collect`** — `git_resolve_default_sha` + the gate against **real git repos**: the tree SHA tracks commits, is stable when nothing moved, and is cleared on a non-git dir; the pure gate's four cases.
- Route tests — skip-when-unchanged and scan-and-persist-on-move (stubbed SHA + runtime_state).
- `aimee-kb` builds clean; line-check + docs-gen clean.

## Scope
The post-merge/fetch **hook install** (mirroring `verify_install_git_hook`) + the watch loop that *fire* this gate on a git event remain the integration follow-up (need a running KB + git events to validate). Also reconciles the §6 memory-fusion + live status in the proposal doc.

### DIFF
--- a/docs/proposals/pending/code-graph-intelligence.md
+++ b/docs/proposals/pending/code-graph-intelligence.md
@@ -3,8 +3,9 @@
 - **State:** IN PROGRESS — §0.5/§1/§3/§4/§5/§7/§8-backend implemented (on the
   `testing` branch), including the §5 code+graph+vector hybrid, §7 graph-informed
   delegation, the §4 surprising-links route + its LLM-judge relevance gate, and the §8
-  read-only `/v1/code/graph` backend route; remainder (§2 tree-sitter, §6 live fusion,
-  §8 webchat frontend, and the §4 precision self-suppress monitoring) still open, as
+  read-only `/v1/code/graph` backend route, and the §6 cross-session memory-fusion
+  leg in `/v1/code/hybrid`; remainder (§2 tree-sitter, §6 *live* reindex hook, §8
+  webchat frontend, and the §4 precision self-suppress monitoring) still open, as
   build-/integration-/frontend-tier work. See the
   "Implementation status" section below.
 - **Thesis:** aimee should treat the codebase as a *living* graph that is (a) fully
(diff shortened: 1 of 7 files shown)


JBailes:
## Review gate — code-review + roundtable (4/4) both caught a real bug

**Both** passes independently flagged a **shell-injection** (high severity): `git_capture_line` runs via `popen()`/`sh -c`, git ref names may contain a single quote, and the resolved default-branch ref is the live `origin/HEAD` target (attacker-controllable for an untrusted repo). My hand-wrapped `rev-parse '%s^{tree}'` let a branch named `main'$(...)'` break out → RCE during indexing. **Fixed** (`e2d4f6d`): build the revision then `shquote()` it (the proven pattern from `code_collect_from_git`). Regression test `test_default_branch_sha_quote_in_ref` clones a repo whose default-branch name contains a quote and asserts the SHA still resolves.

**Second real finding** (roundtable): the skip keys on the *default-branch* SHA, but under `AIMEE_CODE_INDEX_SOURCE=worktree` the index tracks the working tree — which the branch SHA can't see — so it would wrongly skip a needed WIP re-scan. **Fixed** (`96df528`): new `code_index_source_is_worktree()` predicate; the gate bypasses the SHA optimization in worktree mode. Tests added (route + env-based).

Also fixed a low-severity miss (code-review): a `force` scan now also persists the SHA (resolve runs whenever `!worktree`; the skip only when `!force`).

Refuted vs source: `sha_key` snprintf is bounded (no overflow), `stored[]` is pre-initialized, the skip response *includes* `files:0`/`inspected:0`, `kb_runtime_state` is a persistent DB2 table (not process-local). `aimee-kb` builds clean; `unit-test-code-collect` + route tests green.

PR #769, opened 2026-06-26
feat(kb): §6 live reindex hook — post-merge git hook fires the change-detection gate

JBailes:
## Summary

Completes **§6 live end-to-end**: **detect** (the #764 default-branch SHA gate) → **fire** (this hook) → **rebuild** (the §1 idempotent drain).

- **`code_index_install_branch_hook`** (`code_collect.c`): writes a **marker-guarded `post-merge` git hook** (mirroring `verify_install_git_hook` — won't clobber a foreign hook, returns -2) whose body backgrounds `aimee index scan <project> <root>` (shquoted). The #764 SHA gate makes that scan a **cheap no-op unless the default branch actually moved**, so a pull/merge keeps the code graph fresh without blocking the merge.
- **Opt-in** via `/v1/code/scan {install_hook:true}` (so `workspace add` can enable live reindex); **best-effort + fail-open** — a foreign hook or any error just sets `hook_installed:false` and never fails the scan.

## Tests
- `test_install_branch_hook` (real git repos): install / executable bit / "installed by aimee" marker / the backgrounded scan command with the project name / idempotent re-install / foreign hook left untouched (-2) / non-git → -1.
- `test_code_scan_installs_hook` (route): `install_hook:true` → installs + `hook_installed:true`; without the opt-in → installer never called + `hook_installed:false`.
- `aimee-kb` builds clean; build-integrity (tests ≤2000) + line-check + proposal gates green.

## Scope
An always-on **fanotify/inotify watch** (for non-git trees or freshness without a git event) remains an optional follow-up; the git-event path is the proposal's primary.

### DIFF
--- a/docs/proposals/pending/code-graph-intelligence.md
+++ b/docs/proposals/pending/code-graph-intelligence.md
@@ -4,9 +4,9 @@
   `testing` branch), including the §5 code+graph+vector hybrid, §7 graph-informed
   delegation, the §4 surprising-links route + its LLM-judge relevance gate, and the §8
   read-only `/v1/code/graph` backend route + the §8 webchat Graph view, and the §6
-  cross-session memory-fusion leg in `/v1/code/hybrid`; remainder (§2 tree-sitter, §6
-  *live* reindex hook, and the §4 precision self-suppress monitoring) still open, as
-  build-/integration-tier work. See the
+  live updates (memory-fusion leg + default-branch change-detection gate + post-merge
+  reindex hook); remainder (§2 tree-sitter, the §4 precision self-suppress monitor, and
+  an optional §6 fanotify watch) still open, as build-/telemetry-tier work. See the
   "Implementation status" section below.
 - **Thesis:** aimee should treat the codebase as a *living* graph that is (a) fully
   built without a manual step, (b) parsed broadly, (c) ranked by **graph structure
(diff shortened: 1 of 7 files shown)


JBailes:
## Review gate — code-review (clean) + roundtable (4/4, converged)

**code-review** (reads the source) returned **[]**, confirming the key worry — **shell injection** — is fully closed: `shquote` is correct POSIX single-quote escaping, applied to **both** `project_name` and `project_root`, at install-time (`git_capture_line`'s `-C` arg) **and** in the emitted hook body; paths use bounded `snprintf`; the install is fail-open and properly excluded on worktree/pushed-files/non-git paths.

**Roundtable** (4/4) flagged generic hook-installer concerns; two were worth acting on, fixed in `1858163`:
- **`core.hooksPath` ignored** (real *correctness* bug) → resolve the hooks dir via `git rev-parse --git-path hooks` (honors `core.hooksPath` + worktree/submodule), so the hook lands where git actually runs it.
- **Symlink write** → `open(O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW)` + `fdopen` so a planted symlink can't redirect the write outside the repo.

The rest were refuted against source: shquote IS POSIX escaping (mimo's caveat); paths are bounded `snprintf` (no overflow); the backgrounded scan is **intentional** (must not block the merge); TOCTOU is benign for a single-user local repo; and crucially **git never transfers hooks** (clone/fetch/pull don't copy `.git/hooks`), so there is no remote-attack vector — the symlink/TOCTOU concerns require local same-user FS access, outside the threat model.

`aimee-kb` builds clean; `test_install_branch_hook` + the route tests green.

PR #770, opened 2026-06-26
feat(kb): §4 precision self-suppress — disable noisy surprising candidates below a floor

JBailes:
## Summary

Completes **§4**. The key realization: the proposal samples precision "(LLM-judge **or** human spot-check)" — so the **judge itself is the precision sampler**. `confirmed/judged` is the *structural candidate generator's* sampled precision (the fraction of cosine+distance candidates the judge deems genuine). This adds the self-suppress half:

- **`kb_surprising_precision_suppress(judged, confirmed, min_samples, floor)`** — pure gate: suppress once enough samples accumulate **and** sampled precision < floor (floor ≤ 0 or too-few-samples → never). Unit-tested.
- **`/v1/code/graph/surprising`** keeps **rolling per-project `(judged, confirmed)`** counters in `kb_runtime_state` (halved past a cap so the metric tracks *recent* precision); a `judge=true` request accumulates its verdicts. When **not** judging and sampled precision is below the floor, the route returns no links + `suppressed:true` (the unjudged structural candidates are mostly false positives). The response also surfaces `sampled_precision` + `judged_samples` for observability.
- Config-tunable **`code_surprising_precision_floor`** (`kb.code_hybrid.surprising_precision_floor`; default **0 = off**, opt-in; 6-touch + docs).

## Tests
- `test_precision_suppress` (pure: floor-off / too-few-samples / below-floor → suppress / at-floor → not).
- `test_code_graph_surprising_self_suppress` (route: below-floor → `suppressed:true`/`link_count:0`/`sampled_precision`; at/above-floor → links shown). The runtime-state route stub is now key-aware for the surprising stats.
- `aimee-kb` builds clean; build-integrity (tests ≤2000) + line-check + `unit-test-config`/`cmd-config` + docs-gen + proposal gates green.

With this, **only §2 tree-sitter** (the one build-tier item) and an optional §6 fanotify watch remain.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -22,7 +22,7 @@ aimee config set <key> <value>    # set one value
 
 Structured options (arrays, nested objects — e.g. `ensemble.reference_models`) are not CLI-settable; they are written into the config file under the sections listed at the end.
 
-## CLI-settable keys (123)
+## CLI-settable keys (124)
 
 | Key | Type | Description |
 |-----|------|-------------|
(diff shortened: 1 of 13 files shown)


JBailes:
## Review gate — code-review (clean) + roundtable (4/4, not degraded)

**code-review** (reads the source) returned **[]** across 9 checks. The roundtable reviewed the brief and its 'blocking' items are refuted against the code:
- *sampled_precision div-by-zero* — the emit **is** guarded: `if (stat_judged > 0)`.
- *pure-gate div-by-zero (judged==0)* — `min_samples<=0` returns early **and** the call site passes 20, so reaching the division requires `judged>=20>0`.
- *integer-division precision=0* — the code uses an explicit `(double)confirmed/(double)judged` cast.
- *atoi absent/garbage* — `surprising_stats_get` inits to 0 and only overwrites on `get==0`; absent/garbage→0→no suppression (the safe default), no crash/UB.

Two items have substance but are **acceptable design properties, not bugs** (left as-is):
- **Decay flooring** biases precision *downward* by ≲0.2% per (rare) halving event — negligible, and *conservative* (errs toward more suppression).
- **Non-atomic counter**: a concurrent judge request could lose an increment, but `judged` and `confirmed` are written together so the **ratio** (the decision input) is preserved; for a best-effort self-monitoring heuristic (default-off, opt-in) this eventual-consistency is fine — transactions would be over-engineering. The decision never goes *wrong*, only its sample size is occasionally slightly smaller.

`aimee-kb` builds clean; build-integrity + line-check + config tests + the new `test_precision_suppress` / `test_code_graph_surprising_self_suppress` green.

PR #781, opened 2026-06-27
fix(workspace): base auto-detected worktrees on freshly-fetched default

JBailes:
## Problem

When aimee auto-detects the base for a new session worktree, `worktree_detect_base_branch()` resolved it to `origin/HEAD`'s remote-tracking ref or a local `main`/`master`/`trunk` **without fetching first**. A fresh worktree could therefore fork off a *stale* local copy of the default branch and collide at integration time — the direct cause of a recent spurious merge-conflict episode (local `stable` was behind `origin/stable`).

Basing new worktrees/branches on a stale default is a poor user experience: the work looks fine until it's merged.

## Fix

`worktree_detect_base_branch()` now:
1. Resolves the default branch **name** (`origin/HEAD` with the `origin/` prefix stripped, else local `main`/`master`/`trunk`).
2. Does a best-effort `git fetch origin <name>`.
3. Prefers the freshly-fetched `origin/<name>` as the base ref, falling back to the local branch (remote-less repo) then `HEAD`.

### Scope

Deliberately limited to the **auto-detect path** (SessionStart hooks, guardrails, lifecycle fallback). Explicit-`base_ref` callers are unchanged and already correct:
- `session_checkout_worktrees` already fetches `origin/<primary>` before basing.
- Delegate worktrees intentionally base on the parent anchor's **local HEAD** to inherit uncommitted WIP — fetching would defeat that.

The fetch mirrors `session_checkout_worktrees`' existing best-effort fetch; offline/remote-less repos fall back to local refs.

## Verification

- Builds clean under `-Werror -Wextra` (LTO).
- `unit-test-guardrails` passes, including the three `worktree_detect_base_branch` cases (faked-`origin/main` repo keeps `origin/main`; no-remote-default falls to local `main`; nonexistent repo → `HEAD`).
- Real-remote smoke test: a stale local `origin/main` is refreshed to the new upstream HEAD before being chosen as the base.

## Known consideration

This adds a best-effort network `git fetch` to the auto-detect path, which runs on SessionStart. It is consistent with the existing unbounded fetch in `session_checkout_worktrees`, and in the normal flow that path has already fetched moments earlier so this is usually a fast "already up to date" no-op. Bounding network git calls with a timeout is a cross-cutting concern (it would apply equally to the existing fetch sites) and is left as a possible follow-up rather than introducing a non-portable `timeout` wrapper here.

### DIFF
--- a/src/cmd_branch.c
+++ b/src/cmd_branch.c
@@ -59,7 +59,9 @@ static int branch_ref_exists(const char *branch)
 static int branch_fetch_origin_prune(void)
 {
    int ec = 0;
-   char *out = run_cmd("git fetch origin --prune 2>&1", &ec);
+   const char *fetch_argv[] = {"fetch", "origin", "--prune", NULL};
+   char *out = NULL;
+   ec = git_net_exec(NULL, fetch_argv, &out, 4096);
    if (ec != 0)
       printf("WARN: git fetch origin --prune failed: %s\n", out ? out : "unknown error");
    free(out);
--- a/src/cmd_session_lifecycle.c
+++ b/src/cmd_session_lifecycle.c
@@ -915,9 +915,9 @@ static void session_checkout_worktrees(const session_state_t *state, const char 
       char base_ref[320] = {0};
       if (project_primary_branch(gr, primary_branch, sizeof(primary_branch)) == 0)
       {
-         const char *fetch_argv[] = {"git", "-C", gr, "fetch", "origin", primary_branch, NULL};
+         const char *fetch_argv[] = {"fetch", "origin", primary_branch, NULL};
          char *fetch_out = NULL;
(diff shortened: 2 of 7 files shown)


JBailes:
Pushed a second commit that bounds the internal git fetches so they can't hang a session start (the review concern on commit 1's fetch). Both commits were roundtable-reviewed. The first round's blocking findings were false positives from a truncated brief; the second round surfaced two real points I acted on — forcing `GIT_SSH_COMMAND`/`GIT_TERMINAL_PROMPT` in the child env (rather than relying on `-c core.sshCommand`, which env precedence can override) — and two false ones I verified against the exec primitive (stderr **is** merged via `dup2`, so `2>&1` diagnostics are preserved; argv allocation is correctly sized).

PR #802, opened 2026-06-27
feat(harness-memory): P4 — session-start hydration of central memory

JBailes:
## Central agent-memory interception — P4: session-start hydration

Final slice. Completes the loop: P3 stores an agent's memory writes centrally; P4
**surfaces them back** to new sessions/agents by hydrating the local Claude memory dir
from the central store at session-start.

- `harness_memory_hydrate(cwd)`: resolve the project, fetch its live rows
  (`/v1/harness_memory/list`), and materialize each under
  `~/.claude/projects/<slug>/memory/<name>.md`. Best-effort — never blocks session-start.
- `hmem_slug_from_path` (pure, unit-tested) derives Claude's project-dir slug (non-alnum → `-`).
- Wired into `handle_session_start`; `harness_memory_common` added to the thin CLI link
  (project resolution is a client-side concern).

### v1 scope / follow-ups
Full reconcile (tombstones, spill consumption, divergence policy, flock), the config-driven
multi-client registry, and remote-server project-key resolution are documented follow-ups.
Core interception (P1–P3) is already merged; this makes the centralized memory flow back.

### Roundtable
Reviewed before PR; findings folded.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -761,6 +761,7 @@ set(CLI_SRCS
     ${AIMEE_SRC_DIR}/cli_main.c
     ${AIMEE_SRC_DIR}/cli_agent_setup.c
     ${AIMEE_SRC_DIR}/cli_session_start.c
+    ${AIMEE_SRC_DIR}/harness_memory_hydrate.c
     ${AIMEE_SRC_DIR}/cli_attention_guard.c
     ${AIMEE_SRC_DIR}/cli_code_audit.c
     ${AIMEE_SRC_DIR}/cli_css.c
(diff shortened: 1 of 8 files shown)


JBailes:
Roundtable review (degraded — .254 had an outage; 2/6 panelists, then a clean 2-round pass after recovery). 2 blocking items, both already handled: (1) NULL HOME/realpath/getcwd → guarded with early returns + slug NULL-guard (verified present); (2) slug-rule fidelity → verified against real `~/.claude/projects/<slug>/` dirs (non-alnum→'-', no separator collapsing) and locked with edge-case unit tests. CI green.

PR #804, opened 2026-06-27
feat(harness-memory): multi-client scope registry + atomic hydrate

JBailes:
Deferred items #1 + #2. Introduces `harness_memory_scope` — the single, extensible source of truth for a client's memory surface — consumed by both the interception detector and the session-start hydrator (removes the hardcoded Claude paths; adding an agent is one table row). Makes hydrate writes atomic (mkstemp+fsync+rename), matching P3. Unit-tested. Config-file override of the table is a documented follow-up. Roundtable-reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,6 +501,7 @@ set(CORE_SRCS
     ${AIMEE_SRC_DIR}/plugin_loader.c
     ${AIMEE_SRC_DIR}/memory_provider.c
     ${AIMEE_SRC_DIR}/harness_memory_common.c
+    ${AIMEE_SRC_DIR}/harness_memory_scope.c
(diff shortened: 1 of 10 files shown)


JBailes:
Roundtable (degraded, 3/6). 5 blocking → 2 real fixes applied: explicit-client requirement (no silent claude default → no cross-client contamination once the table grows) and memseg search anchored after the HOME-confirmed prefix. The other 3 were already handled: NULL/empty HOME is guarded (classify→ALLOW, hydrate→-1); every hydrate path snprintf is length-checked; the slug is '..'/'/'-free by construction (mangle maps all non-alnum→'-') and names are '..'-rejected. getpwuid/uid-ownership hardening is out of scope for a non-setuid CLI. CI green.

PR #806, opened 2026-06-27
feat(harness-memory): spill durability + audit log + session-start replay

JBailes:
Deferred items #4 (spill) + the durable part of #3 (reconcile/audit). A server-down fail-open now spills the intended write (atomic JSON envelope) and session-start replays + deletes it, so an outage never loses a memory. Adds the interception audit log (interception.jsonl). AIMEE_HOME via aimee_home(). Unit-tested. Disk-only import + tombstone removal (the external-edit reconcile) remain a documented follow-up — hydrate already does DB→disk divergence. Roundtable-reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -502,6 +502,8 @@ set(CORE_SRCS
     ${AIMEE_SRC_DIR}/memory_provider.c
     ${AIMEE_SRC_DIR}/harness_memory_common.c
     ${AIMEE_SRC_DIR}/harness_memory_scope.c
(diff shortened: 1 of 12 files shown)


JBailes:
Roundtable (clean, 6/6). 4 blocking, all applied: atomic 0600 audit-log creation (no fopen+chmod window); collision-safe spill commit name (mkstemp suffix, not pid/counter); fsync the spill dir after rename (survives power loss); audit 'spill-corrupt' on unparseable envelopes. consume_spills NULL/leak branches were already guarded. POSIX I/O guarded for the Windows build. CI green.

PR #808, opened 2026-06-27
feat(harness-memory): intercept Bash writes to memory files

JBailes:
Deferred item #5. A Bash command that writes a memory file is reject-denied (steer to the Write tool — the command's output can't be captured/stored). Reads are unaffected (write op must precede the memory path). Pure detector is unit-tested. v1 limit: process-subst/eval/var-indirection can evade (documented). Roundtable-reviewed.

### DIFF
--- a/src/memory_redirect.c
+++ b/src/memory_redirect.c
@@ -170,14 +170,29 @@ static int rematerialize(const char *path, const char *content, const char *home
 #endif
 }
 
-/* Any write operator in [a,b)? '>' covers >, >>, 2>, &>; plus tool keywords. */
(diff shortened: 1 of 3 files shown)


JBailes:
Roundtable (degraded, 5/6). Applied: quote/escape-aware '>' scan (no false-positive on quoted data like `grep 'a>b' mem.md`); added perl -i/-pi + patch keywords. Verified-and-tested as already-correct: `;`/`&&`/`|` are simple-command boundaries (a prior write op doesn't leak to a following read), and `>mem.md` (no space) is detected. The interpreter bypass (python -c/node -e/perl without -i) is the documented best-effort limit — the deferred inotify/fanotify backstop is the complete below-shell fix. CI green.

PR #841, opened 2026-06-28
docs(proposals): close out code-graph-intelligence (pending→done)

JBailes:
## What

Closes out the **code-graph-intelligence** proposal: every section §0.5–§8 has shipped to `testing`, so this reconciles the stale "Implementation status" summary and moves the proposal `pending/ → done/` (`State: done`).

### Reconcile
The §2 line in the Implementation-status summary still claimed only the **C grammar** and listed "§2 more grammars + call edges" as deferred — both contradicted the §2/§6 section bodies. Updated to the true shipped state: **all 16 hand-rolled-parity tree-sitter grammars** + nested-member descent + **call-edge extraction** + the opt-in `treesitter` CI lane.

### Close-out decision (roundtable-ratified)
A close-out roundtable (6 panelists, 0 failed, not degraded) converged on **close now**, carrying the two optional items as explicit future-work:
- **§2 grammars beyond the 16-language parity set** — no parity regression (those langs have no hand-rolled coverage); mechanical per-language add recipe recorded; the aspirational ≥30-language target is **not** claimed shipped.
- **§6 always-on fanotify/inotify watch** — recorded as a **known freshness limitation** (non-git / out-of-band trees are only as fresh as the last manual `aimee index scan`), deferred because this dev host SIGTERMs long-running watchers (an operational blocker, not a correctness one); revisit on a watch-capable deployment target.
- Also records a recommended **post-merge AppSec pass** and a **per-language grammar smoke-assertion** hardening note (both panel suggestions).

### Link hygiene
The two `code-graph-intelligence.md` references in `pending/cross-repo-dependency-graph.md` are repointed/delisted to plain text per the proposal-link convention (a done proposal keeps no live link), with the path moved out of its State bullet so the reconcile classifier stays clean.

## Verification
- `scripts/check-proposal-reconcile.py` → exit 0
- `scripts/check-proposal-links.py` → ok (8 links resolve across 15 files)

### DIFF
--- a/docs/proposals/done/code-graph-intelligence.md
+++ b/docs/proposals/done/code-graph-intelligence.md
@@ -1,15 +1,20 @@
 # Proposal: Code-graph intelligence — a living, embedded, reasoning graph over code
 
-- **State:** IN PROGRESS — **every section §0.5–§8 now has an implementation on
-  the `testing` branch**: the §5 code+graph+vector+memory hybrid, §7 graph-informed
-  delegation + blast-radius advisory, the §4 surprising-links route with its LLM-judge
-  relevance gate + precision self-suppress, the §8 `/v1/code/graph` backend route + the
-  webchat Graph view, the §6 live updates (memory-fusion leg + default-branch
-  change-detection gate + post-merge reindex hook), and the **§2 tree-sitter front-end**
-  with grammars for **all 16 supported languages** (opt-in `AIMEE_TREESITTER`
-  build, fall-through to the hand-rolled extractors). Only **optional follow-ups**
-  remain — more §2 grammars + nested-member/parse-tree call edges, and an always-on §6
-  fanotify watch. See the "Implementation status" section.
+- **State:** done
+- **Completed:** 2026-06-28
+- **Moved from:** `docs/proposals/pending/code-graph-intelligence.md`
+- **Summary:** **every section §0.5–§8 shipped to the `testing` branch** — the §5
+  code+graph+vector+memory hybrid, §7 graph-informed delegation + blast-radius advisory,
+  the §4 surprising-links route with its LLM-judge relevance gate + precision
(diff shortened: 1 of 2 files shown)


JBailes:
## Roundtable review (6 panelists, 0 failed, not degraded)

Ran a roundtable review on the diff before merge. Two **blocking** items, both real and now fixed in 619cbe5:

1. **Language-count honesty (real, latent in the original §2 body).** The text enumerated **17** language names (C and C++ separately) while claiming "all **16**". Verified against the code: the hand-rolled extractor has **16 language classes** (C/C++ share one `LANG_C`), and the tree-sitter front-end ships **17 grammars** (`tree_sitter_c` + `tree_sitter_cpp` split). Reworded every enumeration to "16 hand-rolled-supported language classes … as 17 tree-sitter grammars" so the count and the list agree.
2. **Watch wording could read as non-critical.** Promoted the always-on fanotify watch to a **documented freshness limitation** directly in the State summary (non-git / out-of-band trees stay only as fresh as the last manual `aimee index scan`), not just an "optional follow-up."

**Triaged, no change needed:**
- *"30+ target not disclaimed"* / *"State lacks branch linkage"* — already covered (the Deferred section explicitly says the ≥30 target is not claimed shipped; the State summary says "shipped to the `testing` branch").
- *"AppSec pass listed as recommended, not blocking 'done'"* — intentional: the close-out is a docs/lifecycle action; the AppSec pass is a separately-tracked follow-up, recorded as such.
- *Style nits (markdown link → plain text for the moved file)* — **forced by the proposal-link convention** (`check-proposal-links.py`: a done proposal keeps no live sibling link; the path moved out of the sibling's State bullet to keep `check-proposal-reconcile.py` clean). Both gates pass (exit 0).

Also dropped an ambiguous "+ 1" from the panel roster.

PR #850, opened 2026-06-28
feat(ingress): default-on envelope + compression + cache placement; close out proposal

JBailes:
## What

**Operator decision:** ship the ingress-compression levers **DEFAULT ON**, with documented reasons to turn them off. Closes out the **ingress-compression-and-cache-alignment** proposal (all machinery was already shipped behind flags + validated live on .254; this flips the defaults and files it to `done/`).

`config_set_defaults` now sets, default-on:
- `ingress_preinject_enabled` — the `<aimee-context>` memory/code preview on primary ingress turns
- `ingress_compress_enabled` — fold each code hit to a recoverable `file:line` ref (recover via `code_span_get`)
- `ingress_cache_placement_enabled` — place the envelope after the stable instructions prefix so provider prefix caches survive

Staying **opt-in**: `ingress_preinject_anthropic_enabled` (Claude-Code MCP-reachability gate) and failure-mining (`kb.mining.*`).

## Documented reasons to turn it off (proposal §8.0)
- The fold is a **pure win for non-agentic chat ingress** (no tool loop → no recovery): ~48% prompt reduction on code turns (envelope code block 416→93 tok, measured live on .254).
- For **agentic ingress** (Codex/Claude-Code, the dominant traffic) the agent re-opens the folded code, so `code_span_get` **recovery round-trips can erase or invert the saving** (the §6 net-token risk). Disable via per-request **`X-Aimee-Compress: 0`** or the config flags.

## Why ~50%, not 90%
Most compression systems fail to benchmark honestly — ~90% rates come only from dropping information the consumer needs or handpicking the dataset. ~50% is near-optimal for compression that **preserves the critical info** (folds stay recoverable via `code_span_get`); we use reliable datasets with numbers end-users can expect in production rather than a cherry-picked headline.

## Validation
- **Full `make unit-tests` passes.** The flip is test-safe: tests stub `config_load` / set the flags explicitly, so none depend on the off-default; Anthropic injection stays independently gated off.
- Regenerated `docs/gen/configuration.md` (3 flag descriptions). `check-proposal-reconcile`/`-links` exit 0.
- Machinery previously validated live on .254 (line-field on/off, `code_span_get` fail-closed, all flags toggle).

Roundtable review of this diff will be posted before merge.


JBailes:
## Roundtable review (6 panelists, 0 failed, not degraded)

Reviewed the diff before merge. Two items flagged "blocking" — both **refuted against the code**; the rest were good suggestions, the legit ones applied in dee817e.

**Refuted blockers:**
- *"compress↔preinject dependency not runtime-enforced → unrecoverable output"* — it **is** enforced by control flow: `ingress_preinject_build` early-returns `NULL` when neither `ingress_preinject_enabled` nor `typed_facts_enabled` is on, **before** the `compress` flag is read (`ingress_preinject.c:437` vs `:452`). So compress-without-envelope is a safe no-op — nothing is folded, so nothing is unrecoverable. Added a comment noting this; the default sets both on anyway.
- *"X-Aimee-Compress:0 uses unsafe thread-local state"* — already resolved (B1): the override is **request-scoped** via `request_context_t.compress_disabled` (parsed in `server_http_reqctx.c`, read via `request_context_get()`), not `__thread`. Verified in code.

**Applied (suggestions):**
- Trimmed the verbose `config_set_defaults` comment (metrics live in proposal §8.0) + added the control-flow-dependency note.
- Softened the non-agentic "pure win" → "when no recovery is needed."
- Noted the ~48%/~78% is a **single representative live measurement** (full per-task-class distribution = the §6 bench).
- Recorded the **per-ingress-class auto-detect** refinement (fold on for plain chat, off for detected agentic) as future-work.

**Acknowledged, not changed:** cache-placement's realized-cache_read benefit is unmeasured but the after-stable-prefix position is cache-friendly by design (low risk); the global default-on for agentic traffic is the operator's explicit decision with the per-request escape hatch.

Full `make unit-tests` passes; the flip is test-safe (tests stub `config_load`/set flags explicitly).

PR #854, opened 2026-06-28
feat(substrate): §1 structured verify verdict + ephemeral runner

JBailes:
Closes the §1 ("build-and-verify runner") residual of **autonomous-dev-execution-substrate** — the part the proposal explicitly defers to "an environment that can exercise it" (docker). Two pieces:

### §1 structured verify contract (`format=json`)
`handle_git_verify` gains `format=json` on `action=run`/`status`/`check`, returning a machine-stable verdict for the autonomous driver:
`{schema_version, verdict, reason, total, passed, failed, has_uncommitted_changes, steps:[{name, tier, status, exit, seconds[, log]}]}`.
- `verdict: "unavailable"` (no verify section / out-of-scope / cancelled) is kept **distinct from `"passed"`**, closing the §1 silent false-pass for the driver. The human `verify_gate_blocks` degrade-for-humans behaviour is unchanged.
- optional per-step `tier:` in project.yaml (default `mechanical`).
- text path byte-for-byte unchanged when `format!=json`; freshness cache written identically.
- extracts the verify-state persistence cluster into `git_verify_state.inc` to stay under the line-check ceiling (pure relocation).
- new `unit-test-git-verify-contract` exercises the real `verify_build_verdict`.

### §1 ephemeral sandboxed runner
- `docker/dev-runner/Dockerfile`: ubuntu:24.04 + the exact ci.yml toolchain + the deps CI gets implicitly from ubuntu-latest. **Validated on a live docker host: `make -j all server` builds clean inside; `make lint`/`line-check` green.**
- `scripts/dev-runner-pin.json` + `scripts/dev-verify-runner.sh`: ephemeral, hardened (`--rm --network=none --cap-drop=ALL --no-new-privileges`, non-root, mem/PID limits, timeout), tree shipped via `git archive` (no .git/tokens), toolchain pin-check, exit-code→verdict mapping. **Validated end-to-end: passing step → `passed`, failing step → `failed`.**

Roundtable code-reviewed (the §1 verdict change); the panel's grep-only memory-safety flags were verified false, and the three valid sub-points (serialize-OOM verdict, redundant config load, bounded CLI exit-code heuristic) are folded in (`fix(...) (review)`).

### DIFF
--- a/docker/dev-runner/Dockerfile
+++ b/docker/dev-runner/Dockerfile
@@ -1,0 +1,38 @@
+# Ephemeral verify runner — autonomous-dev-execution-substrate §1.
+#
+# Carries the CI toolchain so a local mechanical/integration-tier verify run
+# reproduces .github/workflows/ci.yml exactly (gcc/make + the same -dev libs +
+# clang-format-19 + postgresql-client). The autonomous driver runs a work-item
+# worktree's build/unit/lint steps inside a throwaway container started from this
+# image (see scripts/dev-verify-runner.sh) — never on the host, never with the
+# host's secrets or network.
+#
+# Pinned: the resolved toolchain versions + this image's id are recorded in
+# scripts/dev-runner-pin.json; dev-verify-runner.sh refuses to run if the running
+# image drifts from that manifest (so a local pass that CI would fail cannot ship).
+# The base is pinned by tag (the homelab base registry exposes no content digest);
+# the pin-check binds the toolchain versions, which is what actually gates parity.
+FROM ubuntu:24.04
+
+# Mirror ci.yml's apt sets: the lint job installs clang-format-19; the build /
+# build-integrity jobs install the gcc/make + -dev libs (+ postgresql-client).
+RUN apt-get update -q && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y -q --no-install-recommends \
+        gcc make pkg-config clang-format-19 \
+        libsqlite3-dev libcurl4-openssl-dev libpam0g-dev libssl-dev libpq-dev \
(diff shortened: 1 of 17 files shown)


JBailes:

**§1 verdict change (git_verify):** the panel raised 6 "blocking" memory-safety items; triaged against the actual code, **all 6 were false positives** from a grep-only view — the `cJSON_Delete(v)` after each `cJSON_PrintUnformatted`, the `job->lock` held across the async-status read (and `verify_job_reset_locked` frees under the same lock), and the assign-then-return of the json `mcp_text` result are all present. The 3 genuinely valid sub-points were folded in (`fix(...) (review)`): serialize-OOM keeps an explicit verdict instead of downgrading to NULL; the check-action probes config once (skips verify_check's freshness I/O when there's no section); the CLI exit-code heuristic scans only the JSON's leading region so a `"verdict":"failed"` substring in a step log can't spuriously flip the exit code.

**§1 runner wrapper (dev-verify-runner.sh):** the panel found **4 real defects**, all fixed — pin-check now fails closed on a missing/malformed manifest (was: ran unpinned); IMAGE/MEM/PIDS/TIMEOUT/SSH validated against an allowlist (was: spliced unquoted into the root remote shell); portable base64 + empty-encoding guard (was: GNU-only `-w0` → empty step → false pass on non-GNU clients); `git stash create` snapshots uncommitted tracked changes (was: `git archive HEAD` dropped the in-progress work item).

Re-validated end-to-end on a live runner: passing step → `passed`, failing → `failed`, injection/bad-arg → rejected, unreachable/pin-broken → `unavailable` (never a false pass).

PR #856, opened 2026-06-28
feat(fullauto): WP-5 safety rails — merge-target guard + per-run caps

JBailes:
First safety-floor packet for **full-autonomous-development** (closeout). Lands the rails BEFORE any live forge, so an autonomous run can never run away or merge to a protected branch.

- **Merge-target rail** (`wfe_iface`): one source of truth `wfe_autonomous_base()` (default `testing`); `wfe_autonomous_target_ok()` checked at the top of `exec_pr_open`/`exec_merge` — a misconfig to main/master/release* fails closed. Promotion to main stays human. F4's live forge resolves its base/target through this.
- **Per-run caps** (`wfe_autonomy_run`): cumulative turn cap (from the persisted audit log → bounds the whole run across resumes) + per-resume wall-clock cap → park `budget_exceeded` on breach. Env-overridable; malformed override falls back to default (rail can't be silently disabled).

Tests: `test_wfe_autonomy` A6 (guard) + A7 (turn cap); existing wfe suite regresses clean.

Per the closeout design roundtable (Q2 resolved **default-OFF** for the live forge): the broader floor (worktree, verify-wiring, single-flight, live forge default-OFF, intake auth) follows; this self-contained rail gates them.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -213,7 +213,7 @@ Scalar keys read directly from the config root (not via the CLI allowlist above)
 
 ## Environment variables
 
-The binaries read 125 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
+The binaries read 126 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
 
 ### Paths & assets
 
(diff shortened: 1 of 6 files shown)


JBailes:

The panel found real soundness defects in the rails (it reviewed the actual diff). Fixed:
- **merge-target guard**: now case-insensitive (`Main`/`MASTER` no longer slip past) and the release-train rule is anchored (`release/<v>` / `release-<digit>`) so it protects the train without snagging a `release-notes-*` branch.
- **caps fail closed**: `wfe_env_long` rejects `strtol` overflow (ERANGE); a negative event-list (DB error) now **parks** rather than running unbounded; a budget breach that can't read the row to park surfaces a hard error (−1) instead of claiming success; an already-parked item still gets the breach audited.
- **wall-clock** uses `CLOCK_MONOTONIC` (immune to clock jumps), framed as a per-resume guard (total run bounded transitively by the cumulative turn cap).

The panel's "missing header declarations" item was a false flag — the decls are in `wfe_iface.h`; the diff sent to the panel just didn't include the header. The turn-cap-precision/return-sentinel points are addressed by documenting the event-count as a deliberately conservative cross-resume upper bound + the established `pause_reason=budget_exceeded` signal callers already inspect.

Tests extended (case-insensitive + anchored-release); wfe suite regresses clean.

PR #860, opened 2026-06-29
feat(fullauto): F3a — mechanical verify gate in the implement loop (WP-1b)

JBailes:
Wires the §1 `git_verify format=json` verdict (from autonomous-dev-execution-substrate) into the manager loop: **implement only advances a unit that PASSES verification**.

- new `wfe_verify_provider_t` seam (mirrors the delegate/forge seams; wfe lib never calls server git_verify directly — module-boundary). Default NULL → implement skips verification (drivable without a provider).
- `exec_implement`: after dispatch + freeze, run the gate. `verdict:passed` → advance; anything else (failed / unavailable / gate-unrunnable) → loop (**fail closed**). The engine bounds the loop via `stage_attempt` → parks `max_attempts`; a re-dispatched fresh engineer delegate has the verify tool to fix findings. The verdict token is matched only in the leading region (a `"verdict":"passed"` substring in a step log can't spoof it).
- server registers a live verify provider (`handle_git_verify` sync, format=json).

Test: `test_wfe_delegate_seam` case D (skip / passed / failed / unavailable / gate-unrunnable / leading-region spoof). wfe suite regresses clean.

Per the roundtable, the manager loop is essential to close the proposal; the N-skeptic adversarial fan-out + patch-coordinator remain Phase-C depth.

(Branch name says f2-worktree; this PR is F3a — F2 worktree follows.)

### DIFF
--- a/src/server/wfe_live_delegate.c
+++ b/src/server/wfe_live_delegate.c
@@ -23,6 +23,8 @@
 #include "agent_config.h"
 #include "agent_exec.h"
 #include "agent_types.h"
+#include "cJSON.h"
+#include "headers/git_verify.h"
 #include "log.h"
 #include "util.h"
 #include "wfe_approval.h"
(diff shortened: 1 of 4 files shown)


JBailes:
All three blocking items were real and are fixed: the gate now **parses the verdict with cJSON and reads the top-level `verdict` field** (a nested/echoed `verdict:passed` can't spoof it; no truncation slicing the real field out of a large document); a **missing provider fails closed** (a safety gate must never advance unverified work because registration was skipped — the live provider is installed by `wfe_autonomy_register`, default-on); an unparseable verdict fails closed. Test D updated accordingly; wfe suite green.

PR #862, opened 2026-06-29
feat(fullauto): F1b — make the per-run USD budget cap functional (WP-5)

JBailes:
The engine has a per-work-item USD cap but delegate steps reported `cost 0.0`, so it never tripped. This threads a real server-side cost so the cap bites.

- new `wfe_autonomy_cost_estimate(secs)` — authoritative server-side USD = delegate wall-clock × `AIMEE_AUTONOMY_USD_PER_SEC` (default 0.0005). Provider-agnostic (never trusts a provider figure → can't be understated); negative clamps to 0; malformed rate → default.
- `wfe_delegate_dispatch` times `CLOCK_MONOTONIC` around the run; author/implement/document report the estimate (was 0.0) → the engine's existing cost check parks `budget_exceeded`.
- intake sets a default per-run cap ($5.00, `AIMEE_AUTONOMY_MAX_USD`-overridable; 0 disables).

Test: `test_wfe_autonomy` A8 (estimate math, clamp, env override + malformed fallback). wfe suite green. Completes the WP-5 budget story (the turn + wall-clock caps shipped in #856; this makes the USD cap live too).

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -213,7 +213,7 @@ Scalar keys read directly from the config root (not via the CLI allowlist above)
 
 ## Environment variables
 
-The binaries read 126 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
+The binaries read 128 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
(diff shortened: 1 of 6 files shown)


JBailes:
All five blocking items were real and are fixed (all on 'the cap must bite'): **reject inf/NaN** for both the rate and the cap envs (strtod accepted `inf`; `d>=0` is true for +inf → silently voided the cap); rate must be strictly >0. **Charge cost on every return path** (`with_cost` helper) — a dispatch failure or downstream gate failure previously discarded the measured cost, so a retry-loop runaway paid nothing; the engine adds `r.cost_usd` before the status branch so looped/failed now accrue. **clock_gettime checked + zero-init** — no indeterminate subtraction on the rare failure path. wfe suite green.

PR #865, opened 2026-06-29
feat(fullauto): F2 — per-work-item git worktree isolation

JBailes:
Each autonomous run's producing blocks act in a per-work-item locked worktree (`aimee/wi/<id>`) instead of one shared checkout — a prerequisite for F4 + multi-run scale.

- schema: `worktree` column on lifecycle_work_item (canonical + legacy migration) + setter + `wfe_ctx_worktree()`.
- `wfe_worktree_ensure` (lazy, `git worktree add --lock` per Q5, persisted, idempotent) + `wfe_worktree_cleanup`.
- author/implement/document/freeze resolve their workdir to the worktree, **falling back to the shared repo dir on any failure** (degrade, don't crash). Dispatch takes the workdir.

Note: a worktree is a path convenience, not a sandbox (the real seccomp/namespace sandbox stays a GA gate); terminal cleanup-hook + orphan-sweep are a follow-up (helper is wired-ready).

Test: `test_wfe_delegate_seam` E over a real git fixture. module-boundary + schema-sync green.


JBailes:
All six blocking items (worktree lifecycle robustness) fixed: **DB persist rc checked** (scrub the worktree+branch if recording fails, so no unrecorded collision); **reuse only if the path still exists on disk** (is_dir; stale → clear+recreate, never a vanished CWD); **per-item flock serialization** + re-check under lock (no concurrent `worktree add` of the same branch); **scrub partial state on git-add failure**; **truncation-checked snprintf** (corrupt path never reaches git → fallback); **cleanup wired** into the autonomy terminal/failed path (orphan-sweep for non-autonomous paths is GA). wfe suite green.

PR #868, opened 2026-06-29
feat(fullauto): F4 — live forge provider (default-OFF)

JBailes:
The final floor packet — a real git push + `gh` PR/CI/merge provider for the wfe forge seam, closing the criterion-5 gap the substrate proposal deferred here.

**Security — registered only on operator opt-in:**
- new `wfe_live_forge_enabled` config flag, **default OFF** (test_config asserts default 0). Turning it on is a deliberate operator deployment action (branch protection + scoped creds + break-glass).
- `wfe_live_forge_register()` installs the provider only when the flag is on; else the engine keeps its fail-closed stub (pr.open re-loops, merge parks).
- **every op re-checks the flag AND the F1a merge-target rail** and fails closed if either is off — a mid-run flip or protected-base misconfig can never open/merge a real PR.

Reuses the vaulted handlers (`mcp_git_run`): open = push + `gh pr create --base <autonomous_base>`; ci/mergeable/is_merged/merge = `gh pr checks|view|merge` mapped to the enums (unknown → fail closed/park). Shell-arg quoted. `exec_pr_open` now derives `aimee/wi/<id>` (F2 branch) when the env override is unset.

Live PR roundtrip stays a deployment-tier validation-pending check (never a closeout test); production enable is an operator gate.


JBailes:
### Roundtable security-review summary
All six blocking items fixed: **TOCTOU re-checks** of `forge_allowed()` immediately before each mutating call (push, pr create, merge); **read ops** (ci_status/mergeable/is_merged) now use the full `forge_allowed()` (flag AND rail), uniform; **shq() returns −1 on truncation** and `live_open` aborts (never a half-built command); the **autonomous base is validated** (NULL/empty → fail closed) before reaching shq/gh/log; **parse_pr_number** anchors on the canonical `/pull/<N>` URL so a later digit can't substitute the wrong PR. config_load failure already reads as disabled. Build + config + wfe suites green.

PR #870, opened 2026-06-29
fix(fullauto): F2 polish — custom blocks in the worktree + orphan-sweep

JBailes:
Two isolation/cleanup gaps F2 left:
- **exec_custom** (custom command blocks + their freeze) ran in the shared `repo_dir()`, not the per-work-item worktree — so a custom block couldn't see the run's own changes. Now uses `resolve_workdir()` like the other producing executors (same fallback-on-failure).
- a worktree could **leak** when an item went terminal via a non-autonomous path (API gate reject, override-cap reject) — the autonomy terminal cleanup never ran. The scheduler sweep now tears down + clears the worktree of any terminal item still holding one, covering all paths.

Build + wfe suite green.

### DIFF
--- a/docs/proposals/pending/full-autonomous-development.md
+++ b/docs/proposals/pending/full-autonomous-development.md
@@ -43,28 +43,35 @@ per-item USD cap).
   worktree is a path convenience, **not** a sandbox — the real seccomp/namespace
   sandbox stays a GA gate.
 
-**Resolved by design / remaining:**
-- **F5a — single-flight is SATISFIED BY THE SEQUENTIAL SCHEDULER.** `wfe_autonomy_run`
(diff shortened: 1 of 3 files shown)


JBailes:
### Roundtable review summary
- **terminal whitelist**: the sweep now matches `accepted/rejected/abandoned` explicitly (not `!= active`), so a parked item (state stays active) or a future non-terminal state is never reaped; and since those terminal states are immutable, acting on the snapshot can't race an item back to life (closes the TOCTOU concern).
- **workdir guarantee**: `resolve_workdir` now guarantees a non-empty path for every caller (on top of the repo_dir() fallback) — exec_custom can't hand an empty CWD to exec/freeze. (The original concern was a false flag — resolve_workdir already fell back to repo_dir(); this makes it bulletproof + uniform across all 4 executors.)
wfe suite green.

PR #877, opened 2026-06-29
fix(webchat-git): cross-origin /vscode WS gate + revoke recycles editor/ssh-agent

JBailes:
Closes two confirmed security gaps in the already-shipped webchat git-projects + in-browser-VSCode feature (proposal `docs/proposals/pending/webchat-git-projects-and-vscode.md`), surfaced by a gap audit + roundtable review of the shipped code.

## 1. Cross-origin WebSocket gate on `/vscode` (CSRF)
The `/vscode` reverse proxy gated on the session cookie (`requireAuth`) and stripped upstream auth, but did **not** validate `Origin`. A cross-origin page in the victim's browser could open a WebSocket to `/vscode/*` carrying the victim's cookie and drive code-server's integrated terminal — cookie `SameSite=Strict` was the only mitigation, not an explicit gate.

`sameOriginVSCode` now rejects cross-origin requests. A WebSocket upgrade with a **missing** `Origin` is refused (browsers always send one on an upgrade); a missing `Origin` on a non-upgrade request is allowed only for safe methods or an explicit `Sec-Fetch-Site: same-origin`. Scheme/host are compared **normalized** (default port stripped, `X-Forwarded-Proto` honored) so same-origin `:443`/`:80` and TLS-terminating-proxy deployments are not wrongly `403`'d. 18-case unit test.

## 2. Revocation leaves no live credential handle (G5)
`DELETE /v1/git/credentials` removed the vault entry but left the user's running code-server (token baked into its spawn env) and ssh-agent holding live handles, so a revoked credential survived in a running editor. On successful delete, aimee-server now stops the editor (respawns lazily with a credential-free env) and the ssh-agent — **deleting the vault entry first** so a racing respawn re-reads an already-empty vault. webchat evicts its cached editor port on revoke so the next `/vscode` request re-ensures cleanly.

## Review
Roundtable code-review run on the diff (6/6 panelists). Both blocking items it raised — the TLS-terminating-proxy scheme mismatch and default-port normalization (each would have `403`'d legitimate same-origin traffic) — are fixed and covered by added test cases. Suggestions addressed: empty-Origin non-safe-method tightening, `editorPorts` eviction on revoke, delete-first ordering rationale documented. The "capture stop() return values" suggestion does not apply (both stop functions are `void` + idempotent by design; documented in-code).

Tests: `go test ./webchat/...` green; `aimee-server` links clean; gofmt + clang-format-19 clean.

### DIFF
--- a/src/server/server_http_routes.inc
+++ b/src/server/server_http_routes.inc
@@ -20,6 +20,7 @@
 #include "git_host_cred.h"    /* per-host git credential store for /v1/git/credentials */
 #include "git_oauth_github.h" /* GitHub device-flow sign-in (/v1/git/oauth/github) */
 #include "git_ops.h"          /* git_ops_run for /v1/workspace/git (WP-E) */
+#include "git_ssh_agent.h"    /* git_ssh_agent_stop — drop live key handles on revoke */
 #include "git_project.h"      /* git_project_clone for /v1/workspace/clone (WP-D) */
 #include "webuser_editor.h"   /* webuser_editor_ensure for /v1/workspace/editor (WP-I) */
 #include "workspace_scope.h"  /* ws_scope_user_root — project workspace root */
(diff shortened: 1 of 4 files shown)


JBailes:

**Blocking — fixed before this push:**
- *TLS-terminating-proxy scheme mismatch*: `forwardedProto` only consulted `r.TLS`, so behind an external TLS terminator it computed `http://` while the browser sent `https://`, 403'ing every same-origin request. Now honors inbound `X-Forwarded-Proto`. Test: `behind TLS proxy, XHR/WS https origin`.
- *Default-port normalization*: raw `scheme+"://"+r.Host` compare failed when `r.Host` carried `:443`/`:80` but Origin omitted the default port (the common deployment). Now URL-parsed + port-normalized. Tests: `... default 443/80 stripped`.

**Suggestions — addressed:**
- Empty-Origin fast path tightened to safe methods + `Sec-Fetch-Site: same-origin` (no non-safe cross-origin slip-through without an Origin).
- webchat now evicts its `editorPorts` cache on revoke so the next `/vscode` re-ensures cleanly instead of self-healing after a failed dial.
- Revoke ordering: vault-delete-first → stop, documented (a racing respawn re-reads an empty vault).

**Verified-not-applicable:**
- "Capture `webuser_editor_stop`/`git_ssh_agent_stop` return values": both are `void` + idempotent by design (signal/reap, no recoverable failure mode); documented in-code. `git_ssh_agent_stop` is non-blocking; `webuser_editor_stop` reaps synchronously with a ~500ms bound — acceptable for a rare revoke.
- "Forgeable same-origin Origin via the browser": panel found none in the reviewed code.

PR #878, opened 2026-06-29
feat(webchat-git): wire idle-reaping of per-webuser code-server editors

JBailes:
Part of finishing off `docs/proposals/pending/webchat-git-projects-and-vscode.md`. The gap audit found that `webuser_editor_reap_idle()` exists but is **never scheduled** — so every per-webuser code-server lives until server shutdown. The proposal's resource-cost mitigation (V2 idle teardown) was effectively unimplemented; N concurrent users meant N editors forever.

## Changes
- **Schedule the reaper.** The server park loop's existing 1s tick now calls `webuser_editor_reap_idle()` every ~30s, bounded by a new `AIMEE_WEBCHAT_EDITOR_IDLE_SECS` (default 1800 = 30 min; `0` disables; malformed/negative → default).
- **Keep active editors alive.** code-server holds a single long-lived WebSocket and makes no further `/vscode` HTTP requests during a session, so the server-side idle timer (refreshed only by `ensureEditorPort`) would otherwise reap an actively-open editor mid-use. The proxy now runs a keepalive ticker for the WebSocket's lifetime that re-ensures the editor (refreshing `last_active`); the ticker stops when the connection ends, so a closed tab's editor reaps normally.

## Tests
`unit-test-webuser-editor` extended to cover `webuser_editor_idle_secs` parsing (default/explicit/zero/negative/malformed/trailing-junk). `docs/gen/configuration.md` regenerated for the new env var. Go + C build clean; gofmt + clang-format-19 clean.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -216,7 +216,7 @@ Scalar keys read directly from the config root (not via the CLI allowlist above)
 
 ## Environment variables
 
-The binaries read 131 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
+The binaries read 132 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
 
 ### Paths & assets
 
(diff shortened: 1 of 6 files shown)


JBailes:
**Roundtable code-review summary** (5/6 panelists)

**Blocking — fixed before this push:**
- *Lock held across reap.* `webuser_editor_reap_idle` held `g_lock` across `reap_locked`'s up-to-500ms kill/wait, so a bulk-idle tick would freeze every concurrent `webuser_editor_ensure` (and stall the park loop) for N×500ms. Refactored to two-phase: under the lock it collects stale pids and clears their slots; the lock is dropped before the kill/wait (new `reap_pid` helper). A racing `ensure` for a just-cleared principal simply spawns a fresh editor with a different pid — no double-reap.
- *Keepalive vs idle-threshold invariant.* The keepalive cadence (`editorPortTTL`, ~30s) and the operator-set idle threshold were an unenforced invariant; a sub-30s `AIMEE_WEBCHAT_EDITOR_IDLE_SECS` would reap live WS sessions. `webuser_editor_idle_secs()` now clamps any positive value to a **60s floor** (2× the keepalive cadence).

**Suggestion — fixed:** `strtol` overflow (`ERANGE`) now falls back to default instead of silently returning `LONG_MAX`; added a 7-day ceiling so a fat-fingered value can't silently disable reaping. Tests cover floor/ceiling/overflow.

**Acknowledged, not changed:**
- Idle value is sampled once (cached in `reap_tick`) — env is startup config, documented; runtime changes need a restart.
- One-tick keepalive skew after WS close (an in-flight `ensure` may complete post-close) — cosmetic, harmless.

**Confirmed non-issues (panel):** `defer close(stop)` does fire on the hijacked-WS path (no goroutine leak); `last_active` is fully `g_lock`-protected on both sides; `reap_tick`'s counter has no race (single-threaded park loop).

PR #879, opened 2026-06-29
docs(proposals): close out webchat git-projects + in-browser-VSCode → done/

JBailes:
Files the **webchat git-projects + in-browser-VSCode** proposal (and its implementation plan) to `done/` now that both phases have shipped.

## Context
The feature was implemented across many merged PRs (WP-A #451 per-user workspaces, WP-F #461 git backend, WP-J #467 /vscode proxy, credential panels #486–#489, nav #496, session-worktree #611). A gap audit + roundtable review of the *shipped code* drove two final hardening PRs before close-out:
- **#877** — cross-origin `/vscode` WebSocket gate (CSRF) + revoke recycles editor/ssh-agent.
- **#878** — wire idle-reaping of per-webuser code-server editors (they previously lived until shutdown).

## This PR
- Fixes the residual rev.1 design-text contradiction (the SSH-key `0600`-temp clause vs the vault-only/no-disk invariant) and folds the same-origin WS-upgrade requirement into V3.
- Adds a **Status at close-out** section: shipped surface, acceptance-criteria audit (incl. the HTTPS-token-in-`/proc/environ` PARTIAL and the deliberate **default-ON** deviation), and six deferred follow-ups (SSH web intake, open-PR op, socket-fed askpass, extension-host containment, OS-jail hardening, leak-test depth) carried as future work — **done ≠ GA**.
- `git mv` proposal + plan `pending/ → done/`.

Doc-only; no code change.

### DIFF
--- a/docs/proposals/done/webchat-git-projects-and-vscode.md
+++ b/docs/proposals/done/webchat-git-projects-and-vscode.md
@@ -1,6 +1,10 @@
 # Webchat git projects + in-browser VSCode
 
-- **State:** proposed; awaiting roundtable + user proposal-gate.
+- **State:** IMPLEMENTED + CLOSED OUT (2026-06-29). Both phases shipped across a
+  series of merged PRs; this proposal is filed to `done/`. See **Status at
+  close-out** at the foot of this file for the shipped surface, the
+  acceptance-criteria audit, and the follow-ups carried as future work
+  (done ≠ GA).
 - **Scope:** deterministic / webchat (Go) + server workspace plane + a new
   per-user code-server sidecar. Not an intelligence-surface proposal (no
   Architecture Charter role).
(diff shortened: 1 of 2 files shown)


JBailes:
**Roundtable accuracy review** (6/6 panelists) — addressed before merge.

The section-level audit was judged structurally honest (PARTIAL + DEVIATION surfaced, not buried), but the panel flagged the rewritten **Risks credential bullet** as overclaiming — stating the *target* vault-only invariant ("only via askpass", "never touches the filesystem", "/proc-clean binding check") as if already *achieved*, against the same page's PARTIAL admission that the HTTPS token also lives in the git child's env. (Some confusion was a `sed` artifact in my review prompt that cut off the status note; the substance stood.) Fixes:

- **Risks bullet rewritten** as *target invariant* + **status by credential type**: SSH = MET (memfd/agent, never on disk); HTTPS = PARTIAL (`GH_TOKEN` in git child env → visible in `/proc/<pid>/environ`; broker-fed askpass is deferred #3). The binding `/proc`-clean leak test is named as the *target* gate, with what `test_webchat_git_leak.c` covers today vs. deferred #6 made explicit.
- **MET lines scoped:** cross-principal isolation → "at the `/v1` route level (WP-A realpath/openat)"; SSH-key leak invariant → "for the implemented + tested agent-load path"; loopback/proxy editor → "route boundary, not yet an OS-jail one". Browser SSH intake (#1) and editor-runs-as-server-UID/PATH + extension-host env inheritance (#4, #5) moved into PARTIAL.
- **§V shipped claim softened:** editor env curated of *server* secrets, but `GIT_ASKPASS`/`SSH_AUTH_SOCK` still inherited by the extension host (#4) and process runs as server UID/PATH (#5).
- **DEVIATION made explicit:** editor is toggleable, the **git surface ships always-on with no opt-out**; dropped the uncited "other default-on closeouts" precedent; added **deferred #7** to re-evaluate default-ON once #4/#5 land + add a git opt-out.

PR #880, opened 2026-06-29
feat(webchat-git): SSH private-key web intake (/v1/git/sshkey)

JBailes:
Closes deferred follow-up **#1** from the webchat-git/VSCode close-out: the per-user in-memory ssh-agent path was built + tested but **unreachable from the browser** (only PAT/token + GitHub OAuth had intake), so a webchat user couldn't use git over SSH.

## Change (end-to-end)
- **`POST`/`DELETE /v1/git/sshkey`** (`rh_git_sshkey`). POST stores an unencrypted OpenSSH/PEM private key in the **caller's own vault** under `(webuser:<name>, "git", "ssh_key")` via `vault_service_set` — dual-wrapped, so the already-shipped `git_ssh_agent_ensure → git_forge_vault_sshkey` loads it into the user's in-memory agent **autonomously** (server-wrap read, no user KEK needed at load time). The key is write-only (never returned), shape-checked (rejects empty / non-`PRIVATE KEY` / passphrase-encrypted), zeroed from the parse buffer after store, and never logged. A **locked vault → 423** so the UI can prompt an unlock. DELETE removes the entry + drops any live agent handle; idempotent.
- **webchat `/api/git/sshkey`** proxy forwarding under the webuser trust headers; documented in `api/openapi-server-v1.yaml` (route↔spec conformance gate).
- **`Projects.tsx`**: a collapsible "SSH private key" field in the Git accounts panel.

## Verification
C + Go build clean; frontend `tsc` clean; webchat proxy tests (forward-with-trust-headers, no-echo-to-browser, empty-reject) green; `server-api-conformance` green. The storage↔reader contract is already covered by `test_git_forge_vault.c`. The full **unlock → store → agent-load → git-over-SSH push** round trip needs a deployed server to exercise (no local code-server/forge here) — flagged honestly rather than claimed.

### DIFF
--- a/api/openapi-server-v1.yaml
+++ b/api/openapi-server-v1.yaml
@@ -4037,6 +4037,73 @@ paths:
           content:
             application/json:
               schema: { $ref: '#/components/schemas/ErrorResponse' }
+  /git/sshkey:
+    post:
+      summary: Store (or replace) the caller's SSH private key for git over SSH
+      description: >
+        Persist an unencrypted OpenSSH/PEM private key in the CALLER's own
+        encrypted vault under (`webuser:<name>`, `git`, `ssh_key`), dual-wrapped so
+        the server can load it into the user's in-memory ssh-agent autonomously
+        (it never touches disk and is never returned to the browser). Storing
+        requires the caller's vault to be unlocked; a locked vault returns 423.
+        Capability: tool:execute.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required: [ssh_key]
+              properties:
+                ssh_key:
+                  type: string
+                  description: Unencrypted OpenSSH/PEM private key (write-only)
+      responses:
+        "200":
+          description: Stored
+          content:
+            application/json:
+              schema:
+                type: object
+                properties: { ok: { type: boolean } }
+        "400":
+          description: Missing or passphrase-encrypted key
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/ErrorResponse' }
+        "403":
+          description: Not a webchat user
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/ErrorResponse' }
+        "413":
+          description: Key too large (> 64 KiB)
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/ErrorResponse' }
+        "423":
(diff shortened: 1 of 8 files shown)


JBailes:

The panel confirmed the load-bearing primitives are correct: the `webuser:` principal gate + `CAP_TOOL_EXECUTE`, the `vault_service_set` arg order matching `git_forge_vault_sshkey`'s lookup, the dual-wrap (server reads via server-KEK without the user's KEK), DELETE idempotency (`VAULT_NO_ENTRY` -> ok), and `VAULT_ERR_LOCKED` -> 423. (Two panelists flagged the cJSON `valuestring` zero-before-`cJSON_Delete` as a double-free -- incorrect: cJSON owns and frees that pointer, the in-place wipe is safe. The synthesis caught the false flag.)

**Blocking -- fixed:**
- No body cap (DoS): added a 64 KiB cap at both layers -- `http.MaxBytesReader` -> 413 in webchat, and an `rq->body_len` guard -> 413 before `cJSON_Parse` in C.

**Addressed:**
- Anchored the key shape-check on the PEM/OpenSSH armor (`-----BEGIN ... PRIVATE KEY-----`) instead of a loose substring, so arbitrary text can't slip through.
- React state no longer retains the key after submit -- `credSSHKey` is cleared on every branch (incl. 423), not just success.
- Go `json.Marshal` error is now checked; DELETE's "stop agent only on clean delete" made explicit in a comment.
- OpenAPI updated with the 413 response.

**Verified non-issues:**
- 423 propagation: `gitRelay` already passes the upstream status through verbatim (`writeJSONError(w, st, ...)`), so the UI's `r.status === 423` branch fires correctly.
- Plaintext residue in `rq->body` / Go string: a pre-existing, family-wide pattern -- the sibling `/v1/git/credentials` token route doesn't wipe `rq->body` either, and the body is TLS-protected in transit + vault-encrypted at rest. Flagged as a cross-route hardening item rather than a risky one-off const-cast wipe here.
- NUL-truncation: an embedded NUL in the JSON yields a malformed key that simply fails at `ssh-add`; no leak (the user's own data), and the 64 KiB cap + anchored armor bound the surface.

Tests: `unit-test-server-http` (re-linked with the new stubs), webchat proxy tests, frontend `tsc`, and server-api conformance all green.

PR #882, opened 2026-06-29
feat(webchat-git): AIMEE_WEBCHAT_GIT opt-out for the git surface

JBailes:
Closes deferred follow-up **#7** from the webchat-git/VSCode close-out: the git surface shipped **always-on with no operator opt-out**, unlike the editor (`AIMEE_WEBCHAT_EDITOR=0`).

## Change
Adds a `git_surface_enabled()` gate (env `AIMEE_WEBCHAT_GIT`, on by default; only the exact value `0` disables) checked first in each of the 9 git route handlers — `rh_workspace_clone` / `git` / `projects` / `session_dir`, `rh_git_credentials`, `rh_git_sshkey`, `rh_git_oauth_github_start` / `poll` / `config` — returning **503** before any other work. The editor surface is independently controllable and untouched, so an operator can ship the image with the in-browser editor but no git intake.

## Test
`unit-test-server-http` asserts disabled→503 (short-circuits ahead of the 403 webuser check) and enabled→403; `docs/gen/configuration.md` regenerated for the new env var. C build + server-http test green; clang-format clean.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -216,7 +216,7 @@ Scalar keys read directly from the config root (not via the CLI allowlist above)
 
 ## Environment variables
 
-The binaries read 132 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
+The binaries read 133 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
 
 ### Paths & assets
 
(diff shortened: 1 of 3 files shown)


JBailes:

Confirmed correct: the env parse (default-on; only exact `"0"` disables; `getenv` deref is immediate, matching the editor's `AIMEE_WEBCHAT_EDITOR` gate, and aimee never `setenv`s at request time).

**Blocking — fixed:**
- *Missed a git route:* `rh_workspace_forge_token` (`POST /v1/workspaces/{id}/forge-token`) mints a brokered forge token but was ungated — exactly the intake an operator setting `AIMEE_WEBCHAT_GIT=0` wants closed. Now gated (helper moved above it); 10 handlers total.
- *session-dir vs editor:* verified `rh_workspace_session_dir` is git-panel-only — the editor path (`/v1/workspace/editor` → `webuser_editor`) never touches it — so gating it under `AIMEE_WEBCHAT_GIT` does **not** break an editor-without-git deployment. Documented in the helper comment.

**Addressed:**
- Test now covers **all 10** gated routes via a table (disabled→503), plus enabled→403 on the context-free routes.
- Documented that 503-before-403 is intentional (the disabled state isn't a secret).

**Considered, kept as-is:** caching the flag at init was suggested, but per-call `getenv` matches the existing editor-flag pattern and keeps the value runtime-togglable (which the regression test relies on); aimee doesn't mutate env at request time, so there's no concurrent-`setenv` window. The 4-line gate is duplicated per handler (vs a macro) to stay consistent with the explicit style of the surrounding route handlers.

C build + `unit-test-server-http` green; clang-format clean; docs/gen regenerated.

PR #886, opened 2026-06-29
feat(fold): P1 Coordinate Closet — verbatim identifier conservation (default-off)

JBailes:
## Summary

First implementation slice of the **deterministic context folding** proposal (design PR #881). Adds the **Coordinate Closet** (§2): a default-off module that conserves exact identifiers (uuids, commit shas, paths, digit-bearing `key=value`, issue/PR refs, `handle:`/`memory:` tokens) extracted from a tool result **before** compaction truncates them, so the agent never loses an id it needs to keep working.

- New self-contained module `src/coord_closet.{c,h}` + `test_coord_closet` (12 cases).
- Deterministic nomination (single left-to-right scan, fixed matcher priority) with provenance `{lane,turn,tool_call,result}` on every coordinate.
- Byte-identical render via total-order sort — independent of insertion/hashmap order (proven by a reversed-array test).
- **No silent loss:** a nominated coordinate that won't fit the cap (`min(budget, raw_len*max_ratio/100)`) yields `COORD_EVICT_FAIL`; the render surfaces a visible `(partial — … omitted)` header and the caller logs at WARN.
- **Injection guard:** user-pasted content is quarantined in `COORD_LANE_USER`, rendered untrusted under a `-- user-supplied (untrusted) --` divider, and can't mint a trusted label.
- **Secrets never echo:** value prefixes (`ghp_`/`sk-`/`AKIA…`/`AIza`/JWT/PEM/…), credential paths, sensitive label names, and a config deny-list are redacted at render time.
- Wired return-only into `agent_compress_tool_result` (`agent_policy.c`) within the `AGENT_TOOL_OUTPUT_MAX` cap. Config nested under `compact.coord_closet.{enabled,budget_bytes,max_ratio_pct,denylist}` — **default-off**. Self-contained (no POSIX-only `strdup`/`strncasecmp`) for the Windows gate.

## Local gates (all green)
`make lint` · `make schema-sync-check` · server+kb build · `make -j unit-tests` (full suite incl. new test) · docs regenerated.

## Roundtable code review

Ran a 6-panelist C code roundtable. Resolved every real blocker:
- ratio_cap collapse for small raw → saturating form;
- `render_line` 512-byte fixed buffer truncation → removed (snprintf-sizing, no truncation);
- combined body+closet vs `AGENT_TOOL_OUTPUT_MAX` → render budget bounded by the hard cap;
- locale-dependent ctype → ASCII-only helpers (determinism);
- SHA matcher prefix-of-longer-run → right-boundary required;
- eviction only DEBUG-logged → visible partial marker + WARN;
- secret-redaction gaps → extended prefixes + label-name redaction;
- user-lane divider added.

Two findings were assessed **not real** and documented: the qsort comparator *is* a strict weak order (lexicographic over totally-ordered fields; per-lane dedup prevents equal-but-distinct entries), and the all-AGENT-lane wiring is safe for P1 because the closet is a subset of the same tool result the agent already sees — real per-source lane assignment lands with the transcript fold in P2.

## Not for merge
Opened for your review per request; merge gated on your approval + roundtable sign-off.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -162,7 +162,7 @@ Set in the config JSON as `{"<section>": {"<key>": ...}}`. Keys are derived from
 - **`auxiliary`** — _Auxiliary (cheap/background) model used for side tasks._ Keys: `default_max_tokens`, `default_model`, `default_provider`, `enabled`, `tasks`
 - **`cache_shaping`** — _Prompt-cache shaping._ Keys: `enabled`, `min_chars`
 - **`charter`** — _Operating charter: values, constraints, safety axioms, tone._ Keys: `hard_constraints`, `safety_axioms`, `tone_boundaries`, `values`, `working_profile_drift_limit`
-- **`compact`** — _Transcript compaction thresholds._ Keys: `enabled`, `head_bytes`, `per_tool`, `tail_bytes`, `threshold`
+- **`compact`** — _Transcript compaction thresholds._ Keys: `coord_closet`, `enabled`, `head_bytes`, `per_tool`, `tail_bytes`, `threshold`
 - **`computer_use`** — _Computer-use (browser) tool settings._ Keys: `allowed_domains`, `default_navigation`, `enabled`, `redact_sensitive_screenshots`
 - **`concurrency`** — _Per-model / per-provider concurrency limits._ Keys: `default`, `per_model`, `per_provider`, `preempt`
 - **`context`** — _Context-engine selection._ Keys: `engine`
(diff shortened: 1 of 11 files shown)


JBailes:

Two passes (initial + confirmation, 6 panelists each).

**Initial pass — 8 blocking, all fixed:** saturating ratio cap (small-raw); removed the 512-byte fixed render buffer (snprintf-sizing, no truncation); render budget bounded by `AGENT_TOOL_OUTPUT_MAX` so body+closet never exceed the cap; ASCII-only ctype helpers (locale-independent determinism); SHA right-boundary; visible `(partial — … omitted)` marker + WARN log on eviction; extended secret prefixes + sensitive-label redaction; user-lane `(untrusted)` divider.

**Confirmation pass — no blockers** (only suggestions/nits; the verify stage contradicted the re-raised correctness concerns). Folded in three cheap quality nits: uuid/sha right-boundary parity (`token_continues`), kv trailing-dot trim, and an accurate trust-boundary comment. Added a regression test (`boundary_and_punctuation_nits`).

**Assessed not-real (documented):** the qsort comparator *is* a strict weak order (lexicographic over totally-ordered fields; per-lane dedup prevents distinct-but-equal entries); all-AGENT-lane wiring is safe for P1 because the closet conserves values from the same tool result as the body (same trust level), with real per-source lane deferred to P2.

13 unit cases green; `make lint`/`schema-sync`/build/full-unit-tests green locally.

PR #887, opened 2026-06-29
feat(fold): P1.5 per-model fold-budget resolver + pipeline-order spike

JBailes:
## Summary

Second slice of deterministic context folding (design PR #881; P1 = #886). Lands the §7 budget resolver and the pipeline-order spike — the prerequisite for P2.

- **`src/fold_budget.{c,h}`** — `fold_budget_resolve(model_id, cfg, out)`: a pure, deterministic function of `(model, config)` that resolves the fold knobs (context window, retained band, tail cap, pressure ceiling, prefix saturation, closet budget). Window from `model_context_window()`; unknown models fall back to a configured/default window (`known_model` flag records which). Percentages are window-relative, clamped.
- **`docs/dev/fold-pipeline-order.md`** — the spike that, before any P2 code, pins: the normative 5-stage delegate-path pipeline order and the exact `agent_runtime.c` seam; the single-owner `payload_rewrite` span registry (internal API, post-stage-4 byte capture, `(ptr,len)` lifetime) + the CI scan that enforces it; and the rationale for deferring the budget config knobs to P2 (consumed there) rather than landing inert config now.
- **`test_fold_budget`** — known/unknown models, overrides, determinism, bad args.

## Local gates
`make lint` · build (server+kb) · `make -j unit-tests` (full suite incl. new test) — all green.

## Review
C code roundtable in progress; I'll post the summary and fold in any blocking fixes before merge.

## Note
No runtime behavior change — the resolver has no caller until P2; default-off feature. Branches off `testing` (includes merged P1).

### DIFF
--- a/docs/dev/fold-pipeline-order.md
+++ b/docs/dev/fold-pipeline-order.md
@@ -1,0 +1,98 @@
+# Fold pipeline-order spike (P1.5)
+
+De-risks the P2 rolling-fold + fold-freeze implementation by pinning **where** the
+transcript fold runs and **how** it contributes to the single cache-prefix owner,
+before any P2 code is written. Companion to `fold_budget.{c,h}` (the §7 resolver
+landed in the same slice). Ruling from the plan roundtable: target the **delegate
+path first** (it already owns `payload_rewrite`); the ingress proxy path is a later
+slice.
+
+## Normative pipeline order (delegate path)
+
+`agent_runtime.c` is the actor throughout: it calls each step and owns the buffers.
+Six steps per turn in `agent_execute_with_tools_internal()`
+(`src/posix/agent_runtime.c`); step 1 is pre-stage input, steps 2–6 are actions:
+
+1. **raw DB1 transcript** — append-only `messages` cJSON array (never mutated).
+2. **session compaction** — `maybe_compact_before_request()` (≈`agent_runtime.c:677`).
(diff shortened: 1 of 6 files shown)


JBailes:

6 panelists. **One blocking bug, fixed:** `pct_of` did its intermediate multiply in `long`, which is 32-bit on the Windows (LLP64) CI gate / ILP32 → signed-overflow UB for large window×pct reachable from P2 operator config. Now uses `int64_t` and clamps before the `int` cast.

**Quality fixes folded in:** clamp `closet_budget_tokens` to the window (§7 single enforcement point); `known_model` → `is_known`; tests now derive expected windows from `model_context_window()` (no longer brittle to model-data churn) and add clamp coverage (pct>100 → window; oversize closet → window); header documents the `<=0 means default` sentinel, the pct>100 clamp, and that the resolver does mechanical translation only (P2 enforces cross-field ordering).

**Pipeline-order doc sharpened** (tightens the P2 contract): explicit actor (`agent_runtime.c` throughout; fold produces-and-returns, runtime registers); semantic-freeze vs deterministic 1:1 render distinction; `payload_rewrite_register_span` returns `int` and is state-scoped (`payload_rewrite_state_t*`) with mandatory idempotent reset; the single-owner CI-lint **bootstrap gap** called out (the registry PR must be reviewed for stray `fnv1a_update` callers since the lint ships in the same PR); and an explicit "NOT in this slice" deferred-knobs list.

6 unit cases green; `make lint`/build/full-unit-tests green locally.

PR #889, opened 2026-06-29
feat(fold): P2a rolling-fold module — deterministic turn skeletonization (default-off)

JBailes:
## Summary

Core algorithmic slice of deterministic context folding (design #881; P1=#886, P1.5=#887). **P2a = the rolling-fold module, pure and default-off**; the live request-path wiring + byte-identical freeze + `payload_rewrite` span registry are **P2b** (next slice). Splitting keeps the live-path change reviewable on its own.

`src/context_fold.{c,h}` — `context_fold_view(messages, cfg, &out)`:
- Folds a contiguous prefix of old Anthropic-shaped messages into a synthetic summary: one skeleton line per turn / tool call + a Coordinate Closet (§2) of the exact identifiers from the folded region.
- Returns a **new** array (synthetic user preamble+skeleton+closet → synthetic assistant ack → deep copy of the retained tail). Input is never mutated.
- **Atomic tool pairs / valid alternation:** the boundary is chosen at a *clean user turn* (role user, not a tool_result), so a `tool_use` and its `tool_result` are never split and the tail begins with a user message. No clean boundary with enough to fold → no fold.
- Pure (cJSON + dstr + coord_closet); deterministic; default-off.

`test_context_fold` (5 cases): fold structure, verbatim identifier conservation, non-destructive + byte-identical determinism, clean-boundary selection, disabled/too-short no-ops.

## Local gates
`make lint` · build (server+kb) · `make -j unit-tests` (full suite incl. new test) — all green.

## Note
No runtime behavior change (no caller until P2b); default-off. Branches off `testing` (includes P1+P1.5). Review in progress via roundtable; blocking fixes folded in before merge.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -290,7 +290,7 @@ MCP_GIT_OBJS = $(MCP_GIT_SRCS:%.c=$(OBJDIR)/%.o)
 # --- Layer 0: Foundation (db, config, util, text, render, log, platform, cJSON) ---
 CORE_SRCS = db1/db.c db1/db_schema.c config.c config_sections.c config_database.c config_learning.c config_memory.c config_charter.c config_trigger.c config_kb_maintenance.c config_kb_curator.c config_server_api.c config_skills.c config_save.c config_mode.c config_fields.c toolset.c util.c util_url.c forge_credentials.c forge_app_token.c workspace_mirror.c report_enrichment.c delivery_target.c text.c render.c markdown.c json_fluent.c code_audit_graph.c log.c shutdown_forensics.c dstr.c yaml.c diff.c sse_parser.c \
             aimee_home.c shared/kb_paths.c provider_client.c kb_curator_provider.c \
-            compact.c coord_closet.c fold_budget.c platform_random.c slop_detect.c proxy_bootstrap.c reasoning_cap.c \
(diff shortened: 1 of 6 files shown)


JBailes:

6 panelists, 0 failed. **4 blocking (all defensive-hardening, fixed):**
- unchecked `cJSON_Create*`/`cJSON_Duplicate` NULL on OOM → now NULL-checked with full cleanup (delete partials, `dstr_free`) and a clean no-fold return;
- `cJSON_GetArrayItem` NULL on a sparse array → explicitly skipped in both the folded-bytes and retained-copy loops;
- `memset(out,0)` could leak a reused result → documented that `*out` is overwritten and a reused *folded* result must be `fold_result_free`d first;
- `retained + min_fold` could overflow near INT_MAX → rewritten as overflow-safe comparisons (`retained >= count || count - retained < min_fold`).

**Suggestions folded in:** UTF-8-boundary-safe excerpt truncation (never emits a split multibyte char into a JSON value); documented `coord_closet_nominate` copies bytes (no retained pointer); documented determinism depends on `coord_closet_render`'s total order; +2 edge tests (`retained_ge_count`, `odd_shapes_no_crash` covering missing-content + array-content tool_result).

The atomic-boundary design (fold only at a clean user turn) was accepted — no design changes. 7 unit cases green; `make lint`/build/full-unit-tests green locally.

PR #894, opened 2026-06-29
feat(fold): P2b wire rolling fold into the delegate request path (default-off)

JBailes:
## Summary

Wires the P2a rolling-fold module into the live Anthropic delegate request path — the value-delivery step (token reduction). Default-off.

- `posix/agent_runtime.c`: at **both** Anthropic build sites (primary + fallback rebuild), build a folded view of the message array and build the request from it when `fold.enabled`. The view is freed **only after the request is serialized** — `agent_build_request_anthropic` can keep a *reference* to the messages (fallback path), so the view must outlive `req`; frees are placed on every post-build path including the gateway-pipeline-failure break. Non-Anthropic providers are untouched (the fold is Anthropic-shaped).
- `build_fold_view()` maps config → `fold_config_t` and renders the closet **within the local config's lifetime** (the `denylist` pointer never escapes).
- New **`fold`** config section (default-off): `fold.{enabled,retained_msgs,min_fold_msgs,excerpt_bytes}`; the fold's Coordinate Closet reuses the existing `compact.coord_closet.*` policy (one closet policy). Docs regenerated.

## Scope / honesty
Delivers token reduction. The byte-identical **fold-freeze + `payload_rewrite` span registry (§3)** — which keeps the folded prefix cache-stable across turns — is **P2c**; without it, folding reduces tokens but can move the cache boundary, so this stays default-off until P2c lands. (Documented in `docs/dev/fold-pipeline-order.md`.)

## Local gates
`make lint` · server+kb build · `make -j unit-tests` (full suite) — all green. When `fold.enabled=0` (default) `build_fold_view` returns immediately → behavior byte-identical to before.

## Note
Live end-to-end token/cache validation (plan §E, pve testbed) is best run once P2c lands; gated default-off until then. Review via roundtable in progress.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -157,7 +157,7 @@ Structured options (arrays, nested objects — e.g. `ensemble.reference_models`)
 
 > **Undocumented** (add to `CFG_KEY_DESC` in gen-reference-docs.py): `wfe_live_forge_enabled`
 
-## Config-file sections (49)
+## Config-file sections (50)
 
 Set in the config JSON as `{"<section>": {"<key>": ...}}`. Keys are derived from the section parsers in `src/config*.c`; a key shown as a bare name that is itself a nested object is noted in the section description (see *Coverage & limitations*).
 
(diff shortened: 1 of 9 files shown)


JBailes:

6 panelists, 0 failed. Two items were flagged "blocking"; both **verified false positives** against the real code:
- *"`config_parse_fold_section` not wired"* — it **is** wired (`config.c:1011`, declared `config_sections.h:23`); the panel couldn't see it because the diff I shared was scoped to 3 files (excluded `config.c`).
- *"`config_load` leaks without `config_free`"* — `config_free` **does not exist** anywhere and `config_t` has **zero pointer members**; it's a plain stack struct used as `config_t cfg; config_load(&cfg)` throughout (e.g. `agent_policy.c`). No leak.

**Suggestions folded in:**
- Reordered `track_anthropic_payload_rewrite` to run **after** `build_fold_view` and observe the actually-sent (folded) message array — at both build sites — so token accounting matches the payload.
- Added the load-bearing ordering comment at both `cJSON_Delete(req)` → `fold_result_free` sites (req may hold a non-owning reference into the fold view).
- Documented `fold_result_free` as NULL-safe on zeroed/no-fold results (justifies the unconditional call on non-Anthropic branches) and that `cfg->closet.denylist` is borrowed only for the duration of `context_fold_view`.

The default-off `config_load`-per-request note is acknowledged: `config_load` is cached (mtime-gated), and this matches the existing `agent_compress_tool_result` pattern.

Reordering verified: server+kb build, `make lint`, `unit-test-agent`, full unit-tests green.

PR #895, opened 2026-06-29
feat(fold): P2c fold-freeze — pin the fold boundary across turns (default-off)

JBailes:
## Summary

§3 fold-freeze: keep the folded prefix **byte-identical turn-to-turn** so the provider prompt cache stays warm. The fold is already deterministic (P2a), so the cache-busting behaviour was the boundary *sliding* every turn — freeze pins it.

- `context_fold` gains `fold_freeze_t` (active, frozen_split, tail_cap_msgs, epochs) + a `freeze` param on `context_fold_view`. When the frozen boundary is still a clean user turn and the un-folded tail is within `tail_cap_msgs`, the fold **reuses** it — re-folding the identical prefix `[0,frozen_split)` yields a byte-identical synthetic prefix. The boundary advances only at an **epoch** (tail exceeds cap). `out->reused_boundary` reports cache-warm reuse; `NULL` freeze = P2a per-turn behaviour.
- `posix/agent_runtime.c` threads a **per-run** `fold_freeze_t` across the turn loop into both Anthropic build sites; honoured only when `fold.freeze.enabled`.
- Config `fold.freeze.{enabled,tail_cap_msgs}`, default-off; docs regenerated.
- `test_context_fold`: freeze pins the boundary (byte-identical prefix across an appended turn) and advances exactly one epoch when the tail exceeds the cap.

## Scope (honest)
The provider cache benefit here comes **directly** from emitting byte-identical prefixes, independent of aimee's internal `payload_rewrite` hash. The `payload_rewrite` **span-registry integration + single-owner CI lint** (the §3 substrate) remain a documented follow-up tied to the `ingress-compression-and-cache-alignment` convergence (one shared span-registry PR). Within-run freeze assumes the folded prefix isn't mutated mid-run by session compaction; if it is, the re-fold differs (cache miss) but correctness holds — documented.

## Local gates
`make lint` · server+kb build · `make -j unit-tests` (full suite, 8 context_fold cases incl. freeze) — green. Default-off: `fold.freeze.enabled=0` ⇒ P2b behaviour unchanged.


JBailes:

6 panelists, 0 failed. **2 blocking, both fixed:**
- **Reuse guard checked the index, not the prefix *content*.** Compaction can mutate `messages[0..frozen_split)` while preserving indices, so "reuse" could fold a different prefix and falsely claim a warm cache. Fixed: `fold_freeze_t` now carries an FNV-1a `prefix_digest`; reuse requires the current folded-prefix digest to match — a mid-run mutation forces an epoch (re-fold). Regression test `freeze_prefix_mutation_breaks_reuse` added. (The digest is computed in the same pass that sizes the folded region, so no extra serialization.)
- **`tail_cap < retained` ⇒ re-epoch every turn** (boundary drift). Fixed: `tail_cap` is clamped to ≥ `retained`.

**Suggestions folded in:** freeze state is now committed **after** a successful build (an OOM return leaves the prior boundary intact), and `epochs++` only counts genuine advances; `reused_boundary` is now consumed via a debug log in `build_fold_view`.

Verified: prefix-mutation forces an epoch; shared freeze across both build sites in one turn reuses (digest matches) with no double-advance; default-off (freeze NULL) is exactly P2b. 9 context_fold cases green; `make lint`/build/full-unit-tests green.

PR #896, opened 2026-06-29
feat(webchat-git): git HTTPS token via inherited memfd, not /proc/environ (#3A)

JBailes:
Closes deferred follow-up **#3** (git_ops slice — #3A): the webchat git ops injected the forge token as `GH_TOKEN` in the git child's environment, so it was visible in `/proc/<pid>/environ`. This moves the `git_ops` path to an **fd-fed askpass** so the token never enters the environment.

## Mechanism
- **`git_cred_inject` FD mode** (`out_token_fd`): the resolved HTTPS token is written to an anonymous **CLOEXEC memfd** (never a named path); the env carries `AIMEE_GIT_TOKEN_FD=<n>` (a *number*, not the secret) instead of `GH_TOKEN`. `memfd_create` failure **fails closed** (drops the token) rather than leaking it to the env.
- **`safe_exec` `_fd_` variant**: hands the child exactly one fd at a fixed number (`GIT_CRED_TOKEN_TARGET_FD=21`) with CLOEXEC cleared via `dup2`. The source memfd stays CLOEXEC in the parent, so a **concurrent exec on another thread can't inherit it**.
- **askpass shim** reads `/proc/self/fd/$AIMEE_GIT_TOKEN_FD` (re-readable), falling back to `$GH_TOKEN` so the long-lived **editor path keeps working until it migrates (#4)**. git's askpass invocation is otherwise unchanged.
- `run_git` uses FD mode; **clone** (`git_project`), `workspace_turn`, and the `aimee git` CLI query stay env-mode for now (signature updated; clone migration is #3A.2).

## Validation (local)
`test_git_cred_inject` extended: the FD-mode env has `AIMEE_GIT_TOKEN_FD` + **no** `GH_TOKEN`, and the secret appears **nowhere** in the env strings; and a child exec'd via the `_fd_` variant reads the token through `/proc/self/fd/<target>` (what the askpass does) while its `/proc/self/environ` contains only the fd **number**, never the token. git's askpass path is unchanged from the shipped, working behavior — only the token *source* moved (env → fd).

Roundtable security review run on the diff; summary to follow as a comment.

### DIFF
--- a/src/forge_credentials.c
+++ b/src/forge_credentials.c
@@ -357,10 +357,23 @@ const char *forge_cred_askpass_shim(void)
       path[0] = '\0';
       return NULL;
    }
+   /* Password source, in order: a token delivered over an inherited file
+    * descriptor (AIMEE_GIT_TOKEN_FD — re-readable via /proc/self/fd, so the
+    * secret never enters this process's environment or /proc/<pid>/environ),
+    * else the legacy GH_TOKEN env var (still used by the long-lived editor path
+    * until it migrates). The username is always the GitHub x-access-token. */
+   /* AIMEE_GIT_TOKEN_FD must be all-digits before it is used as a /proc/self/fd
+    * path component (defense-in-depth); a missing/non-numeric value, or a read
+    * that yields nothing, falls back to the legacy GH_TOKEN env var. cat reopens
+    * /proc/self/fd/<n>, so the memfd is read from offset 0 on every invocation. */
    fputs("#!/bin/sh\n"
          "case \"$1\" in\n"
          "*[Uu]sername*) echo x-access-token ;;\n"
-         "*) printf '%s\\n' \"$GH_TOKEN\" ;;\n"
+         "*) case \"$AIMEE_GIT_TOKEN_FD\" in\n"
+         "     ''|*[!0-9]*) printf '%s\\n' \"$GH_TOKEN\" ;;\n"
+         "     *) cat \"/proc/self/fd/$AIMEE_GIT_TOKEN_FD\" 2>/dev/null "
+         "|| printf '%s\\n' \"$GH_TOKEN\" ;;\n"
+         "   esac ;;\n"
          "esac\n",
          f);
    fclose(f);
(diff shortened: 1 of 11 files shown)


JBailes:
**Roundtable security review** (6/6 panelists). The panel **confirmed the fd-inheritance mechanics are correct**: `MFD_CLOEXEC` source (no concurrent-exec leak to other children), child-only `dup2` to fixed fd 21, the `inherit_fd==target_fd` `F_SETFD` path, memfd closed on every path (no-token / write-fail / build-fail / post-exec), `memfd_create` failure fails-closed by dropping the token, and `/proc/<pid>/fd/21` readable-by-same-uid matching the already-accepted `/proc/<pid>/environ` boundary.

**Two "blocking" claims — both verified FALSE:**
- *"memfd shared read-offset → 2nd askpass read empty"*: empirically disproven. `cat /proc/self/fd/N` **reopens** the memfd (fresh offset 0), so it's re-readable on every invocation. A standalone test read the same token 3× via `cat /proc/self/fd/21`; the panel confused this with reading the *inherited* fd directly (shared offset → empty), which the shim does not do.
- *"build_env may emit GH_TOKEN in fd-mode"*: `build_env` branches on `token_fd_target` and emits `AIMEE_GIT_TOKEN_FD` (a number), never `GH_TOKEN`, when fd-mode is on. The unit test asserts no `GH_TOKEN` and that the secret appears **nowhere** in the env strings. (build_env wasn't in the focused prompt, hence the assumption.)

**Valid nits — fixed:**
- **Uninitialized-read UB**: `token[]` is now zero-initialized and wiped unconditionally (was `if (have_token || token[0])` reading an uninitialized byte).
- **memfd name**: changed `"aimee-git-token"` → opaque `"c"` so `/proc/<pid>/fdinfo` doesn't advertise a credential.
- **askpass hardening**: `AIMEE_GIT_TOKEN_FD` is now numeric-validated before use as a `/proc/self/fd` path component, with a clean `$GH_TOKEN` fallback on any empty/failed read.

**Acknowledged, low-priority:** the existing `wipe()` already uses a `volatile` byte loop (not elidable). A lingering fd-21 in git subprocesses would require a daemonizing `credential.helper`, which `git_ops` does not configure.

PR #898, opened 2026-06-29
feat(fold): P3 register grammar — classify assistant turns (default-off)

JBailes:
## Summary

§6 register grammar: classify assistant turns so the fold's skeleton preserves which folded turns were settled conclusions vs transient chatter (and P5 can gate episode-seal harvesting on settled turns).

- `src/fold_register.{c,h}` — `fold_register_parse()` recognizes a leading glyph (🔍 ▶ 🏁 ⚠ ❓) or a bracket tag (`[verdict]`/`[done]`, `[hazard]`/`[warning]`, `[executing]`/`[exec]`, `[blocked]`, `[wip]`) → in-progress / executing / verdict / hazard / blocked. Deterministic, ASCII/UTF-8-safe; untagged → in-progress (soft, so the fold is correct with or without registers). `fold_register_is_settled()` flags verdict/hazard.
- **Consumer:** `context_fold` annotates folded assistant skeleton lines (`assistant/verdict:`) when `fold.register_enabled` — the folded summary now shows which turns were settled. Default-off: `register_enabled=0` ⇒ skeleton byte-identical to P2.
- config `fold.register_enabled` (default-off); docs regenerated.
- `test_fold_register` (glyphs, bracket tags, defaults, is_settled) + a `context_fold` register-annotation test.

## Local gates
`make lint` · server+kb build · `make -j unit-tests` (full suite; fold_register 3 cases + context_fold 10 cases) — all green.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -182,7 +182,7 @@ Set in the config JSON as `{"<section>": {"<key>": ...}}`. Keys are derived from
 - **`dedup`** — _Response deduplication._ Keys: `enabled`, `window_seconds`
 - **`dogfood`** — _Session capture for dogfood data._ Keys: `commit_raw`, `enabled`, `inline_tagging`, `log_dir`
 - **`ensemble`** — _Roundtable ensemble panel + aggregator._ Keys: `aggregator`, `max_cost_usd`, `min_successful`, `reference_models`, `reference_personas`
-- **`fold`** — `enabled`, `excerpt_bytes`, `freeze`, `min_fold_msgs`, `retained_msgs`
+- **`fold`** — `enabled`, `excerpt_bytes`, `freeze`, `min_fold_msgs`, `register_enabled`, `retained_msgs`
 - **`guardrails`** — _Semantic guardrail policy._ Keys: `blast_radius`, `semantic`
 - **`identity`** — _Working-profile identity injection._ Keys: `working_profile_injection`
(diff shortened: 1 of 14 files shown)


JBailes:

6 panelists (1 dropped). The panel reviewed the diff captured at submit time, which predated a fix I'd already pushed — so the headline OOB items were resolved before the verdict landed:

- **`memcmp` OOB on short glyph inputs (1–3 bytes)** — **already fixed**: switched to `strncmp`, which stops at the NUL mismatch and never reads past a short buffer; `test_short_inputs_safe` covers lone/partial `0xF0`/`0xE2` leads.
- **`tag_at` OOB** — assessed **false positive**: when `p[i]` is the NUL terminator it mismatches the (non-NUL) tag char and returns immediately, reading only the in-bounds terminator, never beyond.

Genuine improvement folded in:
- **Bracket tags now require the closing `]`** (`btag`), so prefix false-positives are gone (`[executable]`/`[verdicts]`/`[doner]` no longer match `exec`/`verdict`/`done`); `[exec]`/`[executing]` still match exactly. New `test_bracket_anchor`.
- Documented `a_lower`/`btag` as ASCII-only (safe for the bracket grammar).

Glyph byte sequences are validated by `test_glyphs` (parses the real UTF-8 bytes). Default-off (`register_enabled=0`) leaves the skeleton byte-identical to P2, asserted by the off-branch of the `context_fold` annotation test. All fold tests + `make lint`/build/full-unit-tests green.

PR #900, opened 2026-06-29
feat(webchat-git): open PR via in-process GitHub REST, not gh (#3B)

JBailes:
Closes deferred follow-up **#3B**: the open-PR op shelled out to `gh pr create`, which reads `GH_TOKEN` from the child's environment — the **last** spot a webchat git action put the forge token in a child's `/proc/<pid>/environ`. Replaced with an **in-process REST call** so the token rides the `Authorization` header in aimee-server memory and never reaches a child.

## Change
- **`git_pr_api.c`** — `git_pr_create_via_api`: resolves owner/repo from the repo's `github.com` origin, `head`=current branch, `base`=origin default branch (fallback `main`), title defaults to the last commit subject; POSTs to `/repos/{o}/{r}/pulls` via `agent_http_post_content_type` with a **Bearer token resolved vault-first** (per-host github token, else the principal's personal vaulted token). Token is **wiped from both buffers** after the call. GitHub origins only — a non-github origin returns a clean error (also resolves the #2 review's opaque-error note). owner/repo are charset-validated before URL interpolation; `repo_dir` rejects a stray quote.
- **`git_ops` `pr` op** now calls the REST helper (intercepted like `commit`, no child exec); the `gh` argv path is removed. Output is the new PR's `html_url`.

## Validation
The **exact REST request** the code builds (`Authorization: Bearer` + `Accept: application/vnd.github+json` + `Content-Type: application/json` + `{title,head,base,body}`) returns **HTTP 201** against the real GitHub API — a throwaway PR (#899) was created then closed. The aimee HTTP client itself is already proven (it drives GitHub OAuth + LLM providers). `unit-test-git-ops` green via a pr-api stub.

Security roundtable run on the diff; summary to follow.


JBailes:
**Roundtable security review** (6/6 panelists) — two blockers + suggestions.

**Blocking — fixed:**
- **Host substring-match (real integrity bug):** `parse_github_slug` used `strstr(url,"github.com")`, so `evilgithub.com` / `github.com.evil.com` / `attacker.tld/github.com/...` would smuggle an arbitrary owner/repo into the hard-coded `api.github.com` URL and open a PR on an unintended repo with the principal's token. Rewrote it to parse scheme://[user@]host[:port]/path and the SCP `git@host:path` form and require an **exact** host (`github.com`/`www.github.com`). Validated: all of `evilgithub.com`, `github.com.evil.com`, `attacker.tld/github.com/`, `git@evil.com:github.com/` are **rejected**; https/ssh/scp github forms accepted.

**Blocking — already resolved (panel saw the pre-switch code):**
- *Token resolution order:* the reviewed artifact still called `git_host_resolve_token` then `git_forge_vault_token` directly. The merged code routes through **`git_cred_inject_resolve_token`** — the ONE credential policy — so the precedence (preferred → per-host vault → principal vault → server identity) is the documented unified ladder, not a hand-rolled order. (This switch was also required to pass the `git-cred-centralized` lint gate.)

**Suggestions — addressed/verified:**
- owner/repo must now **start alphanumeric** (rejects `-bad`/`.evil`), in addition to the `[A-Za-z0-9._-]` charset.
- *Auth-header logging:* audited `agent_bridge.c` — `auth_header` is only written into the request buffer; the HTTP client logs host + byte-count only, never headers. Not logged.

PR #901, opened 2026-06-29
feat(fold): P4 fold recall — page folded coordinates back in on re-touch (default-off)

JBailes:
## Summary

§4 fold recall: when the agent re-references a coordinate that was folded away, surface a hint so the body can be paged back in (via the existing `code_span_get`/`memory_get` tools).

- `src/fold_recall.{c,h}` — a per-run **page table** of folded-away coordinates (paths, `handle:`/`memory:` ids) + a **re-touch detector** with a **residency TTL** (anti-thrash: a key isn't re-surfaced within `ttl_turns`). Pure, deterministic.
- **Live wiring** (`posix/agent_runtime.c`, default-off): a per-run `fold_recall_index_t` threads across the turn loop. On a folding turn, `build_fold_view` detects re-touch of prior-folded coordinates in the newest user message, injects the hints into the fold preamble, then indexes this fold's path/handle coordinates for future turns. Detection precedes indexing (a just-folded key can't trivially match the same turn).
- config `fold.recall.{enabled,ttl_turns}` (default-off); docs regenerated.
- `test_fold_recall`: add/dedup, detect, TTL anti-thrash, NULL/empty.

## Scope
The hint points the agent at the on-demand recovery tools rather than auto-fetching bodies inline (those resolvers need project/kb context and are the agent's existing surface). Recall detection runs on folding turns (the common case in long sessions). Default-off ⇒ no behaviour change.

## Local gates
`make lint` · server+kb build · `make -j unit-tests` (full suite) — all green.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -182,7 +182,7 @@ Set in the config JSON as `{"<section>": {"<key>": ...}}`. Keys are derived from
 - **`dedup`** — _Response deduplication._ Keys: `enabled`, `window_seconds`
 - **`dogfood`** — _Session capture for dogfood data._ Keys: `commit_raw`, `enabled`, `inline_tagging`, `log_dir`
 - **`ensemble`** — _Roundtable ensemble panel + aggregator._ Keys: `aggregator`, `max_cost_usd`, `min_successful`, `reference_models`, `reference_personas`
-- **`fold`** — `enabled`, `excerpt_bytes`, `freeze`, `min_fold_msgs`, `register_enabled`, `retained_msgs`
+- **`fold`** — `enabled`, `excerpt_bytes`, `freeze`, `min_fold_msgs`, `recall`, `register_enabled`, `retained_msgs`
 - **`guardrails`** — _Semantic guardrail policy._ Keys: `blast_radius`, `semantic`
 - **`identity`** — _Working-profile identity injection._ Keys: `working_profile_injection`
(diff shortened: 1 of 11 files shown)


JBailes:

6 panelists (1 dropped). Triaged:

- **Substring false-positives (`/x` matches `/xyz`, `memory:42` matches `memory:420`)** — **real, fixed**: detection now uses whole-coordinate-token matching (`contains_token`), requiring the match not be extended by a coordinate char (`[A-Za-z0-9/._:-]`) on either side. New `whole_token_match` test.
- **Double `build_fold_view` per turn → same-turn self-match** — **real, fixed**: recall now runs only at the **primary** build site; the fallback rebuild passes `NULL` recall, so the index is detected/updated once per turn.
- **`cJSON_CreateString` NULL on OOM** — **fixed**: the replacement item is NULL-checked; on OOM the original preamble content is left intact (the UAF claim was a false positive — `cJSON_CreateString` copies, so freeing the `dstr` after is correct).
- **realloc partial-OOM** — assessed **memory-safe** (one panelist concurred: `cap` stays at the smaller array's size so writes never go OOB); restructured to commit each realloc as it succeeds, with a comment making the invariant explicit.
- **Free on all exit paths** — verified: the turn loop exits only via `break` into the post-loop free; no early `return` inside the loop; `fold_recall_index_free` is NULL-safe regardless.

Detection-before-indexing, default-off==prior-behaviour, and determinism were confirmed correct by the panel. All fold tests + `make lint`/build/full-unit-tests green.

PR #902, opened 2026-06-29
feat(webchat-git): clone token via memfd (#3A.2) + leak-test no-environ invariant (#6)

JBailes:
Extends the FD-mode token delivery (#3A) to the **clone** path and strengthens the leak test.

- **`git_project` clone** now uses the memfd FD mode (`safe_exec_capture_cwd_env_fd_timeout` with an inherited token fd) instead of `GH_TOKEN`-in-env — so the clone child's `/proc/<pid>/environ` no longer carries the token, matching `run_git`. (A 300s timeout also replaces the previous unbounded clone exec.)
- **`test_webchat_git_leak`**: in addition to the existing "canary plaintext in no file under `AIMEE_HOME`" + cross-principal denial, now asserts the **FD-mode env carries no `GH_TOKEN`** and the canary appears in **no env string** — the assembly-layer form of "not in `/proc/<pid>/environ`" (the exec-time proof lives in `test_git_cred_inject`'s `safe_exec` fd test).

The mechanism itself was security-reviewed + merged in #3A; this reuses it for clone. The remaining env-mode path is the long-lived **editor** (`webuser_editor`), deferred to the extension-host work (#4) where node's fd-inheritance can be validated on a live code-server. Real-push + SIGKILL-during-push is a deploy-env check (the memfd is anonymous → a SIGKILL leaves no orphan by construction).

Tests: `unit-test-git-project` + `unit-test-webchat-git-leak` green. Roundtable run; summary to follow.

### DIFF
--- a/src/server/git_project.c
+++ b/src/server/git_project.c
@@ -109,9 +109,11 @@ int git_project_clone(const char *principal, const char *url, const char *name, 
     * already-stored vault token > the principal's own vaulted token > the
     * server's own identity (App token / AIMEE_FORGE_TOKEN); plus the principal's
     * vaulted SSH agent. The token crosses only via GIT_ASKPASS, never argv. */
-   /* TODO(#3A.2): migrate clone to the memfd fd-mode (out_token_fd) so the token
-    * is not in the clone child's environment; env mode (GH_TOKEN) for now. */
-   char **envp = git_cred_inject_build_env_for_repo(principal, url, NULL, token, environ, NULL);
+   /* FD mode (#3A.2): the HTTPS token rides an inherited memfd (token_fd), so the
+    * clone child's /proc/<pid>/environ never carries it (parity with run_git). */
+   int token_fd = -1;
+   char **envp =
+       git_cred_inject_build_env_for_repo(principal, url, NULL, token, environ, &token_fd);
    /* Clone over HTTPS, never SSH: our credential model is per-host HTTPS tokens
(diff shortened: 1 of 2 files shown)


JBailes:
**Roundtable review** (6/6, 0 failed) — no blockers; the panel called the clone fd-mode migration "structurally sound". Triage:

- *NIT (envp NULL but token_fd≥0 partial-success):* cannot occur — `git_cred_inject_build_env_for_repo` closes the memfd and resets `token_fd` to -1 if env assembly fails (`if (!envp && token_fd >= 0) { close; -1; }`), so `envp==NULL ⇒ token_fd==-1`.
- *SUGGESTION (rc/out parity vs safe_exec_capture_env):* `safe_exec_capture_cwd_env_fd_timeout` returns the same child-exit-code contract and merges stdout+stderr into the capture (verified in #3A), so clone's error surfacing is unchanged.
- *NIT (300s timeout is a behavior change):* intentional and documented — the prior clone exec was unbounded; a hung clone is worse than a bounded one.
- *SUGGESTION (leak test only asserts negatives):* the *positive* fd-wiring proof (a child reads the token via `/proc/self/fd/<target>`) lives in `test_git_cred_inject`'s `safe_exec` fd test; this leak test deliberately covers the assembly-layer "canary in no env string" invariant.
- *NITs (close() return unchecked; combined assert):* a `close()` failure on a memfd we own is non-actionable; left as-is.

No code changes required. CI green.

PR #903, opened 2026-06-29
feat(fold): P5 sealed episodes + task rail — portable plan/episode FSMs

JBailes:
## Summary

Final implementation slice — two pure, tested building blocks completing the fold surface:

- **`src/task_rail.{c,h}` (§8)** — portable plan FSM: objective + locked step list (PENDING/RESERVED/DONE + evidence), `start`/`reserve`/`ack`/`next`/`done_count`, deterministic JSON `serialize`/`restore` so the plan survives folds, epochs, and session boundaries (persists to the existing DB1 `checkpoints.snapshot`).
- **`src/episode_seal.{c,h}` (§5)** — sealed work episode: conclusion + file inventory + a file-touch auto-recall predicate, JSON `serialize`/`parse`. Stored as a distinct `unit_type` (`EPISODE_SEAL_UNIT_TYPE`) on the existing `memory_units` row — **no schema column, no migration** (per the plan F.3 ruling), and does not overload `KIND_EPISODE`.

Both pure (cJSON only), deterministic, fully unit-tested (round-trip, progress, touch predicate, bad-args).

## Scope
This lands the tested FSM cores. The **storage/live binding** — DB1 checkpoint persistence for the rail; DB2 `unit_type` store + cross-session file-touch recall for seals — is the documented close-out follow-up (it needs DB2 helpers + cross-session retrieval paths). Consistent with the staged, default-off rollout of the whole feature.

## Local gates
`make lint` · server+kb build · `make -j unit-tests` (full suite; task_rail 3 + episode_seal 3 cases) — all green. No config/schema change.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -290,7 +290,7 @@ MCP_GIT_OBJS = $(MCP_GIT_SRCS:%.c=$(OBJDIR)/%.o)
 # --- Layer 0: Foundation (db, config, util, text, render, log, platform, cJSON) ---
 CORE_SRCS = db1/db.c db1/db_schema.c config.c config_sections.c config_database.c config_learning.c config_memory.c config_charter.c config_trigger.c config_kb_maintenance.c config_kb_curator.c config_server_api.c config_skills.c config_save.c config_mode.c config_fields.c toolset.c util.c util_url.c forge_credentials.c forge_app_token.c workspace_mirror.c report_enrichment.c delivery_target.c text.c render.c markdown.c json_fluent.c code_audit_graph.c log.c shutdown_forensics.c dstr.c yaml.c diff.c sse_parser.c \
             aimee_home.c shared/kb_paths.c provider_client.c kb_curator_provider.c \
-            compact.c coord_closet.c fold_budget.c context_fold.c fold_register.c fold_recall.c platform_random.c slop_detect.c proxy_bootstrap.c reasoning_cap.c \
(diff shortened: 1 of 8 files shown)


JBailes:

6 panelists (1 dropped). The modules were happy-path-correct; the findings were error-path robustness gaps (mostly OOM / malformed-input), all addressed:

- **Serialize is now all-or-nothing** (both modules): every `cJSON_Create*`/`Add*` return is checked; any failure deletes root and returns NULL, so a successful serialize is always complete (round-trip idempotent).
- **Restore/parse validate-into-temp, swap-on-success** (both modules): root must be an object, `steps`/`files` must be arrays; the result is built into a temporary and only swapped into the caller's struct on **complete** success — so malformed JSON or OOM **never destroys the caller's existing state**, and OOM mid-restore returns −1 (not a silently-truncated structure).
- **`rail_push`**: allocations into temporaries, slot published only after both succeed (no leak / no NULL-title slot on partial OOM).
- **`task_rail_start`**: frees the rail on any push failure so −1 means empty/initial.
- **`task_rail_ack`**: duplicates replacement evidence first; on OOM leaves the step unchanged and returns −1 (no data loss). PENDING/RESERVED→DONE documented.
- **`episode_seal_add_file`** now returns `int` so `parse` can propagate OOM.

Documented: `task_rail_next` returns the first non-DONE step (RESERVED counts as unfinished); out-of-range step state normalizes to PENDING; restore/parse leave prior state intact on failure. The negative-enum-cast claim was a non-issue (only `==1`/`==2` are assigned). All tests + `make lint`/build/full-unit-tests green.

PR #908, opened 2026-06-30
docs: webchat git credential security model

JBailes:
Documents the complete webchat git credential handling — at rest (per-principal vault), in transit to git (memfd fd-askpass for clone/fetch/pull/push; in-process GitHub REST for PR creation; in-memory ssh-agent for SSH — token out of `/proc/<pid>/environ` on every API git path), and cross-tenant isolation — plus the editor's one remaining bounded exposure (`GH_TOKEN` in the code-server env) and the **empirically-established** reasons it persists: node-pty (the integrated terminal) does not forward an inherited fd while `child_process.spawn` (the SCM Git view) does (measured on code-server 4.126/Node24), so the editor needs a **socket-fed askpass**, not the fd-askpass; and ext-host containment as proposed is contradictory (SCM *is* the extension host). Scopes the remaining editor hardening (socket askpass, bwrap jail) as future work. Linked from the README docs index. Roundtable-reviewed for accuracy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ Focused references:
 | [Workflows](docs/WORKFLOWS.md) | The composable dev-lifecycle workflow engine, block catalog, and authoring |
 | [Workspace Management](docs/WORKSPACES.md) | Multi-repo workspaces and session isolation |
 | [Security Model](docs/SECURITY.md) | Threat model, trust boundaries, capability system |
+| [Webchat git security](docs/WEBCHAT_GIT_SECURITY.md) | How webchat handles a webuser's git forge token — at rest, in transit to git, and in the in-browser editor; where exposure is and isn't closed |
 | [Benchmarks](docs/BENCHMARKS.md) | Latency measurements and performance budget |
 | [Compatibility](docs/COMPATIBILITY.md) | Supported OS, shell, and provider matrix |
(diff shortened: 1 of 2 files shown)


JBailes:
**Roundtable accuracy review** (6/6) — doc judged "broadly aligned with the verified facts." Precision fixes applied:
- **Goal #3 scoped** to the API git paths; the editor is now explicitly stated as out-of-scope for that invariant today (resolves the absolute-vs-editor-admission inconsistency).
- Editor exposure broadened: "**any same-UID code-server descendant** — extension host + children, terminal, any spawned shell/git" (was understated as "an extension").
- `memfd` → **`MFD_CLOEXEC`** (the auditable flag); node-pty parenthetical rephrased to the precise mechanism ("the dup2'd fd is not open in the pty child — node-pty does not inherit it").
- Open-VSX/no-preinstalled-extensions reframed as a **deployment policy (not code-enforced)**; server-secret env names marked illustrative (`e.g.`).

The controls the panel couldn't confirm from the prompt facts (loopback bind, requireAuth proxy, `credential.helper` disable, `RLIMIT_CORE=0`, the `git-cred-centralized` lint gate, the ssh-agent `DUMPABLE=0`/buffer-zero, the `AIMEE_WEBCHAT_GIT=0` → 503) were all verified in-session against the code and are accurate as written.

PR #931, opened 2026-07-01
ci(images): build aimee-delegate amd64-only with /mnt disk headroom

JBailes:
Prereq for promoting the delegate work to main. The delegate matrix entries (from #928) would break the release build: the 22GB `delegate-mid` overruns `/` (the `/mnt` data-root move was scoped to `aimee-llm` legs only), and the multi-arch build would bake the x64 Vulkan binary into a non-runnable arm64 image + double the 22GB cost.

- Extend both disk steps (`/mnt` move + free-space) to `aimee-delegate*`.
- Skip delegate `build`/`export`/`upload` on arm64 → the merge job emits a single-arch **amd64** manifest (it already tolerates `count>=1`).

Not exercised by PR CI (`publish-images.yml` runs on tags only); validated by YAML parse + review. Real test is the release build after promotion.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -127,7 +127,7 @@ jobs:
         # The combined (aimee-server-kb) image now bundles the aimee-llm CPU
         # runtime + ~5.5 GB of baked GGUFs (COPYd from aimee-llm-cpu), so it needs
         # the same disk headroom as the llm legs.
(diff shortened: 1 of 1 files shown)


JBailes:
Superseded: retiring the standalone aimee-delegate split in favor of unified aimee-kb-{cpu,gpu-small,gpu-mid} tiers (gpu-mid = Qwen 3.6). The /mnt + arm64 disk handling carries over to aimee-kb-gpu-mid in the new work.

PR #937, opened 2026-07-01
feat(xrepo): implement 'aimee index deps --reverse' (direction=in/both)

JBailes:
## Summary
Implements `aimee index deps --reverse` (direction=in/both), closing the §B `--reverse` gap in the **cross-repo-dependency-graph** proposal. The resolver was OUT-only — `direction=in`/`both` returned empty (kb 501). Reverse traversal now lists the repos that depend **on** a project.

## Approach
Reverse edges are computed by **reusing the OUT engine per candidate caller**, so a reverse edge is byte-identical to the forward edge the OUT query emits (symmetric consistency), inheriting every precision/recall/suppression rule:
- Candidate callers = union of `cross_repo_route` + `cross_repo_build_dep` rows whose `definer_project` is the target (a bounded superset of the emitters).
- For each caller, run `crd_compute_out`, keep edges whose `definer == target`.
- `include_review=0` on the per-caller fan-out — a reverse query is a read and must not enqueue AMBIGUOUS candidates for other repos.
- `BOTH` = forward OUT ∪ reverse IN. Candidate cap honored with early-out; `truncated` propagated.

## Changes
- `db2/cross_repo_deps.c`: extract `crd_compute_out` core; add `crd_compute_in` + `agg_push`; dispatch on direction.
- `kb/http/kb_http_code.c`: map `direction` out|in|both (400 on unknown) instead of 501.
- `server/server_state_index.c` + `cli_v1_routes.inc`: accept in/both; add `--reverse` flag (direction=in).
- openapi (v1 + server-v1) + docs-gen refreshed.
- `tests/test_cross_repo_deps_orch.c`: reverse + both traversal coverage (all unit tests green).

## Testing
`unit-test-cross-repo-deps-orch` passes incl. new `test_reverse_direction`; kb + server binaries build clean.

## Review
Memory-safety reviewed: edge structs are POD → value-copy in `agg_push` is safe and freeing the per-caller source array cannot invalidate copies; the dispatcher frees `agg.e` on every error path. Roundtable/delegate code-review requested with the diff embedded (agents run sandboxed and cannot see the worktree directly).

### DIFF
--- a/api/openapi-server-v1.yaml
+++ b/api/openapi-server-v1.yaml
@@ -1247,8 +1247,8 @@ paths:
       description: >-
         Proxies the kb cross-repo dependency query. Returns confidence-tiered
         edges with evidence + a version stamp, or the AMBIGUOUS review queue when
-        status=ambiguous. The canonical query is OUT-only today (S4a); direction
-        in/both is rejected with 400 until reverse traversal lands.
+        status=ambiguous. direction: out = deps OF the project (default), in =
+        repos that depend ON the project (--reverse), both = union.
       requestBody:
         required: true
         content:
@@ -1258,7 +1258,7 @@ paths:
               required: [project]
               properties:
                 project: { type: string }
-                direction: { type: string, enum: [out], default: out }
+                direction: { type: string, enum: [out, in, both], default: out }
                 min_tier: { type: string, enum: [high, medium, tentative] }
                 status: { type: string, enum: [ambiguous] }
       responses:
--- a/api/openapi-v1.yaml
+++ b/api/openapi-v1.yaml
@@ -1650,8 +1650,9 @@ paths:
           in: query
           required: false
           description: >-
-            Dependency direction. Only "out" is implemented today; "in"/"both"
-            return 501. Ignored when status=ambiguous.
+            Dependency direction: "out" = deps OF the project (default), "in" =
+            repos that depend ON the project (reverse), "both" = union. Ignored
+            when status=ambiguous.
           schema:
             type: string
             enum: [out, in, both]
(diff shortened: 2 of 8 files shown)


JBailes:
### Roundtable code-review (glm-5.2 + minimax, diff embedded)

**Applied:**
- **[glm #2, medium]** `crd_compute_in` silently `continue`'d on a per-caller `crd_compute_out` failure → a transient DB/OOM error could make a reverse query under-report as a complete "no dependents". Now flags `*trunc` (§4.2 partial-result contract). Fixed in 8a04bbd.
- **[glm #5 / minimax #6, low]** Added NULL guard on `agg`/`trunc` in `crd_compute_in` (defense-in-depth). Fixed in 8a04bbd.
- **[glm #4, low perf]** Outer-loop early-out once the candidate cap is hit — already in the initial commit (f2d26d6).

**Verified false / non-issues:**
- **[minimax #1, high — shallow-copy UAF]** FALSE POSITIVE. `xrepo_dep_edge_t` is pure POD (fixed char arrays + ints, zero pointer fields), so `agg_push`'s value-copy is safe and `free(e)` frees only the array container. glm correctly did not flag this as a definite bug.
- **[both, stack size]** `callers[CRD_MAX_REPOS][128]` with `CRD_MAX_REPOS=256` = 32 KB frame — well within thread stacks.
- **[minimax #2 / glm unverified]** `include_review=0` is honored by `crd_compute_out` (verified: `ctx.include_review = opts->dry_run?0:opts->include_review`).
- **[glm #3, cap==0]** `cfg.kb_curator_cross_repo_max_candidates` defaults > 0 (crd_compute_out would emit nothing at cap≤0), so the `cap>0 &&` guard is never reached with cap≤0 in practice.

PR #938, opened 2026-07-01
feat(xrepo): implement 'aimee index deps --dry-run' (acceptance #4)

JBailes:
## Summary
Implements `aimee index deps --dry-run` (acceptance #4) — offline candidate inspection for the cross-repo dependency resolver. Second closeout slice of the **cross-repo-dependency-graph** proposal (follows #937 `--reverse`).

## Behaviour
`--dry-run` emits **every confidence band down to LOW** (not just `>= min_tier`) and surfaces the **AMBIGUOUS candidates inline**, writing **nothing** (no review-queue rows). So LOW/AMBIGUOUS appear *only* under `--dry-run`, never in default output, and nothing is materialized. This makes `--dry-run` meaningfully distinct from `--tier tentative`: the query engine is already non-materializing, so dry-run's value is the full candidate set (incl. the AMBIGUOUS pipeline held back) for offline inspection.

## Approach
- `xrepo_deps_opts_t.dry_run` + new `xrepo_amb_cand_t`. `canonical_index_cross_repo_deps_ex()` returns the AMBIGUOUS candidates captured **in-memory** (an `amb_acc` sink appended in `crd_flush` instead of the review-queue write). The existing `canonical_index_cross_repo_deps()` forwards NULL amb params (API unchanged).
- `dry_run` forces `min_tier=LOW` and `include_review=0` (no writes).
- kb http parses `dry_run`, serializes `dry_run:true` + inline `ambiguous[]`. kb client + server op + CLI `--dry-run` flag + printer (deps down to LOW plus an ambiguous section).
- kb-http-routes test stub for the new `_ex` symbol (link-time gc-sections requirement).
- openapi (v1 + server-v1) + docs-gen refreshed.

## Testing
`test_dry_run_candidates` (ambiguous captured in-memory, never emitted as an edge; plain API ignores it) + full `unit-test-cross-repo-deps-orch` green; kb + server binaries build clean.

### DIFF
--- a/api/openapi-server-v1.yaml
+++ b/api/openapi-server-v1.yaml
@@ -1261,6 +1261,11 @@ paths:
                 direction: { type: string, enum: [out, in, both], default: out }
                 min_tier: { type: string, enum: [high, medium, tentative] }
                 status: { type: string, enum: [ambiguous] }
+                dry_run:
+                  type: boolean
+                  description: >-
+                    Offline candidate inspection — emit every band down to LOW plus
+                    inline AMBIGUOUS candidates, writing nothing (acceptance #4).
       responses:
         "200":
           description: Cross-repo dependency edges or ambiguous review queue
--- a/api/openapi-v1.yaml
+++ b/api/openapi-v1.yaml
@@ -1674,6 +1674,17 @@ paths:
           schema:
             type: string
             enum: [ambiguous]
+        - name: dry_run
+          in: query
+          required: false
+          description: >-
+            Offline candidate inspection: emit every confidence band (down to LOW,
+            overriding min_tier) plus the AMBIGUOUS candidates inline, and write
+            nothing (no review-queue rows). The response carries dry_run:true and an
+            additional `ambiguous` array alongside `deps`.
+          schema:
+            type: string
+            enum: ["1", "true"]
       responses:
         "200":
           description: Cross-repo dependency edges (or ambiguous review queue)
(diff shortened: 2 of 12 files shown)


JBailes:
### Roundtable code-review (minimax, diff embedded) — triage

All findings verified against the actual code; **no code changes required** (the reviewer ran sandboxed and could not see the kb handler / header, which resolve every open item):

- **[#1, high — amb array leak/double-free]** VERIFIED SAFE. The kb handler frees `amb` exactly once: the `!resp` early-return does `free(amb); free(edges)`, and the normal path `free(amb)` after serializing (no early returns in between; `crd_emit` runs after). No leak, no double-free.
- **[#4, medium — evidence truncation]** NON-ISSUE. `xrepo_amb_cand_t.evidence` is `char[512]` and the source `ev` in `crd_flush` is also `char[512]` — sizes match, the copy cannot truncate.
- **[#3, medium — --review + --dry-run]** Reviewer had precedence backwards: `status=ambiguous` is checked first and returns the review view, so `--review` wins and `--dry-run` is ignored (clean precedence, no data loss).
- **[#6, low — min_tier=LOW==0 aliasing]** NON-ISSUE. `XREPO_TIER_NONE=0`, `XREPO_TIER_LOW=1`; the dry-run path sets the local `min_tier` var directly to LOW (not via the `?:` default), and `crd_flush` still explicitly excludes NONE/AMBIGUOUS/UNIMPLEMENTED. Emits exactly LOW/MED/HIGH.
- **[#2, "high" — reverse+dry-run captures no ambiguous]** BY DESIGN and documented (header: "under direction=in it is not populated"). Per-caller OUT-ambiguity is not meaningfully "ambiguous dependents of the target"; reverse dry-run returns confidence-banded reverse edges without an ambiguous section.
- **[#5, low — empty-deps printer]** Intended: dry-run prints "No cross-repo dependencies found" then still shows the ambiguous section.

Verified-clean by the reviewer too: dry-run → no review-queue write (`include_review=0`, gate `(include_review || dry_run)` true → amb sink, DB `else` skipped); `amb_push` OOM leaves `a->c` intact, no double-free.

PR #941, opened 2026-07-02
feat(workflow): primary-as-manager S2 sub-slice 3 — advance_request tool + interactive driver

JBailes:
## Primary-as-manager S2 sub-slice 3 — `advance_request` tool + interactive driver

Part of the [primary-as-manager-enforced-workflows](docs/proposals/pending/primary-as-manager-enforced-workflows.md) proposal (§4/§6). Sub-slices 1–2 (`wfe_enforce`, `wfe_externalization`, `wfe_binding`, the staged dial) already landed via #936; this is step 3 of the S2 build order.

A bound interactive session does **not** let the engine auto-run its work-item (that's the autonomous scheduler's job). Instead the primary, having produced the current block's artifact, calls **one explicit `advance_request` tool** to move the work-item forward exactly one block. The advance is a **server-authoritative** tool-executor action — never inferred from prose ("I've completed X") — so a free-text claim can never cross a gate (consult Q1 / RT1).

### What's here
- **`wfe_advance.{h,c}` — pure core**: arg parse + validate (id-charset, fail-closed); the CAS/replay/unbound/terminal decision (REPLAY ordered before STALE/TERMINAL so a retried turn is an idempotent no-op, never a double-advance or error); and the tool JSON-Schema + description as the single source of truth for the sub-slice-4 injector.
- **`wfe_advance_exec.{h,c}` — the driver**: resolves the session→work-item binding from **authoritative DB state**, reads the actual stage/state, applies the pure decision, and **only on OK** calls `wfe_engine_advance` under the engine's own invariants (it never writes `gate.deliver`/run-state directly). Every decision is audited to `lifecycle_event`; the replay nonce is read back from there (no new table), scoped to applied (`cas=ok`) rows.
- **`server_mcp`**: `advance_request` is fulfilled as a **server-side builtin** handler so a call is authoritative when it arrives — but deliberately **not advertised** in the served tool list, so it is dormant until sub-slice 4 injects it per bound block.

### Safety posture
- **Default-OFF**: inert while `AIMEE_WORKFLOW_ENFORCE_STAGE` is unset.
- Fail-closed on engine/DB faults (`status:error`, no advance); confused-deputy guard (client-supplied work-item only matched against the bound one).

### Verification
- `test_wfe_advance` (pure, cmake + make) and `test_wfe_advance_exec` (real DB1 + engine, stub executors) — both green.
- Production `aimee-server` links clean; format / line-check / module-boundary / schema-sync / dispatch-caps gates pass.
- **Roundtable-reviewed** (commit `783f637`): scoped the replay read-back to applied rows only (real correctness fix + regression test) and moved the audit-detail build to cJSON (defense-in-depth). Other panel items verified not-real (see commit message).

### Not in this slice
Session→work-item binding *creation* on session start, and the per-block tool **injection/strip** (that's sub-slice 4, which surfaces `advance_request` to the primary and enforces the read-only lanes). Live `.253→.254` verification also pending; the wiring is default-off until then.


JBailes:
Approval roundtable (degraded 2/7, non-deterministic panel) returned 3 items, triaged against the tree:

- **[blocking ×2] TOCTOU: advance vs. audit / nonce write** — real only under *true concurrency* on one work-item. `wfe_engine_advance` reads `current_stage` before its own txn (the same property the autonomous single-runner scheduler already relies on). Prevented here by the single-writer session↔work-item binding + the per-turn snapshot; **sequential retries** are handled correctly (the applied nonce is audited before the retry arrives → `replay`). The structural per-work-item lock is the Q4 single-writer **lease, owned by sub-slice 6** by design. Documented the contract inline (`d6df0b8`).
- **[suggestion] orphan audit row on vanished work-item** — real & cheap; **fixed** in `d6df0b8` (skip the refusal audit unless the work-item exists).

Everything is default-OFF and not yet wired to live ingress (sub-slice 4). Unit tests green, production server links, lint gates pass.

PR #942, opened 2026-07-02
feat(delegate): CPU-only Qwen3.6-27B llama.cpp coding delegate + SmoothNAS plugin

JBailes:
## What

A dedicated, single-model, OpenAI-compatible **coding delegate**: one CPU llama.cpp `llama-server` serving a **baked Qwen3.6-27B UD-Q6_K_XL** GGUF, defaulting to **4 slots × 128K = 512K aggregate** context with `q8_0` KV + flash-attention. Ships as a dedicated image + a **CPU-only SmoothNAS plugin**, to be registered against aimee-server as a local delegate.

## Why it works on CPU

Qwen3.6-27B is **dense** but uses a **Gated-DeltaNet hybrid** attention stack (64 layers = 16 blocks × [3 Gated-DeltaNet + 1 Gated-Attention]); only the **16 attention layers keep a KV cache**, so a 512K aggregate context is affordable in system RAM. Native ctx is **262K**, so each 128K slot sits below native → **no RoPE/YaRN scaling**. Runs on llama.cpp **`b9775`** — the same release already serving the MoE sibling (Qwen3.6-35B-A3B) live — using the CPU `ubuntu-x64` build with `-ngl 0`.

## Files

- `Dockerfile.aimee-qwen-delegate` — 2-stage; GGUF pinned by HF revision OID + sha256 (build fails on mismatch); CPU llama.cpp; runtime profile baked as ENV.
- `scripts/aimee-qwen-delegate-{entrypoint,health}.sh` — env-driven launch (`--ctx-size 524288 --parallel 4 --threads 16 -ngl 0 -fa on --cache-type-k/v q8_0 --jinja`), readiness wait (decoupled from the healthcheck start-period), bounded SIGTERM, degraded sentinel.
- `deploy/smoothnas/aimee-qwen-delegate.plugin.yaml` — CPU-only (no `gpu-amd` profile), `hostExpose` :8744, `:latest` default (pin `:testing` via the tierd image override at deploy time).
- `.github/workflows/publish-images.yml` — matrix + merge entry, `/mnt` data-root move + free-disk guards widened, amd64-only build/export/upload guards, layer-cache exclusion.
- `.github/workflows/publish-llm-testing.yml` — dispatch-only `build-qwen-delegate` job (mirrors the gpu-mid pattern: `/mnt` move + df pre-flight).
- `docs/AIMEE_QWEN_DELEGATE.md` — model/arch, 4×128K rationale, tuning knobs, deploy/register/verify runbook.

## Model pins

`unsloth/Qwen3.6-27B-GGUF` @ `82d411acf4a06cfb8d9b073a5211bf410bfc29bf`, `Qwen3.6-27B-UD-Q6_K_XL.gguf`, sha256 `8746881d40f280b1b6b858c656a347c754ed3d9cc8d2e1ad46b3635b87f611f8` (~23.9 GiB).

## Review

Code was run through the aimee roundtable before opening. Applied: manifest defaults to `:latest` (not `:testing`); readiness deadline decoupled from the healthcheck start-period; bounded SIGTERM wait; `AIMEE_DELEGATE_JINJA` declared. Verified out-of-band: the `llama-b9775-bin-ubuntu-x64.tar.gz` asset exists and extracts `llama-server` to `/opt/llama/llama-server` after `--strip-components=1`.

**Deliberately left consistent with the existing aimee llama.cpp images:** the `b9775` release tarball is pinned by (immutable) tag, not sha256 — matching `Dockerfile.aimee-llm`; a tarball-checksum hardening pass across all llama.cpp images could be a follow-up. Auth-off + `hostExpose` (internal bridge only) matches `aimee-llm.plugin.yaml`.

## Register

```
aimee agent local qwen3.6-27b http://10.100.0.1:8744/v1 \
    --model qwen3.6-27b --provider openai --slots 4 --ctx 131072 \
    --roles code,reason,execute --cost-tier 0
```

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -9,6 +9,7 @@
 #   aimee-kb-cpu        (Dockerfile.aimee-llm)   ghcr.io/<owner>/aimee-kb-cpu
 #   aimee-kb-gpu-small  (Dockerfile.aimee-llm)   ghcr.io/<owner>/aimee-kb-gpu-small
 #   aimee-kb-gpu-mid    (Dockerfile.aimee-llm)   ghcr.io/<owner>/aimee-kb-gpu-mid
+#   aimee-qwen-delegate (Dockerfile.aimee-qwen-delegate) ghcr.io/<owner>/aimee-qwen-delegate
 #   aimee-css-render (deploy/css-render/Dockerfile) ghcr.io/<owner>/aimee-css-render
 #
 # Multi-arch (linux/amd64 + linux/arm64) via the canonical build-by-digest then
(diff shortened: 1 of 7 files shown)


JBailes:
Closing: validated live on .254 but CPU throughput for Qwen3.6-27B dense Q6 is ~1-2 tok/s, too slow to be usable as a delegate. Removing the scaffolding rather than keeping dead code.

PR #944, opened 2026-07-02
feat(workflow): primary-as-manager S2 sub-slice 4 — per-block resolver + dispatch-time externalization guard

JBailes:
## Primary-as-manager S2 sub-slice 4 — per-block resolver + dispatch-time externalization guard

Step 4 of the S2 build order (proposal §4/§6, consult Q2/Q3). Follows sub-slice 3 (#941). Reuses the sub-slice-1 policy cores (`wfe_surface_allows`, `wfe_block_default_surface`, `wfe_externalization_*`) — this slice is the **wiring**.

### What's here
- **`wfe_block_resolve.{h,c}`** — resolves a session's per-block enforcement context from **authoritative DB state** (binding → work-item → pinned workflow → current node): `{bound, surface, delivered, advanceable, stage}`. An unbound session, a vanished work-item, or any load failure → not bound / no restriction (a binding-layer failure is isolated, never fails the turn).
- **Dispatch-time externalization guard (Q3)** — the narrow run-state invariant *behind* tool visibility: a bound enforced run whose `gate.deliver` has not passed may **not** call an externalization primitive (`pr.open`/`merge`, `git_push`, `curl`, `post_comment`, …). Enforced at the MCP tool-**dispatch** seam (`handle_mcp_call`, keyed on the authoritative session id) so it covers the table / workflow-tool / `git_` paths alike — *"tool visibility alone is not a security boundary."*
  - Pure decision `wfe_toolcall_decide` (fully unit-tested) + composed `wfe_mcp_toolcall_action`.
  - Dial: advisory = observe, soft = warn+allow, hard = refuse. Every non-allow decision is **audited** to `lifecycle_event` (templated, no tool/arg echo). Delivery **lifts** the guard.

### Safety posture
- **Default-OFF** (`AIMEE_WORKFLOW_ENFORCE_STAGE` unset → ALLOW).
- Fail-open only for *availability* (binding lookup failure → generic, no restriction); the **policy** decision (pre-delivery externalization) fails **closed** under `hard`.

### Verification
- `test_wfe_block_resolve` (pure decision table + real DB1/engine resolve + the composed guard incl. deny/warn/allow, delivered-lifts, audit) — green.
- Production `aimee-server` links; local **CMake thin-client build clean** (`wfe_block_resolve.c` in `SERVER_SRCS` only, mirroring `wfe_advance_exec` so the Windows thin client doesn't compile it); lint gates pass.

### Deferred (documented, not missing)
The ingress tool-**visibility** strip feeding `gateway_policy_strip_tools` + `advance_request` per-block **injection** need a session id at the `/v1/messages` gateway seam (`gw_request_t` carries none) — the same session-identity/binding-creation seam. This slice lands the **structural dispatch-time guard**, which enforces even when visibility is bypassed. Live `.253→.254` verification also pending; default-off until then.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -908,6 +908,7 @@ set(SERVER_SRCS
     ${AIMEE_SRC_DIR}/server/server_session_start_stubs.c
     ${AIMEE_SRC_DIR}/workflow/wfe_engine.c
     ${AIMEE_SRC_DIR}/workflow/wfe_advance_exec.c
+    ${AIMEE_SRC_DIR}/workflow/wfe_block_resolve.c
     ${AIMEE_SRC_DIR}/workflow/wfe_blocks.c
     ${AIMEE_SRC_DIR}/workflow/wfe_approval.c
     ${AIMEE_SRC_DIR}/workflow/wfe_verdict.c
--- a/src/Makefile
+++ b/src/Makefile
@@ -294,7 +294,7 @@ CORE_SRCS = db1/db.c db1/db_schema.c config.c config_sections.c config_database.
             client_integrations.c mcp_tools.c mcp_tool_profile.c mcp_tools_extended.c mcp_skill_tools.c mcp_tools_gateway.c events.c \
             model_registry.c models_dev.c models_dev_cache.c lsp_client.c lsp_manager.c plugin.c \
             plugin_ctx.c plugin_loader.c memory_provider.c plugin_c_hook.c config_plugin.c code_collect.c codex_auth.c \
-            workflow/wfe_iface.c workflow/wfe_def.c workflow/wfe_validate.c workflow/wfe_canonical.c workflow/wfe_custom.c workflow/wfe_manager_artifacts.c workflow/wfe_externalization.c workflow/wfe_enforce.c workflow/wfe_deliver.c workflow/wfe_advance.c workflow/wfe_advance_exec.c workflow/wfe_router.c workflow/wfe_router_catalog.c \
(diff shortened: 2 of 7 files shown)


JBailes:
Roundtable review (non-degraded, 3/7) returned 8 items, triaged against the tree:

- **[blocking] fail-open on a BOUND-but-unresolvable session** — a binding referencing a vanished work-item / unloadable workflow returned `bound=0` → ALLOW, so an externalization primitive slipped through under HARD. **Fixed** (`2afc459`): distinguished from a *truly unbound* session — an instrumentation failure on a known-bound session fails **closed** under hard (Q5 rollout split), audited.
- **[suggestion→fixed] `delivered==accepted` soundness** — any terminal node sets `accepted`, so that proxy is only sound for **enforced** workflows (I2 guarantees their terminal is `gate.deliver`). Guard now applies **only to enforced runs** (`ctx.enforced`); a non-enforced bound run is not gated.
- **[blocking-as-filed] unbound → ALLOW** — correct by design (a truly unbound session is a generic session; spoofing to *unbound* grants nothing). The real gap was the bound-but-unresolvable case above, now closed.
- **[suggestion] execution below the MCP seam** (a block/delegate internally shelling `git push`) — the documented known limitation of name-matching (see `wfe_externalization.h`); delegates run under their own run-state. Out of scope for this seam.
- **[nits] audit templated/injection-safe; `block_is_gate` exhaustive; domain-sized buffers** — verified, no change.

Added regression cases (non-enforced allow, bound-but-unresolvable fail-closed/observe). Default-OFF; tests green; server links.

PR #948, opened 2026-07-02
feat(server): primary-as-manager S2 binding-creation seam (live unblock)

JBailes:
## Primary-as-manager S2 — binding-creation seam (the live unblock)

The dispatch guard + advance_request driver (#941/#944) key on the aimee session id, but nothing created a session→work-item **binding**, so all enforcement was dormant. This threads the primary session id to the gateway router and binds there.

### What's here
- **`wfe_bind_ingress.{h,c}` — `wfe_bind_interactive(sid, message, repo)`**: on a routed **enforced** primary turn, create + bind a work-item (`proposal_path=interactive/<sid>`). Idempotent per session (`binding_get` short-circuits, so create runs at most once), only for enforced workflows, **default-OFF** via the dial. Bind is audited (`lifecycle_event`, cJSON-escaped).
- **Session id at the gateway** via a per-turn thread-local `ingress_preinject_session_id()`, published by the **authoritative in-process primary turn** (`chat_stream_worker_agent`, `primary_session_adapter_turn`) right beside its existing `session_id` override — which synchronously wraps the provider→gateway call. Cleared alongside the override, so a reused worker never leaks a turn's session. **Delegate turns never set it** (they use a different override site), so only the primary binds.
- **`gw_stage_router`** calls `wfe_bind_interactive` when a session id is present.

### Security posture (per roundtable)
- The originally-included **client auth-token channel** (`aimee-sess-<sid>` at `/v1/messages`) was **removed**: it was spoofable — any request carrying that token could bind an arbitrary session under an active dial. Binding now flows **only** from the authoritative in-process turn. The external-CLI proxy path is a documented follow-on that must carry an **authenticated** per-session identity.
- Default-OFF; binding grants only *enforcement* (restriction), is single-writer (`db1_wfe_bind` rejects a second session on the same work-item), and idempotent.

### Verification
- `test_wfe_bind_ingress` (real DB1 + router + engine): only enforced-routed turns bind, dial-off inert, idempotent (one work-item across turns), audited — green.
- Production `aimee-server` links; CMake thin-client links (`wfe_bind_ingress.c` in `SERVER_SRCS` only); lint gates pass.

### Deferred (documented)
The external-CLI `/v1/messages` proxy path (authenticated client-stamped identity) and live `.253→.254` verification. With the dial off this is inert; with it on, an enforced-routed built-in primary turn now binds → the guard + advance driver fire.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -909,6 +909,7 @@ set(SERVER_SRCS
     ${AIMEE_SRC_DIR}/workflow/wfe_engine.c
     ${AIMEE_SRC_DIR}/workflow/wfe_advance_exec.c
     ${AIMEE_SRC_DIR}/workflow/wfe_block_resolve.c
+    ${AIMEE_SRC_DIR}/workflow/wfe_bind_ingress.c
     ${AIMEE_SRC_DIR}/workflow/wfe_blocks.c
     ${AIMEE_SRC_DIR}/workflow/wfe_approval.c
     ${AIMEE_SRC_DIR}/workflow/wfe_verdict.c
--- a/src/Makefile
+++ b/src/Makefile
@@ -294,7 +294,7 @@ CORE_SRCS = db1/db.c db1/db_schema.c config.c config_sections.c config_database.
             client_integrations.c mcp_tools.c mcp_tool_profile.c mcp_tools_extended.c mcp_skill_tools.c mcp_tools_gateway.c events.c \
             model_registry.c models_dev.c models_dev_cache.c lsp_client.c lsp_manager.c plugin.c \
             plugin_ctx.c plugin_loader.c memory_provider.c plugin_c_hook.c config_plugin.c code_collect.c codex_auth.c \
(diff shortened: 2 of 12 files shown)


JBailes:
Roundtable review (non-degraded, 4/7) returned 8 items, triaged against the tree:

- **[blocking] spoofable auth-token channel** — the `/v1/messages` `aimee-sess-<sid>` extraction let any request set the bind thread-local. **Fixed** (`26a01f6`): removed it; binding now flows only from the authoritative in-process primary turn. The external-CLI path becomes a documented follow-on needing an *authenticated* identity.
- **[blocking] thread-local leak on early-return** — **verified not real**: there are no early-return/goto paths between the set and clear in either turn site, and the pair mirrors the existing `session_id` override exactly.
- **[suggestion] audit JSON hand-built** — **fixed**: bind detail now built via cJSON.
- **[suggestion] concurrent-bind race** — interactive turns are serial per session; the loser no-ops on the UNIQUE backstop and the next turn's `binding_get` finds the winner. No change.
- **[nits] duplicate clear/check; id length; nested-delegate residual** — verified benign (idempotent bind; delegate override site untouched).

Default-OFF; test green; server + thin-client link.

PR #953, opened 2026-07-02
feat(workflow): primary-as-manager S2 step 5 — templated enforcement surfacing

JBailes:
## Primary-as-manager S2 step 5 — templated enforcement surfacing

`wfe_enforce_user_message` (the sub-slice-1 templated refusal — **gate name + work-item id only, never the attempted tool/args**) had **no callers**. Step 5 wires it into the dispatch guard so a denied externalization returns a consistent, injection-safe user/primary-facing message instead of a hardcoded string.

### What's here
- **`wfe_mcp_toolcall_action`** gains `(msg, msg_n)`: on a non-ALLOW decision it fills a templated refusal via `wfe_enforce_user_message` (naming `gate.deliver` + the resolved work-item id); `""` on ALLOW. Covers both the resolved-DENY and the bound-but-unresolvable fail-closed paths.
- **`handle_mcp_call`** surfaces that message as the refusal content (falls back to the prior constant if empty).

### Verification
- Test asserts the message **names the gate + work-item** and does **not** echo the attempted tool (`pr.open`); ALLOW clears the buffer. Regressions green; production `aimee-server` links.
- **Default-OFF**: the guard returns ALLOW when the dial is off, so no message is produced.

### Note
The `gate.deliver` terminal-executor surface event is a small follow-on (needs an engine-failure-path test); this slice lands the primary-facing **guard** surfacing, which is the path the primary actually hits when it attempts a pre-delivery externalization.

### DIFF
--- a/src/server/server_mcp.c
+++ b/src/server/server_mcp.c
@@ -1666,13 +1666,19 @@ int handle_mcp_call(server_ctx_t *ctx, server_conn_t *conn, cJSON *req)
     * unset -> ALLOW); soft warns + allows; hard refuses. The decision is audited
     * inside the guard. Runs before every dispatch branch so it also covers the
     * workflow-tool and git_ externalization paths. */
-   if (wfe_mcp_toolcall_action(sid, tool) == WFE_TC_DENY)
    {
-      if (owns_jargs)
-         cJSON_Delete(jargs);
-      return server_send_error(
-          conn, "refused: this action externalizes work before the review/delivery gate has passed",
-          NULL);
+      char deny_msg[256] = "";
+      if (wfe_mcp_toolcall_action(sid, tool, deny_msg, sizeof deny_msg) == WFE_TC_DENY)
+      {
+         if (owns_jargs)
+            cJSON_Delete(jargs);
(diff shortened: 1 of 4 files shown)


JBailes:
Roundtable (non-degraded, 5/7) returned 8 items — all suggestions/nits, **no blockers**, verified against the tree; no code changes:

- **[format-string risk]** — not real: `server_send_error` adds the message via `cJSON_AddStringToObject` (escaped string value, not a printf format).
- **[truncation / 256-buffer]** — not real: templated message is ~155 chars max (fixed template + hex work-item id); `snprintf` NUL-terminates.
- **[WARN fills an unused message]** — harmless: the guard is a general API and fills the templated message for both DENY/WARN; `handle_mcp_call` consumes it on DENY. Cheap; forward-compatible with soft-warning surfacing.
- **[nits: `wi` charset, buffer idiom, header wording]** — `wi` is a stored id-charset value, `safe_id`-guarded inside the templater; no injection/overflow.

The message provably names only `gate.deliver` + the work-item id, never the attempted tool/args (test-enforced). Default-off.

PR #957, opened 2026-07-02
feat(workflow): primary-as-manager S2 step 6 (inc. 1) — sliding-lease staleness tracking

JBailes:
## Primary-as-manager S2 step 6 (increment 1) — sliding-lease staleness tracking

First increment of the step-6 lifecycle watchdog: track a bound work-item's **freshness** so one that's bound but idle (no **meaningful advance**) can be detected. Reuses the `workflow_binding.lease_expiry` column that sub-slice 2 already provisioned — **no schema change**.

### What's here
- **`db1_wfe_lease_renew(session, ttl)`**: `lease_expiry := now + ttl` (`ttl==0` clears → never stale; a **negative** ttl forces a past/stale expiry, for reclaim/tests). Plus `db1_wfe_lease_expiry_get` and `db1_wfe_lease_stale_work_items(out, max)` (surfaces lapsed leases — **detection only**).
- **`wfe_lease_ttl_secs()`** (pure): `AIMEE_WORKFLOW_LEASE_TTL_SECS`, default 3600, `0` = disabled, clamped ≤ 7d.
- **Renew on BIND** (`wfe_bind_ingress`) and on each **applied advance** (`wfe_advance_exec`) — **not** on refusals/replays, so trivial turns can't hold the lease (consult #12: slide on *meaningful* advance, not turn traffic).

### Scope / safety
- **Detection layer only.** The sweep surfaces stale work-items; **auto-reclaim** (which would fail *open* by unbinding a session) + **resume-with-auth** + **reject-scope-add** are increment 2.
- **Default-off**: no lease exists unless a session binds, which requires the dial.

### Verification
- Unit tests: lease renew / expiry / stale-query (real DB1) + the TTL parser (pure). Regressions green; production `aimee-server` links; full `make lint` clean (docs/gen regenerated for the new env var).

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -229,7 +229,7 @@ Scalar keys read directly from the config root (not via the CLI allowlist above)
 
 ## Environment variables
 
-The binaries read 138 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
+The binaries read 139 `AIMEE_*` environment variables (scanned from `getenv()` in `src/`, excluding tests). They override config-store values and are mostly for deployment/runtime wiring. Secrets/tokens should be supplied via the environment or the credential vault, never committed.
 
 ### Paths & assets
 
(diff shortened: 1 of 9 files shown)


JBailes:
Roundtable (non-degraded, 5/7) — 4 items, 3 nits + 1 suggestion, no blockers:
- **[suggestion] renew returns 0 on no matching binding** — real contract mismatch; **fixed** (`sqlite3_changes()` → -1) + regression test.
- **[nit] NOT-NULL reliance uncommented** — added a comment (`lease_expiry` is `NOT NULL DEFAULT ''`).
- **[nits] `const char *sql` vs `char sql[]`; negative-ttl SQL comment** — style; matches existing file convention; negative-ttl is already documented in the header + .c comment.

Detection-only (no unbind → no fail-open); default-off; tests green; full `make lint` clean.

PR #966, opened 2026-07-02
feat(workflow): primary-as-manager S2 step 6 (inc. 2) — lease reclaim + resume-after-reclaim

JBailes:
## Primary-as-manager S2 step 6 (increment 2) — lease reclaim + resume-after-reclaim

Acts on the stale leases increment 1 (#957) detects, and lets a session **resume** its work-item afterward. Also fixes a latent rebind bug in the binding seam.

### What's here
- **`db1_wfe_lease_reclaim_stale()`** — unbinds every lease whose expiry has lapsed and records a `lease_reclaimed` lifecycle_event on the work-item. The **work-item is not deleted**, so it frees up for a resume. Bounded (64/sweep, re-run to drain).
- **`db1_work_item_id_by_proposal(repo, proposal)`** — the missing getter that resolves the id behind a `UNIQUE(repo, proposal_path)` collision.
- **`wfe_bind_interactive` resume** — on a create-collision (`interactive/<sid>` already exists after a reclaim), it now **binds to the existing id** (audited `resume`, not `bind`) instead of returning 0. Fixes rebind-after-unbind.
- **Opportunistic reclaim** at the top of `wfe_bind_interactive` — self-healing per enforced turn, no scheduler. Since the lease slides only on a **meaningful advance**, a session squatting a work-item via trivial turns lapses and is reclaimed (anti-squat, consult #12).

### Fail-open safety
Reclaim runs at **turn ingress** (`gw_stage_router` → bind), *before* the model's tool calls, and only touches leases idle past the TTL (default 1h) — an active turn's lease is fresh. So a session is never left unbound mid-turn with tools in flight.

### Verification
- `test_wfe_binding`: reclaim unbinds the stale session + audits it; a never-leased binding is untouched; re-run returns 0.
- `test_wfe_bind_ingress`: resume-after-reclaim — same work-item, no duplicate created, audited as `resume`.
- Full local `make unit-tests` clean; production `aimee-server` links; `make lint` (clang-format-19) + schema-sync + docs-gen clean. **Default-off.**

### Deferred to increment 3
Cross-session **resume-with-auth** (needs the auth principal, consult #11) + explicit **reject-scope-add** surfacing.

### DIFF
--- a/src/db1/wfe_binding.c
+++ b/src/db1/wfe_binding.c
@@ -2,6 +2,7 @@
 #include "wfe_binding.h"
 
 #include "db1_internal.h"
+#include "wfe_store.h" /* db1_lifecycle_event_add — audit a reclaim */
 
 #include <sqlite3.h>
 #include <stdio.h>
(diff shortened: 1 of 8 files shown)


JBailes:
Roundtable (non-degraded, 4/7) — 7 items, 1 blocking, triaged vs the tree:
- **[blocking] proposal-path truncation → cross-session resume** — real defense-in-depth gap (aimee ids are 8 chars so not exploitable today). **Fixed** (`snprintf` truncation guard → refuse) + long-id regression test.
- **[resume assumes collision]** — not real: resume only fires when `db1_work_item_id_by_proposal` actually finds the work-item, so a non-collision create failure refuses (doesn't wrong-resume).
- **[empty repo in audit]** — misread: the 2nd arg is `stage ""`, not repo; `work_item_id` is correct.
- **[no session_id in reclaim audit / hardcoded 64 / ignored reclaim return / ternary null check]** — minor: work-item id keys the event (+ binding history traces it); 64/sweep is bounded & re-run-to-drain; reclaim is best-effort maintenance (matches convention). No change.

Default-off; tests green; server links; clang-format-19 clean.

PR #970, opened 2026-07-02
feat(workflow): primary-as-manager S2 step 6 (inc. 3a) — reject-scope-add surfacing

JBailes:
## Primary-as-manager S2 step 6 (increment 3a) — reject-scope-add surfacing

Completes the **#30** safe-S2 ruling (*reject mid-workflow scope-add*) with observability.

A session already bound to one work-item can't silently expand scope. If a turn routes to a **different enforced workflow** ("also fix X"), the new intent is **held** — the session stays on its current work-item — and a templated `scope_add_held` lifecycle_event is recorded (workflow ids only, no prose). Same-workflow / non-enforced turns are normal work and surface nothing. The binding is unchanged; this is **surfacing on top of the already-enforced rejection** (the idempotent short-circuit).

### Verification
- `test_wfe_bind_ingress`: a session bound to `mc` that routes to `mc2` stays bound + records exactly one `scope_add_held`; a same-workflow turn adds none.
- Default-off; production `aimee-server` links; `make lint` (clang-format-19) + schema-sync + docs-gen clean.

### Deferred to increment 3b
Cross-session **resume-with-auth** (#11) — needs an auth principal on the binding *and* a non-per-`sid` work-item identity (today `proposal_path=interactive/<sid>` keys resume to the same session). That's a plumbing + identity-design fork, flagged separately.

### DIFF
--- a/src/tests/test_wfe_bind_ingress.c
+++ b/src/tests/test_wfe_bind_ingress.c
@@ -75,6 +75,17 @@ static void setup_home(void)
    assert(f);
    fputs(WF_MC, f);
    fclose(f);
+   /* a SECOND enforced workflow so a bound-session turn can route elsewhere
+    * (scope-add). Same shape as mc, distinct name. */
+   snprintf(path, sizeof path, "%s/mc2.yaml", wf);
+   f = fopen(path, "wb");
+   assert(f);
+   {
+      const char *p = strstr(WF_MC, "\n");
+      fputs("name: mc2", f);  /* replace the "name: mc" line */
+      fputs(p ? p : "\n", f); /* + the rest of the workflow verbatim */
+   }
+   fclose(f);
    setenv("AIMEE_HOME", dir, 1);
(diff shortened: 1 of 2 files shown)


JBailes:
Roundtable (non-degraded, 5/7) — 5 items, all suggestions/nits, no blockers:
- **[audit growth]** re-pivoting every turn would emit a `scope_add_held` per turn. **Fixed**: dedupe so it surfaces once per *distinct* held workflow (+ regression test). Better semantics and bounds the log.
- **[swallowed failure paths]** best-effort by design (a catalog/route error must not block the turn); matches convention.
- **[per-turn catalog load]** the advisory router already loads+routes per turn at this seam; acceptable, noted as a future shared-cache optimization.
- **[snprintf bound / #30 ref]** ids are id-charset < 64, buffer 192; nits, no change.

Default-off; tests green; server links; clang-format-19 clean.

PR #975, opened 2026-07-02
feat(server): primary-CLI-ingestor S2 seam (Slice 2) — gate + enforce-before-send

JBailes:
## Primary-CLI-ingestor S2 seam (Slice 2) — gate + enforce-before-send

First code slice of the **primary-CLI-ingestor backend** — the fix for S2 binding being **inert for the Claude primary**. The Claude primary *must* run as an external CLI, so its model call reaches aimee out-of-band (a different HTTP thread than the turn worker) and the gateway router never sees a session id → the primary never binds. The ingestor enforces at the **in-process turn seam** instead.

This slice is the **enforcement seam only** (driving the `agent_shell` stream-json backend + rendering events are later slices), **additive + default-off**:
- `primary_cli_ingestor_enabled()` — opt-in gate `AIMEE_PRIMARY_CLI_INGESTOR ∈ {1,on,true}`; default off, so the primary's existing path is untouched.
- `primary_cli_ingestor_enforce_preturn(sid, msg, repo)` — routes (S1) + binds/guards (S2) the turn **before** it's sent to the CLI, so enforcement is **preventive** for the turn (not detective on its output — panel [13][21]). **Trust boundary ([3]):** a missing/empty session id is a **no-op returning 0** — never a silent "pretend-enforced".

Reuses `wfe_bind_interactive` (dial-gated + enforced-route-only), so converse/research turns stay unbound and `dial=off` is fully inert.

### Verification
- `test_primary_cli_ingestor`: gate default-off + `{1,on,true}`; dial-off inert; no-sid/NULL-sid no-op; enforced turn + sid → bound; converse → unbound.
- Full `make unit-tests`, production `aimee-server` link, `make lint` (clang-format-19, module-boundary, schema-sync, dispatch-caps, docs-gen) all green. Registered in Makefile + CMake + Rules.mk; header in `src/headers/`.

### Provenance
Slice 0 (probe) + Slice 1 (empirical gate on `claude 2.1.198`) confirmed the approach; design roundtabled 4× (strategic decisions, tmux-backend architecture, refined reuse-of-`agent_shell_driver_t`, and this seam). **Next slices:** drive `agent_shell` stream-json for the primary + wire this seam into the turn path; render ingested events; coexistence/cutover.


JBailes:
Roundtable (degraded, 5/7): 4 items, all nits/non-issues, no blockers:
- **build-registration hunks 'missing'** — diff-scope artifact: I scoped the review diff to the 3 source files; `src/Makefile`, `CMakeLists.txt`, `src/tests/Rules.mk` **are** in the commit (CI `build`/`windows-cmake`/`unit-tests` all green).
- **`<string.h>` unused** — incorrect: `primary_cli_ingestor_enabled()` calls `strcmp`.
- **mkdtemp not cleaned up** — intentional: mirrors the sibling `test_wfe_bind_ingress` convention (in-memory DB, per-run tmp home); adding cleanup to just this one would be inconsistent.
- **'DB1+router+engine' comment** — accurate: real in-memory DB1 + real router + engine, stub *executors*.

No code changes. Default-off; tests + full lint green.

PR #978, opened 2026-07-02
feat(server): primary-CLI-ingestor turn orchestrator (Slice 3)

JBailes:
## Primary-CLI-ingestor turn orchestrator (Slice 3)

Second code slice of the primary-CLI-ingestor backend (Slice 2 = the gate + enforce-before-send seam, #975). This adds the **adapter that drives one primary turn** through the `agent_shell` stream-json CLI backend and **ingests its events**, enforcing S1/S2 **before send**.

- `primary_cli_ingestor_turn(sid, msg, repo, driver, resume_id, out, interrupted)`:
  1. calls `enforce_preturn` **first** → S1 route + S2 bind/guard are **preventive** for the turn (panel [13]);
  2. drives the `agent_shell_driver_t` `open → send → recv → close` (the panel's option-b exec backend — already persistent/streaming/cancellable), ingesting the normalized stream into a caller-owned `primary_cli_turn_result_t { text, session, error, bound, tool_calls }`.
  - Backend session id (`SHELL_EVENT_SESSION_ID`) is captured for next-turn resume; native `tool_start` events are observed **for audit only** (detective, not preventive — the shell-tool bypass is a separate track, [21]).
- `primary_cli_turn_result_free()`.

Still **additive + default-off** and **not yet wired into `chat_stream_worker`** (that cutover is the next slice).

### Verification
- `test_primary_cli_ingestor` drives a **stub** `agent_shell` driver: full `open/send/recv/close` lifecycle; the session is S2-**bound at send time** (proving enforce ran **before** send — the preventive-ordering contract); text/session/tool ingestion; a converse turn runs **unmanaged** (`bound=0`, not pretend-enforced); an unknown driver returns a clean error with no leak.
- Full `make unit-tests`, `aimee-server` link, `make lint` (clang-format-19, module-boundary, schema-sync, dispatch-caps, docs-gen) all green.

**Next slice:** wire the seam into the turn path (opt-in cutover behind the flag) + render ingested events to the display.

### DIFF
--- a/src/headers/primary_cli_ingestor.h
+++ b/src/headers/primary_cli_ingestor.h
@@ -31,4 +31,34 @@ int primary_cli_ingestor_enabled(void);
 int primary_cli_ingestor_enforce_preturn(const char *session_id, const char *message,
                                          const char *repo);
 
+/* Result of one ingested primary turn. All heap fields are caller-owned; free via
+ * primary_cli_turn_result_free. */
+typedef struct
+{
+   char *text;      /* accumulated assistant text (owned; NULL if none) */
+   char *session;   /* backend session id to resume the NEXT turn (owned; NULL if none) */
+   char error[256]; /* first error seen (empty if none) */
+   int bound;       /* 1 if this turn is under S2 management (enforced+routed) */
+   int tool_calls;  /* count of native CLI tool_start events observed (audit) */
+} primary_cli_turn_result_t;
+
+void primary_cli_turn_result_free(primary_cli_turn_result_t *r);
+
+/* Drive ONE primary turn through the agent_shell CLI backend (Slice 3), ingesting
+ * its stream events into `out`. Enforcement (S1 route + S2 bind/guard) runs BEFORE
+ * the turn is sent to the CLI, so it is preventive for the turn (out->bound reports
+ * whether the session is now managed). Native CLI tool events are observed for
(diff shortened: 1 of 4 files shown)


JBailes:
Roundtable (non-degraded, 5/7): 11 items, 2 marked blocking — **both verified false positives**, plus real robustness fixes:
- **[blocking] double-free** — refuted: `out->text = strdup(ctx.text.data)` is a *copy* (new pointer); `dstr_free` frees the original `ctx.text.data`. Distinct buffers.
- **[blocking] use-after-close** — refuted: `handle`'s last reference is the single `driver->close()`; nothing uses it afterward, close called exactly once.
- **dstr NUL-term** — confirmed safe: `dstr_append` always writes `data[len]='\0'`; the empty/NULL case is guarded.
- **[fixed] error contract** — `recv` can fail without `SHELL_EVENT_ERROR`; now sets a fallback message on `rc!=0` (+ `failcli` regression test).
- **[fixed] SESSION_ID** — length-capped, strdup-into-temp, free+replace only on success (never drop a valid id on a failed alloc).
- **[fixed] uninit `rc`** — initialized to -1.
- text accumulation is uncapped by design — consistent with the existing delegate path (`shell_collect_append`); the backend is aimee's own subprocess.

Default-off; tests + full lint green.

PR #979, opened 2026-07-02
feat(server): wire S2 enforce-before-send into the tmux CLI TUI primary (S4) + revert claude-p orchestrator

JBailes:
## Wire S2 enforce-before-send into the tmux CLI TUI primary (Slice 4) + revert the `claude -p` orchestrator

**Corrected transport.** The Claude primary runs as a **per-session tmux CLI TUI** (`cli_session`) — each aimee session gets its own persistent interactive Claude pane, 1:1, which is **mandatory for session isolation**. `claude -p` is a dead end (one-shot print mode can't isolate or persist per-session state — empirically confirmed that `claude -p --resume` is *not* stateful across separate processes). Slice 3's `agent_shell`/`claude -p` turn orchestrator was the **wrong transport** and is **reverted** here; `primary_cli_ingestor` is back to the gate + enforce-before-send seam only.

The real fix is transport-simple: the tmux turn is dispatched **synchronously on the worker thread** (`chat_stream_worker_agent → agent_run_with_tools("code") → cli_session`), where the aimee session id **is** in scope. So enforce right there, before the turn is sent to the pane:

```c
if (primary_cli_ingestor_enabled() && aimee_sid && aimee_sid[0])
    primary_cli_ingestor_enforce_preturn(aimee_sid, message, use_cwd);
```

This binds the session (S1 route + S2 bind/guard) **before** the turn is sent — preventive for the tmux primary, whose out-of-band model call the gateway router never sees with a session id. **No transport change, no new driver.**

### Safety
- Double default-off: the ingestor opt-in flag `AIMEE_PRIMARY_CLI_INGESTOR` **and** the `AIMEE_WORKFLOW_ENFORCE_STAGE` dial — the hot path is untouched unless both are set.
- A missing/empty sid is a no-op (never a silent pretend-enforced — the S2 trust boundary).
- `enforce_preturn` binds only an enforced-*routed* turn; converse/research stay unbound.

### Verification
- `test_primary_cli_ingestor` (the seam: gate + enforce-before-send, dial-off inert, no-sid no-op) green.
- Full `make unit-tests` (incl. every `server_compute.o`-linking test — the new `primary_cli_ingestor → wfe_bind_ingress` symbol chain resolves), `aimee-server` link, `make lint` (clang-format-19, module-boundary, schema-sync, docs-gen) all green.

### Scope
Enforcement (binding) is now live for the tmux primary when enabled. The full **ingest-tmux-turn→IR** goal (if pursued) would take the tmux CLI TUI's own structured output as its source — never `claude -p`.

### DIFF
--- a/src/headers/primary_cli_ingestor.h
+++ b/src/headers/primary_cli_ingestor.h
@@ -1,16 +1,18 @@
 /* primary_cli_ingestor.h -- primary-as-manager S2 enforcement seam for the
- * external-CLI primary (Claude Code driven over `-p --output-format stream-json`).
+ * Claude primary, which runs as a per-session tmux CLI TUI (cli_session): each
+ * aimee session gets its OWN persistent interactive Claude pane, 1:1, so sessions
+ * stay isolated and conversation history lives in the live pane. (`claude -p` is a
+ * dead end here -- one-shot print mode cannot isolate or persist per-session
+ * state, empirically confirmed.)
  *
- * Why this exists: the Claude primary MUST run as an external CLI. Its model
- * calls reach aimee out-of-band (a separate HTTP connection/thread from the turn
- * worker), so the in-process session-id publish never reaches gw_stage_router and
- * the S2 binding is INERT for the Claude primary. This adapter enforces at the
- * IN-PROCESS turn seam instead: BEFORE the turn is sent to the CLI it routes +
- * binds the session (S1/S2), so the primary is under management for the turn.
- *
- * Slice 2 scope: the gate + the enforce-before-send seam only. Driving the
- * agent_shell stream-json backend + rendering its events is a later slice. The
- * whole path is additive and DEFAULT-OFF (see primary_cli_ingestor_enabled). */
+ * Why this exists: the tmux Claude pane's model call reaches aimee out-of-band (a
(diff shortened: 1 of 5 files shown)


JBailes:
Roundtable (non-degraded, 5/7): 7 items, all suggestions/nits, **no blockers**:
- **[wiring not unit-tested]** — acknowledged: the hot-path call in `chat_stream_worker_agent` needs `compute_ctx` to exercise directly (the panel's own [10]). The seam it invokes (`enforce_preturn`) is unit-tested; the call is placed literally before `agent_run_with_tools` (review+build verified) + a flag-gated live smoke. No meaningful test without compute-ctx mocking — deferred as a known limitation.
- **[whitespace sid / NULL message,cwd]** — safe by delegation: the `aimee_sid && aimee_sid[0]` guard matches the adjacent existing pattern (server_compute.c:729); `wfe_bind_interactive` already NULL-guards `message` and `repo`.
- **[test with repo=NULL]** — misread: the test asserts the session **is** bound after the enforced turn.
- **[comment verbosity / include indent]** — clang-format-19 authoritative (lint green).

No code change. Reverts the wrong-transport `claude -p` orchestrator; wires enforce-before-send into the correct per-session tmux CLI TUI path. Double default-off; full unit-tests + lint green.

PR #980, opened 2026-07-02
feat(server): boot-time enforcement-posture log for the primary-CLI-ingestor (capstone)

JBailes:
## Boot-time enforcement-posture log for the primary-CLI-ingestor (capstone)

Observability capstone for the primary-CLI-ingestor initiative. The strategic roundtable flagged ([51]) that enforcement can ship **"as enabled" while silently inert**, with no boot-time signal that the dial being on corresponds to the wiring being active — which is the exact failure class this whole initiative fixed (the tmux Claude primary was silently not binding).

At server startup (right after `wfe_autonomy_register`), when `AIMEE_PRIMARY_CLI_INGESTOR` is on:
- **dial off** → `LOG_WARN`: *"flag on but AIMEE_WORKFLOW_ENFORCE_STAGE=off → enforcement INERT: the tmux CLI primary will NOT bind"* — catches the flag-set-but-dial-forgotten trap.
- **dial on** → `LOG_INFO`: *"enforce-before-send ACTIVE for the tmux CLI primary (dial=X)"*.

When the flag is off it's a **no-op** (default, quiet; the hot path is untouched).

This complements the **live-verified** S4 wiring (#979) — S2 now demonstrably binds the tmux CLI primary (validated on .253: a `bind-s2` lifecycle_event on a real enforced turn) — and makes that posture visible **at boot** instead of only via a DB query.

### Verification
- `test_primary_cli_ingestor` exercises all three branches (off / inert-WARN / active-INFO) for no-crash; the active line was observed firing in the test run.
- Full `make unit-tests` (incl. every `server.o`-linking test — `server.c` now calls the new symbol), `aimee-server` link, `make lint` (clang-format-19, module-boundary, docs-gen) all green. (`log.o` added to the test link.)

### DIFF
--- a/src/headers/primary_cli_ingestor.h
+++ b/src/headers/primary_cli_ingestor.h
@@ -33,4 +33,12 @@ int primary_cli_ingestor_enabled(void);
 int primary_cli_ingestor_enforce_preturn(const char *session_id, const char *message,
                                          const char *repo);
 
+/* Log the enforcement posture ONCE at server startup (observability, consult
+ * strategic-roundtable [51]: enforcement can ship "as enabled" while silently
+ * INERT, with no boot-time signal). When the ingestor flag is off this is a no-op
+ * (default, quiet). When on, it logs INFO with the dial stage if enforcement will
+ * actually fire, or WARN if the flag is on but the dial is off -- the exact
+ * silent-inert trap (flag set, dial forgotten -> the tmux primary never binds). */
+void primary_cli_ingestor_log_posture(void);
+
 #endif /* PRIMARY_CLI_INGESTOR_H */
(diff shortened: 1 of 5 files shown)


JBailes:
Roundtable (non-degraded, 5/7): 6 items, all nits/suggestions, **no blockers**:
- **[boot crash on unset var]** — verified NON-ISSUE: `wfe_enforce_stage_parse(NULL)` returns `WFE_ENFORCE_OFF` (`if (!s||!s[0])`); the test's WARN branch already runs with the var **unset** (→ `parse(NULL)`) and passes.
- **[posture parse diverges from the enforce gate]** — both use the **identical** `wfe_enforce_stage_parse(getenv("AIMEE_WORKFLOW_ENFORCE_STAGE"))` (wfe_bind_ingress.c:75); same source of truth. Added a comment noting it.
- **[fixed] WARN wording** — '=off' → 'off or unset' (unset is parsed identically) + trimmed redundancy.
- **[[51] ref / message length]** — the [51] cite is a design trace; message kept explicit for operator clarity.

Observability only; default-off no-op; tests + full lint green.

PR #981, opened 2026-07-02
feat(workflow): runtime bind-health detector (pre-hard-flip hardening, Track 4)

JBailes:
## Runtime bind-health detector (pre-hard-flip hardening, Track 4)

Overnight Track-4 item. The design roundtable ruled ([30]) that a hard flip needs a **runtime** signal that enforcement is actually firing, and ([9]) that *"a detector that only logs gets ignored."*

This catches a **silently-inert enforce path**: the dial is on and enforced-routed turns are arriving, but nothing ever binds — the exact class of regression this whole initiative fixed (pre-S4, the tmux primary bound nothing).

- Two atomic counters in the bind path: **enforced-routes-seen** (once an enforced route is confirmed) and **binds-done**.
- **WARN once** when ≥3 enforced routes have occurred with **0 binds**: *"the tmux primary enforce path appears INERT (S2 not firing)."*
- **No idle false-positive**: only *enforced* routes are counted, so an idle server (no turns) never warns.
- **Not spammy**: CAS-guarded flag fires the WARN at most once.

Complements the boot-time posture log (#980) with **runtime** detection: posture catches the flag-on/dial-off misconfig at boot; this catches the wiring going inert while turns are flowing.

### Verification
- `test_wfe_bind_ingress`: 3 enforced routes / 0 binds → warned; any bind → never warns.
- Full `make unit-tests`, `aimee-server` link, `make lint` green. (`log.o` added to the test link — the detector calls `aimee_log`.)

### DIFF
--- a/src/tests/Rules.mk
+++ b/src/tests/Rules.mk
@@ -1344,7 +1344,7 @@ $(TESTPREFIX)/unit-test-wfe-advance: $(OBJDIR)/tests/test_wfe_advance.o \
 
 # S2 binding seam: auth-token->sid parser + idempotent interactive bind.
 $(TESTPREFIX)/unit-test-wfe-bind-ingress: $(OBJDIR)/tests/test_wfe_bind_ingress.o \
-                                    $(OBJDIR)/workflow/wfe_bind_ingress.o $(OBJDIR)/workflow/wfe_enforce.o \
+                                    $(OBJDIR)/workflow/wfe_bind_ingress.o $(OBJDIR)/log.o $(OBJDIR)/workflow/wfe_enforce.o \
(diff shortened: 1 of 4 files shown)


JBailes:
Roundtable (degraded, 2 items) — both addressed by a redesign:
- **[1] blind-to-regression** (lifetime `binds==0` goes blind after the first bind) and **[0] cross-counter TOCTOU** — fixed by a **single counter: enforced-routes-since-the-last-bind**, reset on every bind (which re-arms the warn). Now catches *never-worked* **and** *regressed-after-working*, with no cross-counter race.
- Test adds the working-path (interleaved binds → no warn) + regression-re-detect cases.

PR #982, opened 2026-07-02
feat(workflow): native-tool externalization classifier (shell-tool-bypass, tracks 2+3 PR-A)

JBailes:
## Native-tool externalization classifier (tracks 2+3, PR-A)

The pure-policy core of the **shell-tool-bypass** fix that `wfe_externalization.h` explicitly calls out as a tracked follow-on: *"externalization performed THROUGH a general shell tool (a Bash tool running `git push` / a remote `curl`) is not caught by name-matching."*

`wfe_native_tool_externalizes(tool_name, command)` classifies a primary agent's **native** tool call as a *known* externalization:
- a named externalization primitive (`wfe_is_externalization_tool`) or web/egress tool (`WebFetch`/…);
- a **shell** tool (`Bash`/`sh`/`run_command`/…) whose command matches a known pattern: git remote write (`push`/`send-email`/`remote add|set-url`), a `gh` **mutation** (pr create|merge|comment|edit / release / issue create|comment|close / api), a package publish (npm/yarn/pnpm/cargo/twine/docker/gem/poetry), a remote transfer (scp/sftp/rsync/ssh), or a fetcher (curl/wget/nc/…) to an explicit **non-loopback** URL.

Word-boundary matching (leading **and** trailing) so `git pushd`, read-only `gh pr view`, and local `curl http://localhost` do **not** false-positive.

### ⚠️ Honest scope (roundtable [1][25][28][35])
This **reduces** the native externalization surface; it does **not** hermetically close it. Obfuscation the string match can't see — subshells (`$()`/backticks/`bash -c`), `base64 -d | sh`, heredocs, shell functions, unenumerated tools — still escapes. **A negative result means "no known pattern", never "safe".** The full seal needs a sandbox (network-off / read-only mounts), tracked separately. The header documents this prominently.

### Verification
- 30+ assertion unit test: externalizing cases caught (push/gh-mutations/publishes/transfers/external-fetch); non-externalizing not false-positived (status/commit/read-only gh/local curl/`git pushd`).
- Pure (no DB); `aimee-server` link + `make lint` (clang-format-19, module-boundary, docs-gen) green. Registered in Makefile + CMake + Rules.mk.

**Next (PR-B):** wire this into the PreToolUse hook (`cmd_hooks pre`) — query the session's S2 binding, and deny (ENFORCE) / log (WARN) an externalizing native tool pre-`gate.deliver`.

PR #843, opened 2026-06-28
docs(proposals): close out central-agent-memory-interception (pending→done)

JBailes:
## What

Closes out the **central-agent-memory-interception** proposal. The feature is built, merged to `testing`, and live-validated on a deployed split server, but the proposal doc was still **State: DRAFT (pre-implementation)** with **no implementation-status section**. This reconciles it and files it to `done/` (with its `.plan.md` companion).

### Shipped (all merged, each roundtable-reviewed before merge)
P1 #794 · P2 #795 · P3 #798 · P4 #802 · v1.1 #804/#806/#808 · v1.2 #809/#811/#813/#815 · e2e fixes #817 (startup segfault) / #819 (split-server wiring) · test-plan fix #829 (per-client scoping). Approved test plan executed: **9/9 unit suites + full split-server e2e (functional/reconcile/fail-open/concurrency/security/Bash/watcher) all PASS.**

### Close-out decision (roundtable-ratified)
A close-out roundtable (6 panelists, 0 failed, not degraded) agreed it may move to `done/` **conditioned on the done record stating the accepted residual risk + GA gates** — the panel's blocking item explicitly offered "a recorded threat model" as the alternative to building the deferred mitigations first. So the done doc now adds a **Close-out scope** section making clear:
- **"Done" ≠ production-ready/GA**, and **not** safe for untrusted/remote/multi-tenant use.
- **GA-blocking gates**, reclassified from generic future-work: untrusted shared-memory prompt-injection provenance (v1 trust model = clients sharing a project are mutually trusted); **RM3/4/8 remote auth/replay before E-PROD**; the `AIMEE_HOOK_CLIENT` cross-client trust boundary (server must authenticate the thin client before multi-principal use); the **Linux-only** backstop platform window (non-Linux = session-start reconcile only); divergence audit counters as the primary drift signal; SHA-256 hash domain (canonicalized body, not raw bytes).

## Verification
- `scripts/check-proposal-reconcile.py` → exit 0
- `scripts/check-proposal-links.py` → exit 0

A roundtable review of this diff will be posted before merge.

PR #984, opened 2026-07-02
feat(server): S2 preventive native-tool gate via the PreToolUse hook (shell-tool bypass, tracks 2+3)

JBailes:
## S2 preventive native-tool gate via the PreToolUse hook (tracks 2+3)

Closes the **shell-tool bypass** that `wfe_externalization.h` flags as a tracked follow-on: an enforced run's pre-delivery externalization guard was *name-only*, so Claude Code's **native** tools (a `Bash` running `git push`, a remote `curl`) escaped it. This wires a **preventive** gate into aimee's **existing deny-capable PreToolUse hook** (`cmd_hooks pre` → the same `emit_pretool_deny_json` channel the central-agent-memory interception uses, #843).

**Flow:** the tmux Claude's PreToolUse hook (`aimee hooks pre`) → resolve the aimee session (from `AIMEE_SESSION_ID`, now stamped into the per-session tmux env) → **authoritatively** query the binding (`db1_wfe_binding_get`) → if bound to an enforced work-item not yet `accepted` (gate.deliver not passed) and the tool is a **known externalization** → **DENY** (hard dial: emit deny + exit) or **WARN** (advisory/soft = warn-soak: log the would-deny to measure false positives before a hard flip).

- **`wfe_native_gate.c`** (pure): `wfe_native_tool_externalizes(tool, command)` + the decision truth table `wfe_native_gate_decision(...)`. Word-boundary = "not a word char" (robust vs `/usr/bin/git push`, `bash -lc 'git push'`, `;`/`}`/tab/quote evasion). Userinfo-bypass fixed (host after `@`); `gh api` dropped (read-by-default FP). *(These were the #982 roundtable findings — that PR is folded in here.)*
- **`cmd_hooks.c`**: `s2_native_gate_pretool` — the wiring. **Fail-closed** on the externalizing surface when the binding is unreadable under a hard global dial.
- **`agent_runtime_tmux.c`**: charset-validated `AIMEE_SESSION_ID` stamp into the tmux CLI env.

### ⚠️ Honest scope (roundtable [1][25][28][35])
**Risk-reduction, not a hermetic seal.** Obfuscation the string match can't see — subshells (`$()`/backticks/`bash -c`), `base64 -d | sh`, heredocs, scheme-less egress (`nc host 443`), `gh api -X POST`, unenumerated tools — still escapes. A negative means "no known pattern", never "safe". The full seal is a sandbox (network-off / read-only mounts), tracked separately. The header documents this prominently.

### Verification
- 40+ classifier/evasion assertions + the decision truth table; full `make unit-tests`, `aimee-server`+`aimee` link, `make lint` green.
- **Next:** a real-Claude E2E denial on an isolated `.253` CT (the roundtable-mandated end-to-end check).


JBailes:
Wiring roundtable (non-degraded, 7/7) — real issues fixed:
- **[16][25][31] security** — raw-string `tool_input` (not a `{command}` object) let Bash externalization slip through. **Fixed:** fall back to inspecting the raw `tool_input` (command text is present; substring match still catches it) → also robust to malformed JSON / OOM ([5][9][26]).
- **[3][24]** — the `AIMEE_SESSION_ID` stamp was unconditional → a delegate got the primary's sid (mis-attribution). **Fixed:** stamp only the primary (`!deleg`).
- **[17][32]** — truncation silently dropped the stamp (fail-open) → now `LOG_WARN`.
- **[1]** — fail-closed + deny/warn audit rows now log the `sid`.
- **[6][14][23]** — documented the fail policy: DB-fault fails **closed** under a hard dial; a cleanly absent binding is allowed ('unknown' ≠ 'none'); a session outside the stamp resolves no binding (inherent env-trust limit).
- **[4]/[20]** — removed unreachable `break`; allow `.` in the sid charset.

Rejected: **[30]** (`sid` *is* in scope — compiles + full suite green), **[0]** (matches the `wfe_block_resolve` precedent), **[8]** (minor TOCTOU, momentarily-correct), **[10]** (`cli_cmd` is aimee-built, not attacker input).

Full unit-tests + server/CLI link + lint green. Real-Claude E2E on an isolated .253 CT next.

JBailes:
**Live E2E validated on .253** — and it caught a real architectural bug: the gate was in the wrong layer. `aimee hooks pre` **forwards to the server's hooks.pre RPC**, so the CLI `cmd_hooks` copy is only the server-unreachable fallback and never ran in the real path. **Fixed:** moved the gate into `handle_hooks_pre` (server.c) via a new module `server/s2_native_gate_hook.c` (deduped the blocked-response builder with the memory-interception path to keep server.c ≤2000).

**E2E results (real server hooks.pre, dial=hard, bound pre-delivery session):**
| command | result |
|---|---|
| `curl` external / `npm publish` / `docker push` | **DENY** (S2 message + audit `DENY … stage=hard`) |
| `git status` / `curl localhost` | allow (no false positive) |
| external `curl` on an **advisory** binding | allow + `WOULD-DENY` audit (warn-soak) |
| external `curl` on an **unbound** session | allow |

The shell-tool bypass is **closed** for the enumerated externalizing commands, with the warn-soak calibration path and no false positives. Full unit-tests + server/CLI link + `make lint` (line-check, module-boundary, docs-gen) green. Torn down.

PR #1107, opened 2026-07-06
feat(webui): Workflow Actions lifecycle controls + webchat proxy fix

JBailes:
## What

Two commits:

1. **`feat(webui)` (6f2b96f0)** — Workflow Actions lifecycle controls (Start/Pause/Stop/Delete), clickable Logs detail modal, and an in-project proposal picker. Adds `db1_work_item_delete`, `wf_api_item_{pause,resume,stop,delete}` + `/v1` routes, `wfe_autonomy` skip-when-`operator_paused`, and the read-only realpath-confined repo tree/file browser.

2. **`fix(webui)` (this branch)** — the lifecycle controls were dead on arrival: the frontend calls `POST /api/workflow/items/<id>/{pause,resume,stop}` and `DELETE /api/workflow/items/<id>`, and the C `/v1` routes exist, but the **webchat Go proxy** (`handleWorkflowItems`) predated those routes — its switch only mapped `/gate`, `/events`, `/proposal`, and bare-GET. Every lifecycle mutation fell through to `default:` → `{"error":"bad path"}` (HTTP 400). Operator-reported symptom: "bad path" on Stop/Delete (and Start).

## Fix

Add the missing switch cases. pause/resume/stop (body-less POST) and bare-id DELETE forward via `webuserPass` under the caller's webuser identity, mirroring the existing read forwarding. **Authorization unchanged** — the `/v1` handlers all run `wf_load_for_mutation` (owner-or-operator, fail-closed on empty submitter), same route cap (`CAP_DASHBOARD_READ`) as `/events`/`/proposal`.

## Tests

New `webchat/workflow_test.go`:
- `TestWorkflowItemsLifecycleRouting` — each of pause/resume/stop/delete reaches the correct `/v1` path with the `X-Aimee-Webuser` header and a body-less (Content-Length 0) forward.
- `TestWorkflowItemsRejectsUnmapped` — wrong method on a lifecycle suffix and a mutating method on the bare id path still return 400 "bad path".

Verified **red-before-green**: reverting the new cases reproduces exactly the reported `{"error":"bad path"}` 400. `go build`, `go vet`, and full `go test ./...` for webchat pass.

## Roundtable

Reviewed via aimee roundtable (7 panelists, 0 failed). No blocking correctness bugs — its two "correctness" flags (DELETE case-ordering, nil-body handling) were **rejected at verification** as contradicted. Surviving suggestions addressed: verified the C handler owner-or-operator enforcement out-of-diff, and strengthened the tests (body-less assertion + method-mismatch table).

## Deploy note

The C server already has the `/v1` routes; this fix is in the **webchat Go binary**, which must be rebuilt/redeployed for the controls to work live.

### DIFF
--- a/api/openapi-server-v1.yaml
+++ b/api/openapi-server-v1.yaml
@@ -459,6 +459,56 @@ paths:
           content:
             application/json:
               schema: { type: object }
+  /workflow/repo/tree:
+    get:
+      summary: List directories + .md files under the server's local project checkout (composer proposal browser), path-confined
+      parameters:
+        - name: path
+          in: query
+          required: false
+          schema: { type: string }
+          description: Relative path under the project root ("" = root); rejects .. / symlink escape
+      responses:
+        "200":
+          description: "{path, entries:[{name,type:dir|file}]}"
+          content:
+            application/json:
+              schema: { type: object }
+        "400":
+          description: Path escapes the project root
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/ErrorResponse' }
+        "404":
+          description: Not a directory / not found
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/ErrorResponse' }
+  /workflow/repo/file:
+    get:
+      summary: Read one .md file under the confined project root (composer proposal browser)
+      parameters:
+        - name: path
+          in: query
+          required: true
+          schema: { type: string }
+          description: Relative .md path under the project root; rejects .. / symlink escape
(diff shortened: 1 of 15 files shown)


JBailes:
Superseded by a clean PR. The Workflow Actions lifecycle feature (commit 6f2b96f0) was already squash-merged to testing under a different SHA, so this branch duplicated it and conflicted. The webchat proxy fix alone is now in a fresh branch off current testing.

PR #1187, opened 2026-07-09
docs(proposals): local-first memory & trust patterns concept survey

JBailes:
## What

A design-only proposal in `docs/proposals/pending/` surveying **seven transferable patterns** — memory synthesis governance, context-assembly trust boundaries, tool-tier gating, self-expansion review, idempotency, skill capture — re-expressed as concepts and mapped onto aimee's existing substrate.

## Framing

Every item is scoped as **reuse/extend what exists**, not rebuild. aimee's memory/curator stack is already more capable than the source; the value is a small set of sharp, self-contained ideas (and two governance patterns) that aimee doesn't yet express and that matter more once it acts autonomously and across trust boundaries.

## Highlights (ranked by value-to-effort)

- **#3 Context-assembly trust boundary** — provenance-tag inbound/tool content inline, owner-gate passive injections. The leak is usually the passive-injection path, not the tool call.
- **#6 Idempotency ledger** — arg-derived request IDs dedupe *sequential* retries of outbound actions; composes with existing job leases.
- **#1 Idle synthesis → quarantine lane** — proactive memory synthesis into an isolated, never-auto-promoted, undoable tier; extends `memory_lifecycle`.
- **#4 Trust-tier gating** — default-deny/fail-closed for non-owner sessions; "structurally absent, not disabled" for the riskiest tools.

## Scope

No code. Recommended first cut: ship #3 + #6 together as a hardening pass, then #1.

_Next: running through `aimee delegate roundtable --mode review` to convergence; revisions will be pushed to this PR._

### DIFF
--- a/docs/proposals/pending/local-first-memory-and-trust-patterns.md
+++ b/docs/proposals/pending/local-first-memory-and-trust-patterns.md
@@ -10,8 +10,9 @@
   Review (out-of-band approval gates, provenance, undo), Execute (idempotent
   side effects), Reason (trust-tier gating).
 - **Reviewed by:** multi-persona roundtable panel (security, architect, QA,
-  contrarian, constructive) — this revision resolves the panel's blocking
-  findings; see "Roundtable resolutions" at the end.
+  contrarian, constructive), two passes. Round 1 raised 8 blocking findings;
+  round 2 (confirmation) returned **zero blocking findings** — only refinement
+  suggestions, now folded in. Converged. See "Roundtable resolutions" at the end.
 
 ---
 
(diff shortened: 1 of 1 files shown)


JBailes:
### Roundtable convergence

Ran the proposal through `aimee delegate roundtable --mode review` to convergence:

- **Round 1** — panel (security · architect · QA · contrarian · constructive) returned **8 blocking findings**: PIN-in-chat secret leak, idle-vs-autonomous-dev trigger mismatch, "prompt labelling ≠ enforcement", unverifiable "already has" claims + a `test_vault_capability` miscitation, compile-check/human-approval conflation, ambiguous 80/20 framing, an unanchored anecdote, and dependency-blind prioritization. Resolved in `a455dd3b`.
- **Round 2 (confirmation)** — **zero blocking findings** (`degraded=0 rejected=0`); only refinement suggestions. Folded in via `cfdf7008`: verbatim-based deterministic undo + snapshot fallback (#1), provisional "Have"-column framing, split of #3 into **3a boundary / 3b mitigation**, and fully-specified audit redaction/retention (#5).

Converged. The proposal now carries a verification-status caveat + `[verify]` traceability table, an explicit threat model, and a per-item Have/missing split — acceptance of any item is gated on turning its traceability row green against the real implementation.

PR #1241, opened 2026-07-11
feat(self-update): thin-client version-drift detection + `aimee self-update`

JBailes:
## Why
The thin client talks **1:1** to one aimee-server, so the server's version is observable on every call and is the exact target to match. When the server moves ahead, the client can drift — which just bit us (a 0.2.176 client's worktree guard behaved differently from a 0.2.182 server). This gives the client a way to *see* the drift and catch up, keeping the two in lockstep.

## What
- **`src/self_update_util.c`** (pure, unit-tested): semver parse/compare (ignores a leading `v` and a git-describe `-N-gSHA` suffix), version-string validation, and the platform release-asset name (`aimee-linux-x86_64`/`arm64`, `aimee-macos-universal`).
- **`src/cmd_self_update.c`**: reads the server version from the existing `GET /v1/version`. `aimee self-update [--check] [--version vX.Y.Z] [--yes]` downloads the matching GitHub release asset over cert-verified HTTPS, **requires the downloaded binary to run and report the exact target version** (catches truncation / wrong-arch / wrong-version), backs up the current binary, then **atomically renames** it into place (running processes keep the old inode). Never downgrades; rejects implausible version strings; Linux swap via `/proc/self/exe`.
- **SessionStart** (thin-client/remote path): emits a one-line *"server is vX, you are vY → run `aimee self-update`"* notice when the server reports a higher semver. Best-effort; never blocks the session.

Conservative first cut: session-start only **notifies**; the swap is **user-initiated**.

## Verification
- `test_self_update` (new): version compare / validation / asset mapping — passes.
- `-Werror` thin-client build clean; unit test green.
- **Live:** `self-update --check` against the configured server; and a full **download → verify → atomic-swap of a real `v0.2.181` release** onto a throwaway binary (v0.2.100 → v0.2.181, backup kept). Guardrails confirmed: downgrade refused, unsafe version rejected, already-up-to-date.

## Follow-ups (called out, not blocking)
1. **Cryptographic verification.** Today's integrity is HTTPS transport trust + the version self-check; the release publishes no per-asset signatures. Publishing a minisign/cosign signature (or verifying the GitHub API asset digest) and checking it before swap would close the supply-chain gap — recommended before enabling any auto-`apply` mode.
2. **Server should report a semver.** A dev/branch server build reports e.g. `testing-<sha>`, which parses to `0.0.0` — so drift is a **safe no-op** (never a false "update") but also won't be *detected*. Release servers report a clean semver and work as intended.
3. **`apply` mode + single install path.** A config-gated auto-apply (this PR is notify-only) and collapsing the client to one canonical install location (this repo currently had two out-of-sync `aimee` binaries) are natural next steps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -789,6 +789,10 @@ set(CLI_SRCS
     ${AIMEE_SRC_DIR}/cli_main.c
     ${AIMEE_SRC_DIR}/cli_agent_setup.c
     ${AIMEE_SRC_DIR}/cli_session_start.c
+    # Thin-client self-update (cli_main dispatch + cli_session_start drift notice
+    # both call these; the CLI does not link aimee-cmd, so they build in here).
+    ${AIMEE_SRC_DIR}/cmd_self_update.c
+    ${AIMEE_SRC_DIR}/self_update_util.c
     # memory_redirect_check (the CLI's remote-only hooks-pre interception) calls
     # hmem_resolve_project/_scope_for_client/_audit/_spill; the standalone `aimee`
     # CLI does not link aimee-core, so their definers must build into the CLI
--- a/src/Makefile
+++ b/src/Makefile
@@ -354,7 +354,7 @@ CMD_SRCS = prompts.c persona.c hardware_probe.c curator_profile.c server/delegat
 CMD_OBJS = $(CMD_SRCS:%.c=$(OBJDIR)/%.o)
 
 # --- Pure client (aimee) ---
-CLI_SRCS = cli_v1_routes.c cli_v1_routes_d.c cli_v1_routes_c.c cli_v1_routes_b.c cli_main.c cli_agent_setup.c cli_session_start.c harness_memory_common.c harness_memory_scope.c harness_memory_audit.c harness_memory_spill.c cli_attention_guard.c cli_code_audit.c cli_css.c cli_tui.c cli_tui_opencode_v2b.c cli_tui_opencode_v2.c cli_tui_misc.c cli_client.c cli_launch.c cli_mcp_serve.c cli_profile.c client_integrations.c server/delegate_plan.c cmd_profile.c cmd_manuscript.c cmd_optimize.c cli_persona.c cli_roles.c manuscript.c http_uds_client.c aimee_client.c cli_remote.c acp_server.c cli_workspace_serve.c cli_agent_keys.c cmd_workflow.c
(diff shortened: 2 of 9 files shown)


JBailes:
Follow-ups from the initial review are now included in this PR (commit `fe0d347b`):

1. **SHA-256 verification** — the download is now checked against the digest GitHub publishes for the release asset (`releases/tags/v<X>` → `assets[].digest`), on top of HTTPS + the version self-check. Confirmed mismatch is fatal; `--require-verify` makes a *missing* digest fatal too. Verified live: `Verified SHA-256 dede88a1…` matched GitHub's published digest on a real `v0.2.181` download. (Key-based signature vs a compromised GitHub remains a further follow-up — the release publishes no per-asset signatures, only the API digest.)
2. **Non-semver server handled honestly** — `testing-<sha>` is deliberately non-semver (`publish-testing.yml`), so instead of forcing the server to change, the client now detects it (`aimee_version_is_semver`) and reports "not comparable, target a release explicitly" rather than the old misleading "client ahead". The drift notice stays silent for non-semver servers.
3. **`apply` mode** — `update_mode: off|notify|apply` in aimee.yaml (default `notify`). In `apply`, SessionStart spawns a detached, `--require-verify` self-update when the server is a strictly-newer semver, rate-limited to once/hour. Non-blocking (atomic rename). Gated conservatively; a safe no-op against the deliberately-non-semver `.254` here.
4. **Sibling-binary warning** — after a manual update, warns about other `aimee` on PATH that now differs (the two-out-of-sync-copies smell), suggesting one canonical install.

A security-focused roundtable review of these changes is running.

JBailes:
Security roundtable (MiniMax-M3, `--via` routed around gpu-mid saturation) returned **CHANGES-REQUESTED** with 2 HIGH + 1 MEDIUM + LOW findings — all addressed in commit `19ee2dab`:

- **#1 (HIGH, TOCTOU)** — backup now taken *before* verification, so SHA-256 + version-self-check and `rename()` are adjacent (minimal substitution window). Trust boundary documented: an attacker who can modify the temp between verify and rename already has install-dir write access and could replace the binary directly — defense-in-depth, not new surface.
- **#2 (HIGH, unbounded home in stamp path)** — snprintf truncation now checked for the rate-limit stamp + child log path.
- **#6 (LOW, child stdio)** — if `/dev/null` can't be opened in the apply fork, stdin/stdout/stderr are closed so a write fails rather than corrupting the hook channel.
- **#3 (MEDIUM, dead code)** — unreachable `tpfx` branch removed.

The review confirmed **no injection or memory-safety sinks** and that the layered integrity gate is sound. Re-verified end-to-end after the fixes (SHA-256 confirmed against GitHub's digest, backup, atomic swap; `-Werror` build + unit tests green).

PR #1244, opened 2026-07-11
fix(roundtable): compose panels from the configured default preset's seat models

JBailes:
## What

Makes `gate.roundtable` **compose its panel from the configured `roundtable.default` preset**: when a review persona has no agent that advertises it, the panelist is dispatched on the **model bound to that persona's seat** in the preset — instead of the gate failing closed to `panel_degraded`.

## Why

`dispatch_review` (live panel) resolves a persona → agent by matching the persona's roles **and** `agent_supports_persona`. On .254 no agent advertises the bare `qa`/`reviewer` (etc.) review personas, so:

```
WARN wfe-panel: no agent for required persona 'qa' -> degrade
```

→ the required lens gets no verdict → `wfe_gate_decide` **degrades** → the run parks `panel_degraded` at the roundtable, **before reviewing anything**. Every autonomous build stalls there (`wi_712049`, `wi_2c954134` observed).

The operator's configured **`default` roundtable preset** already solves this by binding each persona to a concrete model:

| persona | model |
|---|---|
| reviewer-constructive | GLM-5.2 |
| architect | MiniMax-M3 |
| qa | mistral-medium-3-5 |
| security | mimo-v2.5-pro |
| reviewer | codex |

## Change

In `dispatch_review` (`wfe_live_panel.c`), when the role/persona-support search finds no agent, fall back to the active `roundtable.default` preset: find the seat whose `persona` matches and dispatch on that seat's `model`'s agent (enabled, available, not DOWN). So the panel composes from the operator's configured roundtable.

Server-side only; the per-persona dispatch is **integration-exercised** (per this file's own contract — it needs reachable review agents, so it isn't in the unit suite). Full server build + link verified locally.

## Why this matters now

This **unblocks** the plan↔roundtable feedback loop (#1238): that fix is deployed and correct, but its convergence can't be *observed* until the panel stops degrading. With the panel composing, the roundtable actually reviews, returns verdicts, and the feedback loop can drive refinement.

## Validation

Deploy to .254 + re-fire a proposal; confirm `plan_gate` composes a panel (no `panel_degraded`) and the run proceeds through the roundtable. (Validation-pending until deployed.)

### DIFF
--- a/docs/WHATS_NEW.md
+++ b/docs/WHATS_NEW.md
@@ -11,6 +11,14 @@ Thin-client + credential hardening, the client owns the working tree; agent
 credentials live in the server's **sealed vault**; see
 [THIN_CLIENT.md](THIN_CLIENT.md).
 
+- **Roundtable panels use the configured default preset's models**: a review
+  persona that no agent advertises (e.g. `qa`/`reviewer`) no longer degrades the
+  gate. The live panel now falls back to the active `roundtable.default` preset,
+  dispatching each persona on the **model bound to its seat** (e.g.
+  `qa → mistral-medium-3-5`, `security → mimo-v2.5-pro`). So a `gate.roundtable`
+  composes from the operator's configured roundtable instead of failing closed to
+  `panel_degraded` when a bare persona has no assigned agent.
+
 - **Roundtable ↔ planner now refine together**: when a `gate.roundtable` requests
   changes, the panel's blockers are captured and persisted for the work item, and
   the next `author.proposal`/`author.plan` pass folds them into its prompt — so the
--- a/src/server/wfe_live_panel.c
+++ b/src/server/wfe_live_panel.c
@@ -20,9 +20,11 @@
 #include "agent.h"
 #include "agent_config.h"
 #include "agent_exec.h"
+#include "config.h"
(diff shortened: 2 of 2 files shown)


JBailes:
Converting to draft — **misdiagnosis, do not merge as-is.** The panel doesn't degrade because a persona lacks an *assigned* agent; it degrades because the review **models are disabled/unavailable** on the target host (mimo-v2.5-pro and GLM-5.2 are `enabled=false`; the run only got codex to respond). This fallback dispatches personas to their preset **seat model** — but two of those seats point at exactly the disabled models (`security→mimo`, `reviewer-constructive→GLM`), so it can't help. The real fix is operational (enable/repair the review models, or reseat the `default` preset onto enabled models) and/or a deliberate gate-resilience decision (proceed on an available quorum vs. fail-closed on a missing lens). Holding this pending that direction.

JBailes:
Closing — the real fix was operational: GLM-5.2 and mimo-v2.5-pro are now enabled in the .254 agent roster (plus Kimi added), so the roundtable composes from live review agents via the normal role/persona routing. No code workaround needed; my seat-model fallback was a misdiagnosis.

PR #1272, opened 2026-07-12
fix(agent): enforce per-agent max_parallel in the panel/parallel dispatch path

JBailes:
## Problem

The roundtable panel dispatches each review lens via `agent_run_parallel` → `parallel_worker` → `agent_run_named` / `agent_run_ex`, both of which call `agent_execute()` **directly** — bypassing the *only* place the per-agent concurrency cap is enforced (`provider_catalog_concurrent_acquire`, used solely inside `agent_execute_guarded`).

Result: a single model receives **many simultaneous** reviews with no `max_parallel` ceiling. Observed live: codex (`chatgpt/gpt-5.5`) getting ~15 calls in 2 minutes and returning HTTP 400, and rate-limit-sensitive providers like mistral (`max_parallel=2`) hammered — dragging the gate below quorum → `panel_degraded`. (This compounds with agent reuse: a reused agent can be assigned to several lenses at once.)

## Fix

Both by-name (`agent_run_named`) and by-role (`agent_run_ex`) dispatch now **acquire the agent's slot before `agent_execute` and release after**. At the limit, the agent is reported unavailable (by-name) or skipped for a different agent (by-role → added to `tried`), so the caller spreads load instead of piling on. The acquire failure returns **before** the provider-health record block — saturation is a load signal, not a provider fault, so it never counts as a health failure. `acquire()` returns "unlimited" for `max_parallel<=0` / unknown agents, so single-dispatch callers are unchanged.

## Verification
- Compiles clean under `-Werror`; `test_agent_http`, `test_agent_loop`, `test_agent_parallel` all pass (no regression).
- The concurrency primitive itself is already unit-tested; this wires it into the parallel path. Final proof is live: codex should stop being hammered and the panel should compose within provider limits.

Completes the roundtable-resilience series: #1268 (autonomy retry), #1270 (agent reuse), and this (respect concurrency caps so reuse/fan-out doesn't overload a provider).

### DIFF
--- a/src/server/agent_runtime.c
+++ b/src/server/agent_runtime.c
@@ -193,12 +193,24 @@ int agent_run_ex(agent_config_t *cfg, const char *role, const char *system_promp
       }
       free(hint);
 
+      /* Respect the agent's max_parallel cap (same rationale as agent_run_named):
+       * a saturated agent is SKIPPED for a different one rather than piled on.
+       * Being at the limit is a saturation signal, not a provider fault, so it is
+       * added to `tried` (excluded from the next pick) WITHOUT a health failure. */
+      if (!provider_catalog_concurrent_acquire(ag->name, ag->max_parallel))
+      {
+         free(enhanced);
+         if (ntried < MAX_AGENTS)
+            tried[ntried++] = ag->name;
+         continue;
+      }
       int rc;
       if (use_tools)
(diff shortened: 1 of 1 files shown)


JBailes:
Superseding this targeted per-path guard with a single unified agent-dispatch path (agent_dispatch_one) that enforces the max_parallel cap + health/outcome recording in ONE place — the root fix, not per-path patches. New PR to follow.

PR #1332, opened 2026-07-14
fix(server): push web-GUI clones into aimee-kb for synthesis + embedding

JBailes:
## Problem

Repos added through the webchat GUI (`POST /v1/workspace/clone` and `/v1/workspace/clone-org`) were never ingested into the knowledge base. The clone handlers only ran `index_scan_project` — the server-local lexical index — and never touched aimee-kb. On the .254 full-stack appliance this left the kb with **zero** `/v1/code/scan` requests ever received: `projects`, `files`, `kb_code_unit_jobs`, `code_embeddings`, and `curator_code_unit_vectors` all empty despite a full clone sitting in the webuser workspace, and the GPU embedder idle on the code side.

(The `workspace.add` RPC path does push to the kb, but the GUI clone flow never goes through it — and the only registered workspace on the box was a `detached` dev-box path the server can't read.)

## Fix

After a successful clone (single and bulk), also issue `kb_client_index_scan` — the same push the CLI `workspace add` path uses (`server_runner_endpoints.c`). It POSTs `/v1/code/scan` with file contents attached, so a remote/containerized kb needs no host filesystem access; the kb handler then queues code units for curator synthesis + embedding.

- Best-effort: a down knowledge service never fails the clone (mirrors the existing `kb_client_is_live()` gating).
- The outcome is surfaced in the response as `kb_indexed` (+ `kb_reason` on skip) per repo, and logged via `LOG_WARN("server.workspace", ...)` on failure.
- OpenAPI spec updated for the new response fields.

## Testing

- `gcc -fsyntax-only` on the TU with the project's `-Wall -Wextra -Werror` flags passes (full local build unavailable on this dev box — no libssl headers; relying on CI).
- Existing `unit-test-server-http` routing table is unaffected: the `{}` clone request still 400s before the new code path.

### DIFF
--- a/api/openapi-server-v1.yaml
+++ b/api/openapi-server-v1.yaml
@@ -4300,6 +4300,14 @@ paths:
                 properties:
                   ok: { type: boolean }
                   name: { type: string }
+                  kb_indexed:
+                    type: boolean
+                    description: >
+                      Whether the clone was also pushed to the knowledge service
+                      (/v1/code/scan) for code-unit synthesis + embeddings.
+                  kb_reason:
+                    type: string
+                    description: Why the kb scan was skipped (present when kb_indexed is false).
         "400":
           description: Invalid URL/name, or the project already exists
           content:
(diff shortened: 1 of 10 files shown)


JBailes:
Pushed a second commit: the reason large repos still would not have ingested is that aimee-kb's HTTP listener silently truncates request bodies at 1 MB (src/kb/http/kb_http_conn.c), while kb_client_index_scan pushed the whole tree in one POST — truncated JSON, 400 'invalid json', surfaced as 'no_kb'. Verified live: 17 small projects scanned fine, all 5 large ones failed this way. The commit streams pushed-file scans in 600 KB batches (same pattern as the thin client's /v1/index/ingest streamer; kb scan upserts accumulate and the curator enqueue is an idempotent anti-join). New unit test covers the batching; unit-test-kb-client-index-remote passes locally.

PR #1339, opened 2026-07-14
feat(wfe): re-review panels judge the prior blockers, not the world

JBailes:
Panel seats rotate between rounds, and an unscoped fresh reviewer reliably finds novel objections. Observed live on .254 after the feedback fix (#1337) deployed: a flat 100% `request_changes` rate — ~20 verdicts, zero approves — because the author kept fixing round N's blocker list while round N+1's new reviewers raised a different one. Even a one-line-change packet could not pass. With the gate's any-blocker-loops rule, unscoped seat rotation makes convergence statistically impossible.

Fix: carry the previous round's persisted blockers into `wfe_review_packet_t`; when present, the panel prompt declares a **RE-REVIEW** — approve if those blockers are adequately addressed and no **new high-severity** defect is found; stylistic preferences and below-high-severity issues go to `comment` (non-blocking) instead of `request_changes`. First reviews are unchanged, as is the fail-closed rule (any effective `request_changes` still loops; malformed verdicts still coerce to `request_changes`).

Tests: `unit-test-wfe-panel-verdict`, `unit-test-wfe-engine`, `unit-test-wfe-autonomy` pass locally.

### DIFF
--- a/src/server/wfe_live_panel.c
+++ b/src/server/wfe_live_panel.c
@@ -91,22 +91,40 @@ static char *build_prompt(const char *persona, const wfe_review_packet_t *pkt)
    const char *diff = (pkt->diff && pkt->diff[0])
                           ? pkt->diff
                           : "(no code diff — review the plan/proposal artifact against the ask)";
-   size_t cap = 2048 + strlen(focus) + strlen(proposal) + strlen(diff);
+   const char *prior = (pkt->prior_blockers && pkt->prior_blockers[0]) ? pkt->prior_blockers : "";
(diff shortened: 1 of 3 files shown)


JBailes:
Closing: scoping re-reviews to the prior round's blockers undermines the roundtable's purpose — a fresh lens surfacing a NEW genuine defect is exactly what the gate exists for. Convergence needs a different mechanism (e.g. seat stickiness across rounds so a reviewer can see their objections addressed), not a narrowed review scope.

PR #1347, opened 2026-07-14
wfe forge: pin run_cmd cwd to the repo checkout in every live forge op

JBailes:
## Problem
s2's first rt_gate pass reached `pr.open` and every push failed with `fatal: not a git repository (or any parent up to mount point /var/lib)` — 20 loops in one second, then a `max_iters` park.

## Root cause
All live forge git/gh commands go through `mcp_git_run`, which (a) executes in the **thread-local run_cmd cwd** and (b) injects the vaulted forge credential only when that cwd lies inside a registered workspace. The pr/ci/merge executors run on scheduler worker threads that never set a cwd, so git ran in the daemon's working directory — no repo, no credential injection, and `gh` couldn't resolve the repo slug. The forge ops received the work item's repo binding as their first argument and discarded it (`(void)repo`).

## Fix
`forge_cwd_push`/`forge_cwd_pop` pin the thread cwd to `wfe_repo_local(repo)` for the duration of each op — `ci_status`, `mergeable`, `is_merged`, `merge`, `open` (covering trunk resolution, the branch push, and the PR create), and `publish_base` — restoring the previous cwd on every exit path. Same bug class (and same idiom) as the verify-provider cwd pin fixed earlier.

Verified: compiles clean; behavior is deterministic plumbing exercised by the live deployment (the forge is integration-gated, no unit harness links it).

### DIFF
--- a/src/server/wfe_live_forge.c
+++ b/src/server/wfe_live_forge.c
@@ -23,12 +23,35 @@
 #include "config.h"
 #include "log.h"
 #include "mcp_git.h"
+#include "util.h" /* run_cmd_set_cwd / run_cmd_get_cwd (thread-local git cwd) */
 #include "wfe_blocks.h"
 #include "wfe_iface.h" /* wfe_autonomous_base / wfe_autonomous_target_ok (merge-target rail) */
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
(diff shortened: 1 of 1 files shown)


JBailes:
Superseded: wrong mechanism. The live forge should not shell git/gh through mcp_git_run at all — forge credentials must never leave aimee-server. Rework lands the forge on the in-process REST layer (git_pr_api) + FD-mode argv push (git_cred_inject memfd), per operator direction.

PR #1355, opened 2026-07-15
fix(kb/curator): reclaim orphaned job leases; unblock extract_doc throughput; make sidecar failures diagnosable

JBailes:
Found while investigating why ingestion looked stuck on the `.254` appliance. The pipeline turned out to be **healthy but leaking work**: both queues drain cleanly in id order, code-unit extraction runs at ~1,700 jobs/hr. The real problems were a wedged lease and a set of failures nobody could diagnose.

## What this fixes

**1. `extract_doc` jobs orphaned in `running` were never recovered.** `ce_claim_job` only ever selects `status='pending'`, so a job whose worker crashed or whose sidecar wedged stays `running` forever — never extracted, and pinning a db2 pool member long past its 300s ceiling. On `.254`, job 3401 sat wedged for 15h:

```
3401 | running | attempts=1 | kb.curator.drain | claimed_at 2026-07-14 19:19:32
aimee: db2.pool: member 0 leased 54483876ms (> 300000ms ceiling) — missed lease_end?
```

54483876ms is 15.13h — the same incident, not two. The code-unit stage has had `ccu_reclaim_stale_running` since it shipped; `kb_async_jobs` never got one. This mirrors it, scoped to `kind='extract_doc'` so the other stages sharing that table keep their own claim lifecycle.

**2. `pclose(3)` returns a wait(2)-encoded status, not an exit code.** Reporting it raw wrote `sidecar exited 256` for a sidecar that merely `exit(1)`'d (`1 << 8`). That opaque number reads like an exotic fault and sends you hunting a signal that never happened. Now decoded into exit / signal / timeout — on a box where llama-server holds 53% of RAM, an OOM kill and a clean non-zero exit are different incidents. Factored into `kb_curator_sidecar.c`, which both popen callers already sit under.

**3. `mkstemp` failures discarded `errno`.** 5,114 code-unit jobs on `.254` died on a bare `mkstemp failed` — which cannot distinguish `ENOSPC` (root fs is 96% full) from `EMFILE` from `EACCES`. **This does not fix those failures; it makes the next one diagnosable.** Also honours `TMPDIR`, as `code_collect.c` already does, so the spool can be moved off a full filesystem.

**4. The generic async drain claimed jobs of ANY kind.** `db2_kb_service_async_queue_claim_next` had no `kind` filter, while its dispatch marks anything it doesn't recognize `failed`. One call to that endpoint would have silently destroyed all 5,682 pending `extract_doc` jobs. Latent (it isn't on the drain thread's path) but a live footgun.

## Testing

- `reclaim_stale_running` — asserts an orphan is reclaimed, while an in-flight job and another `kind` are left untouched. **Verified failing without the fix** (neutered the reclaim call; the 9001 assertion fires), so it's a real guard rather than a vacuous one.
- `describe_wait_status` — pins each status class, including the exact `256`-means-`exit 1` regression, and that 124 is only read as a timeout when the caller actually wrapped in `timeout(1)`.
- Ran `unit-test-curator-queue`, `unit-test-curator-code-unit`, `unit-test-kb-curator-llm`, `unit-test-curator-version`, `unit-test-curator-custom-stages`, `unit-test-db2-pool` — all pass. `make line-check`, `make module-boundary-check`, and clang-format-19 clean.

## Deliberately NOT changed

- **`ccu_mark_retry_or_fail(job_id, opts->max_attempts, opts->max_attempts, ...)`** at the body-read failure looks like an argument bug, but `kb_curator_extract.c:380` uses the same idiom with a log line that says "marking failed" — it's a deliberate fail-fast for non-retryable conditions. Left alone.
- **Re-queuing the 5,229 already-failed code-unit jobs.** They're at `attempts=3` and the backfill's `ON CONFLICT DO NOTHING` will never reset them, so they need an explicit requeue. Doing that now would just re-fail them until the `mkstemp` root cause is known — which is what fix 3 unblocks. Worth a follow-up once we have a real errno.

## Known gaps (follow-ups, not in scope here)

- `memory_facts` shares `kb_async_jobs` and has the same missing-reclaim gap. No wedge observed, so not fixed blind.
- `extract_doc` throughput is ~1 doc/90s single-threaded, with 5,682 pending — roughly 6 days to drain. That's a capacity question, not a bug.
- Root fs on `.254` is at 96%. Operational, but plausibly related to fix 3.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -22,7 +22,7 @@ aimee config set <key> <value>    # set one value
 
 Structured options (arrays, nested objects — e.g. `ensemble.reference_models`) are not CLI-settable; they are written into the config file under the sections listed at the end.
 
-## CLI-settable keys (179)
+## CLI-settable keys (180)
 
 | Key | Type | Description |
 |-----|------|-------------|
@@ -100,6 +100,7 @@ Structured options (arrays, nested objects — e.g. `ensemble.reference_models`)
 | `kb_curator_extract_code_enabled` | bool | — |
 | `kb_curator_extract_code_workers` | int | — |
 | `kb_curator_extract_docs_enabled` | bool | — |
+| `kb_curator_extract_docs_workers` | int | — |
 | `kb_curator_index_claims_enabled` | bool | — |
 | `kb_curator_index_code_unit_enabled` | bool | — |
 | `kb_curator_index_narrative_enabled` | bool | — |
@@ -206,7 +207,7 @@ Structured options (arrays, nested objects — e.g. `ensemble.reference_models`)
 | `virtual_context_enabled` | bool | Enable virtual-context assembly. |
 | `wfe_live_forge_enabled` | bool | — |
 
-> **Undocumented** (add to `CFG_KEY_DESC` in gen-reference-docs.py): `audit_action_enabled`, `code_trust_actuation_enabled`, `kb_client_bearer_token`, `kb_client_url`, `kb_curator_cross_repo_graph_enabled`, `kb_curator_custom_stages`, `kb_curator_detect_contradictions_enabled`, `kb_curator_extract_code_enabled`, `kb_curator_extract_code_workers`, `kb_curator_extract_docs_enabled`, `kb_curator_index_claims_enabled`, `kb_curator_index_code_unit_enabled`, `kb_curator_index_narrative_enabled`, `kb_curator_link_artifacts_enabled`, `kb_curator_projection_graph_enabled`, `kb_curator_promote_entity_enabled`, `kb_curator_resolve_entities_enabled`, `kb_curator_stage_order`, `kb_curator_synthesize_enabled`, `kb_curator_user_presets`, `kb_evidence_embed_enabled`, `kb_mode`, `llm_embed_backend`, `llm_embed_gpu`, `llm_embed_host`, `llm_embed_tier`, `llm_rerank_backend`, `llm_rerank_endpoint`, `llm_rerank_gpu`, `llm_rerank_host`, `llm_rerank_tier`, `llm_synth_backend`, `llm_synth_endpoint`, `llm_synth_gpu`, `llm_synth_host`, `llm_synth_model`, `llm_synth_tier`, `wfe_live_forge_enabled`
+> **Undocumented** (add to `CFG_KEY_DESC` in gen-reference-docs.py): `audit_action_enabled`, `code_trust_actuation_enabled`, `kb_client_bearer_token`, `kb_client_url`, `kb_curator_cross_repo_graph_enabled`, `kb_curator_custom_stages`, `kb_curator_detect_contradictions_enabled`, `kb_curator_extract_code_enabled`, `kb_curator_extract_code_workers`, `kb_curator_extract_docs_enabled`, `kb_curator_extract_docs_workers`, `kb_curator_index_claims_enabled`, `kb_curator_index_code_unit_enabled`, `kb_curator_index_narrative_enabled`, `kb_curator_link_artifacts_enabled`, `kb_curator_projection_graph_enabled`, `kb_curator_promote_entity_enabled`, `kb_curator_resolve_entities_enabled`, `kb_curator_stage_order`, `kb_curator_synthesize_enabled`, `kb_curator_user_presets`, `kb_evidence_embed_enabled`, `kb_mode`, `llm_embed_backend`, `llm_embed_gpu`, `llm_embed_host`, `llm_embed_tier`, `llm_rerank_backend`, `llm_rerank_endpoint`, `llm_rerank_gpu`, `llm_rerank_host`, `llm_rerank_tier`, `llm_synth_backend`, `llm_synth_endpoint`, `llm_synth_gpu`, `llm_synth_host`, `llm_synth_model`, `llm_synth_tier`, `wfe_live_forge_enabled`
 
 ## Config-file sections (53)
 
--- a/src/config_fields.c
+++ b/src/config_fields.c
@@ -367,6 +367,8 @@ const config_field_t config_fields[] = {
      * nested kb.curator.* YAML the KB reads on its next load. */
     {"kb_curator_extract_docs_enabled", offsetof(config_t, kb_curator_extract_docs_enabled),
      sizeof(int), 0, CFG_BOOL},
+    {"kb_curator_extract_docs_workers", offsetof(config_t, kb_curator_extract_docs_workers),
+     sizeof(int), 0, CFG_INT},
     {"kb_curator_stage_order", offsetof(config_t, kb_curator_stage_order),
      sizeof(((config_t *)0)->kb_curator_stage_order), 0, CFG_STRING},
     {"kb_curator_user_presets", offsetof(config_t, kb_curator_user_presets),
--- a/src/config_kb_curator.c
+++ b/src/config_kb_curator.c
@@ -37,6 +37,7 @@ void config_kb_curator_defaults(config_t *cfg)
    cfg->kb_typed_facts_promote_threshold = 3;
    cfg->kb_curator_extract_code_enabled = 1;
    cfg->kb_curator_extract_code_workers = 1;
+   cfg->kb_curator_extract_docs_workers = 1;
    cfg->kb_curator_resolve_entities_enabled = 1;
    cfg->kb_curator_index_narrative_enabled = 1;
    cfg->kb_curator_index_claims_enabled = 1;
@@ -167,6 +168,11 @@ int config_parse_kb_curator(config_t *cfg, const cJSON *root)
       const cJSON *enabled = cJSON_GetObjectItemCaseSensitive(extract_docs, "enabled");
       if (enabled && cJSON_IsBool(enabled))
          cfg->kb_curator_extract_docs_enabled = cJSON_IsTrue(enabled) ? 1 : 0;
+      /* workers: dedicated extract_doc drain threads (see config.h). Clamped
+       * to [1,8]; non-numeric / out-of-range values keep the default. */
+      const cJSON *workers = cJSON_GetObjectItemCaseSensitive(extract_docs, "workers");
+      if (workers && cJSON_IsNumber(workers) && workers->valueint >= 1 && workers->valueint <= 8)
+         cfg->kb_curator_extract_docs_workers = workers->valueint;
    }
 
    const cJSON *extract_code = cJSON_GetObjectItemCaseSensitive(curator, "extract_code");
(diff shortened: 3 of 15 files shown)


JBailes:
Pushed a second commit closing the two gaps the first one deliberately left open.

**`memory_facts` had the identical missing-reclaim bug.** It shares `kb_async_jobs` and its claim also only selects `pending`, so a wedged job stranded forever. No production wedge was observed — but the defect is the same and the SQL is the same shape. Scoped to `kind='memory_facts'`.

The two reclaim tests now pin the kind scoping **from both sides**: the `extract_doc` test asserts it refuses to touch the stale `memory_facts` row, and the `memory_facts` test asserts its own drain then reclaims that same row. Both verified failing without their fix.

**`extract_doc` throughput — the cause was structural, not model speed.** Measured per-hour completions on `.254`, same window:

| stage | completions/hr |
|---|---|
| `extract_code` | ~1,750 |
| `extract_doc` | ~30 |

Both stages are *one sidecar call per job*, so 1:55 is a scheduling artifact, not inference cost. Individual doc jobs take 13–22s, but the next isn't claimed for ~90s — the gap is the shared LLM lane doing everything else. The reason: `extract_code` reads `kb_curator_extract_code_workers` and gets a dedicated worker pool (the `.254` KB container sets `workers: 4`), while `extract_doc` has only the shared lane, which spends most of each pass on the other LLM stages (a code unit, `resolve_entities`, `synthesize`, `promote_entity`).

So this adds `kb_curator_extract_docs_workers` as the exact analogue of the code knob — default 1 (**no behavior change**), N>1 spawns N-1 dedicated doc workers. Claims are already transactional (`FOR UPDATE SKIP LOCKED`), so workers can't double-claim.

**Note this is inert until configured**, deliberately: enabling it is an operator config change, exactly as `extract_code` works today (the shipped `deploy/container/aimee.yaml` sets neither — the `workers: 4` on `.254` was set operator-side). For `.254`, `kb.curator.extract_docs.workers: 4` is the one-line change. I kept the default at 1 rather than picking a number for every deployment, since the right value tracks the synth backend's parallel slots.

All affected suites pass locally (`curator-queue`, `curator-code-unit`, `typed-facts`, `kb-curator-llm`, `curator-version`, `curator-custom-stages`, `db2-pool`), clang-format-19 clean, and `docs/gen/configuration.md` regenerated so `docs-gen-check` stays green.

JBailes:
**Roundtable verdict: not a pass.** 3 blocking findings and 5 suggestions — and its own verification pass rejected 6 more (correctly, including a bogus `_GNU_SOURCE` claim). Pushed `8bc743e5` acting on the real ones.

**The important one is a genuine design flaw in my own fix.** `kb_curator_sidecar_run` runs a bare `popen` with no timeout, unlike the code-unit stage which wraps its command in `timeout -k 10 300`. **A lease reclaim is only safe if the work under it is bounded well below the lease.** Otherwise a job whose worker is *still running* gets re-armed and handed to a second worker — duplicate artifacts, and the original thread later clobbering the status of a job it no longer owns. The reclaim I added would have introduced exactly that for `extract_doc`, whose sidecar path is the unbounded one. The provider path was already bounded (`PROVIDER_CLIENT_DEFAULT_TIMEOUT_MS` = 120s); only the fallback was open-ended. Now bounded at 300s — well under the 15-minute lease, far above the 13–22s a real doc extraction takes. `kb_curator_llm.c` is its sole production caller, so the blast radius is contained.

**The throttle race became real because of my own second commit.** `static time_t last_run` is now genuinely shared: `kb_curator_extract_one` runs on the main LLM lane *and* on every `extract_docs` worker I added. Now mutex-guarded. It also armed the 60s cooldown *before* the UPDATE ran, so a transient db2 failure silently suppressed the next reclaim; now only a completed UPDATE arms it, and prepare/step failures WARN instead of vanishing. `kb_memory_facts` needs no lock (single caller on the drain thread) but had the same arm-on-failure bug.

The kind allowlist is now a shared `KBS_ASYNC_DRAIN_KINDS_SQL` constant rather than a comment, with the dispatch's unreachable branch made loud if the two ever drift.

Two findings I'd call the panel's best, because they caught this PR contradicting itself:

- **The reclaim destroyed the diagnostic it exists to preserve.** Marking a job terminal overwrote `last_error` — the message from the attempt that actually died — with `'stale running lease reclaimed'`. Now only filled when the worker left none.
- **`"sidecar killed by signal 9"` overclaims.** 128+n is the *shell's* convention; a process can `exit(137)` itself and the two are indistinguishable from here. Asserting a kill would fake an OOM and send someone hunting a phantom — the exact failure mode this PR exists to eliminate. Now reports the exit code as fact and the signal as a reading; a real `WIFSIGNALED` death still states it outright.

Re-running the panel against the updated diff.

JBailes:
**Re-review: the earlier blocking findings are resolved** — every one it re-raised (mutex-across-DB-call, arm-on-success retry storm, `PTHREAD_MUTEX_INITIALIZER` on a function-local static, the 124 ambiguity, `volatile stop`) was **contradicted at its own verification pass**. Score: verified=7 capped=1 rejected=8.

**But it caught a regression I introduced, and it was right.** Pushed `1d03c378`.

The `timeout -k 10 300 sh -c "%s"` wrapper I added interpolates the configured command into a **double-quoted string parsed by a second shell**. Under the bare `popen` it replaced, a command containing `"`, `$`, or a backtick worked fine; wrapped, it gets re-parsed. Demonstrated against a real shell:

| | output |
|---|---|
| bare `popen` (behaviour to preserve) | `lit::"q"` |
| raw interpolation (the bug) | `lit::q` — outer shell ate the `\"` escapes |
| quoted (fixed) | `lit::"q"` ✓ |

Fixed with `kb_curator_shell_quote` (POSIX single-quote form, `'\''` for embedded quotes), applied to both the command and the temp path. Overflow is a hard error, never a truncated string a caller might still run. **The flaw was pre-existing in `ccu_invoke_sidecar`** — which is where I copied the wrapper from — so it's fixed there too. `.254`'s configured command has no metacharacters, so nothing there was affected either way.

Also in this commit: **`extract_code`'s reclaim had the same three defects fixed earlier in `ce_`/`mf_` — and unlike those, its race is live today.** `ccu_reclaim_stale_running` is reached from the main lane *and* from every `extract_code_workers` thread (`.254` runs 4), so the unguarded `static time_t last_run` is a real race now, not a latent one. It also overwrote `last_error`, destroying the diagnostic on exactly the jobs an operator needs to diagnose.

Both new tests verified failing without their fix. `test_reclaim_stale_running_code_unit` must run first in its binary — the throttle is a process-wide static that an earlier successful reclaim would arm, which would make it silently skip.

**Findings I'm deliberately not acting on**, all pre-existing and none introduced here:
- **Process-tree escape**: a sidecar descendant that creates a new session can outlive both `timeout` signals. Real, but deep (needs process-group containment) and unchanged by this PR — the python sidecar spawns no descendants.
- **No aggregate LLM concurrency budget**: doc and code worker counts clamp independently to [1,8] each, with no cap tied to llama-server's slots. Worth doing, but the doc pool ships **default 1** (inert), so this PR adds no new contention.
- **`volatile int stop`** is formally a data race for all worker threads. Pre-existing for the code/index lanes; its own verifier contradicted it.

JBailes:
**Final panel: one blocking finding, and it was right.** Pushed `ef087dff`.

Its verdict on the rest was clean: *"the new single-word quoter itself is correct for NUL-terminated POSIX-shell strings: the standard embedded-quote sequence, empty input, literal newlines, and capacity checks are sound, and the stated buffers leave ample room without reachable full-command truncation. Quoting the pathname is valid after `<`, and the ccu reclaim mutex is unlocked on every visible path with `last_run` armed only after AIMEE_PG_DONE."*

But it caught that my quoting fix preserved the command's **bytes** and not its **parse context** — so the "timeout wrapper changes nothing" invariant was *still* false, just more subtly. The old form `sh -c "<cmd> < tmp"` parses the redirection alongside the command, so for a pipeline it binds to the **last** command. Redirecting the wrapper's stdin instead hands the file to the **first**. Verified against a real shell with `printf ABC | wc -c` and a 2-byte request:

| form | output |
|---|---|
| bare `popen` (correct) | `2` — wc reads the request |
| wrapper stdin redirect (my bug) | `3` — wc counted printf's output |
| redirect inside, path as `$1` (fixed) | `2` ✓ |

Fixed as recommended: append the redirection to the script and pass the path as a positional — `sh -c '<cmd> < "$1"' sh <tmppath>`. `$1` sits inside the single-quoted script, so only the *inner* shell expands it and the path never re-parses. Applied to both sidecars, with the pipeline case pinned as a test (verified failing against the wrapper-stdin form).

Also hedged the 124 message — `timeout(1)` returns 124 at its deadline but also *propagates* a wrapped command's own exit 124, so asserting "timed out" was the same overclaim as the 128+n case fixed earlier.

Worth noting the shape of this review: three rounds, each finding exactly one real defect in the previous round's fix, with the panel's own verification pass rejecting 14 findings across rounds. Every blocking finding was in code I wrote in *this* PR, and two of them existed only because of an earlier fix in it.

JBailes:
**Confirmation panel: two blocking findings. Acted on one, rejected the other with evidence.** Pushed `ebba0d8c`.

**Acted on — a configured command could hijack the redirection path.** With `sh -c '<cmd> < "$1"' sh <path>`, a command containing `set --` rebinds `$1` *before* the redirection expands, so the sidecar silently reads a different file. Measured against a real shell (2-byte request; `/etc/hostname` is 5):

| form | pipeline | `set --` |
|---|---|---|
| bare `popen` (baseline) | 2 | 2 |
| `$1` positional | 2 | **5** — wrong file |
| quoted literal path | 2 | 2 ✓ |

A literal can't be reassigned. I checked all six constructs that matter against the bare-popen baseline — simple, env-prefixed, pipeline, `&&` list, subshell, `set --` — the literal form matches on all six; the `$1` form matched five.

**Rejected — "appending ` < <path>` breaks a command ending in `#`, `&`, or an unterminated quote."** The old code appended ` < %s` by the same string concatenation, so these parse exactly as they always have: `echo hi # < tmp` printed `hi` before and prints `hi` now. Faithfully preserved, not a regression.

Its recommended fix is actively wrong, and I checked before dismissing it: `sh -c 'eval "$1" < "$2"'` redirects **eval's** stdin, so `printf ABC | wc -c` yields **3** where the bare popen yields **2** — reintroducing the precise pipeline defect the *previous* round flagged as blocking. Two rounds of this panel disagree with each other; the empirical baseline settles it.

Its verification pass also rejected 3 of its own findings, including a claim that an embedded NUL truncates the quoter — impossible, since the input is a C string.

That's four review rounds. Every blocking finding landed in code written in this PR, and three of the four existed only because of an earlier fix within it.

PR #1363, opened 2026-07-15
fix(curator): split reasoning at the wire boundary; delete the think-strippers

JBailes:
Found by requeuing the last 22 failed `code_unit` jobs on `.254` and asking why the two that **weren't** OOM casualties failed. Both were deterministic — and the pair names the bug:

| symbol | file |
|---|---|
| `strip_think` | `scripts/curator-synthesize.py` |
| `strip_thinking_blocks` | `src/server/agent_bridge.c` |

Both are functions whose job is **stripping `<think>` blocks** — and the sidecar strips think blocks from the model's answer with the same naive rule:

```python
if "</think>" in text:
    text = text.rsplit("</think>", 1)[1].strip()
```

That splits on the **last** `</think>` **anywhere** in the response, including inside the JSON payload. Asking the model to summarise a function about think-blocks makes it quote the tag, so `rsplit` cut the JSON mid-string. A valid answer became `LLM returned non-JSON: Expecting value: line 1 column 1 (char 0)`, and the job died after three attempts.

Reproduced exactly — same error string, same truncation point:

```
model output parses as JSON?   True
after strip_fences() ->        'preamble; if the tag exists, the desired content follows it.'
parses after?                  NO -> Expecting value: line 1 column 1 (char 0)
```

## The fix

The tag is a **preamble** — reasoning models prepend it — so anchor the strip to the front and leave any later occurrence to the content.

Worth noting: `strip_fences` already called `_first_json_object` (balanced, string-aware, drops surrounding prose), which handles a preamble perfectly well on its own. The `rsplit` ran **first** and destroyed the JSON before that safe parser ever saw it.

Same rule, same bug, in `curator-synthesize.py` — fixed there too.

## Testing

`scripts/tests/test_curator_strip_fences.py` covers the regression (a mention inside the payload survives), a real preamble still being stripped, **both together**, an unterminated preamble, code fences, and surrounding prose. **Verified failing (2 errors) against the old `rsplit`.**

These scripts had no test coverage at all, so I added a `curator-sidecar-parse-check` target wired into `make lint` — `strip_fences` decides whether a model's answer is usable *at all*, which is too load-bearing to leave ungated.

## Context

Only visible because #1357 stopped the sidecar's own error being freed unread — previously all of these read `sidecar exited 256`. The other 20 of the 22 were LLM 503s from an OOM (14GB box, 11GB synth model) and are unrelated to this.

### DIFF
--- a/scripts/check-sidecar-clients.py
+++ b/scripts/check-sidecar-clients.py
@@ -105,10 +105,76 @@ def check_llm_chat() -> None:
     print("  llm-chat.py: ok")
 
 
+def _chat_stub(message: dict):
+    """A stub /v1/chat/completions returning the given assistant message."""
+
+    class ChatStub(BaseHTTPRequestHandler):
+        def log_message(self, *a):  # quiet
+            pass
+
+        def do_POST(self):
+            assert self.path.rstrip("/") == "/v1/chat/completions", self.path
+            length = int(self.headers.get("content-length", "0") or "0")
+            self.rfile.read(length)
+            body = json.dumps({"choices": [{"message": message}]}).encode()
+            self.send_response(200)
+            self.send_header("content-type", "application/json")
+            self.send_header("content-length", str(len(body)))
+            self.end_headers()
+            self.wfile.write(body)
+
+    return ChatStub
+
+
+def _chat(message: dict, env_extra: dict | None = None) -> str:
+    server, url = _serve(_chat_stub(message))
+    try:
+        env = {"LLM_ENDPOINT": f"{url}/v1", "LLM_MODEL": "stub", "LLM_API_KEY": ""}
+        env.update(env_extra or {})
+        return _run("llm-chat.py", env, "say something")
+    finally:
+        server.shutdown()
+
+
+def check_llm_chat_reasoning_split() -> None:
+    """THE CONTRACT: stdout carries the answer, never the reasoning.
+
+    llm-chat.py is the ONE wire boundary for every sidecar (curator-extract,
+    curator-synthesize, llm-rewrite, learning-synthesize). Consumers must never
+    re-derive this from text — doing so is what cut valid JSON mid-string when a
+    payload merely mentioned the tag.
+    """
+    # 1. Compliant server (llama.cpp --jinja): reasoning already separated.
+    out = _chat({"content": "the-answer", "reasoning_content": "my hidden reasoning"})
+    assert "the-answer" in out, f"answer missing: {out!r}"
+    assert "hidden reasoning" not in out, f"reasoning leaked to stdout: {out!r}"
+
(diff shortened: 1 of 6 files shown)


JBailes:
**Reworked per review — don't strip better, don't strip at all.**

The first version of this PR made the stripping *less wrong* (front-anchored). That was still the wrong shape: reasoning is a distinct thing from the answer, and the split belongs at the **wire boundary**, once, where the response is still structured — not re-derived from text by every consumer.

`llm-chat.py` is that boundary: every sidecar popens it. And the contract was **already being re-derived three different ways**:

| consumer | think handling |
|---|---|
| `curator-extract.py` | `rsplit` on the **last** `</think>` anywhere ← the bug |
| `curator-synthesize.py` | same rule, copied |
| `llm-rewrite.py` | **nothing** |
| `learning-synthesize.py` | **nothing** |

The last two are correct today *only* because llama.cpp under `--jinja` already returns a separate `reasoning_content` and `llm-chat.py` reads `message.content`. So the contract — *stdout is the answer* — already existed and was already relied upon. It just wasn't implemented or stated anywhere, and the two that hedged got it wrong.

`split_reasoning()` now states it: trust `reasoning_content` when the server provides it; otherwise strip a `<think>` **prefix** (an occurrence anywhere else is content — e.g. a docstring about stripping think blocks). **Both strippers are deleted.**

And the reasoning is no longer thrown away: `LLM_EMIT_REASONING=1` logs it to stderr — out of the answer, still visible to the kb logs. Today's recurring theme was diagnostics being silently discarded; this stops doing that to the model's own reasoning.

**Verified end-to-end against the live synth model on `.254`:**

```
STDOUT:                DONE          <- the answer, nothing else
think tags in stdout:  0
STDERR:                llm-chat: reasoning (323 chars): The user wants me to...
```

**Tests.** `check-sidecar-clients.py` gains `check_llm_chat_reasoning_split` against stub servers of **both** shapes (separated `reasoning_content`; inlined `<think>` prefix), that a `</think>` in the *answer* survives, and both together. Verified failing **two ways** — with the old `rsplit` (eats a legitimate tag) and with no split at all (leaks reasoning to stdout), so it pins the contract from both sides. `test_curator_strip_fences.py` is rewritten: it previously asserted `strip_fences` strips a preamble — it encoded the responsibility that has now moved.

**The real follow-up:** `aimee_ir.h` already models exactly this as `AIMEE_BLK_THINKING`, with frontend/backend adapters pivoting through a protocol-neutral IR — *"there is NO direct client-shape → provider-shape translation"*. The curator predates that IR and references it nowhere (`src/kb/` and `provider_client.c` have zero references; `provider_client.c` reads `message.content` and silently drops `reasoning_content`). Routing the curator through the IR is the durable fix and deserves its own proposal. This puts the seam in the right place meanwhile — and deletes code rather than adding it.

PR #1522, opened 2026-07-18
economizer: remove the vestigial reduce.* / economizer.enabled|aggressive surface

JBailes:
Follow-up to #1516. That PR introduced the single `economizer = off|safe|aggressive` tier and made it authoritative, but layered it **on top of** the old per-lever booleans instead of removing them. This deletes that vestigial surface so the tier is the only economizer control — the cleanup #1516 explicitly deferred.

## What's removed
13 `config_t` fields (struct + parse + save + `config_fields[]` + defaults): `economizer_enabled`, `economizer_aggressive`, `reduce_measure_enabled`, `reduce_delegate_seam`, `reduce_history_fold`, `reduce_compress`, `reduce_gateway_mutate`, `reduce_command_filter`, `reduce_freeze_guard_enabled`, `reduce_freeze_guard_horizon`, `reduce_gateway_seam` (+`_explicit`), `reduce_gateway_session_disable_ttl_ms`.

Plus the now-dead helpers `config_apply_reduce_consistency` (gateway-seam synthesis) and `config_reduce_validate` (TTL guard), their callers/decls, the `reduce:` save block, and the two startup WARN blocks.

## Behavior-preserving derivation
Every removed lever defaulted **ON**, so each call site collapses to the tier condition:
- delegate seam always runs (measure at minimum); `history_fold = tier!=off && !chatgpt`; `compress = tier==aggressive` (`agent_runtime.c`)
- `tool_condense = tier!=off` (`tool_condense.c`)
- gateway shadow-measure + live mutation = `tier==aggressive` via `econ_gateway_mutate_on` (`openai_chat.c`, `anthropic_http.c`, `gateway_mutate_wire.c`)
- freeze-guard horizon (1) and gateway session-disable TTL (1h) are now fixed constants in `context_reduce.h`

Default tier is `safe`, so default runtime behavior is byte-identical to before.

## Back-compat
- The deprecated `economizer: {enabled, aggressive}` object form still parses and maps onto the tier.
- A legacy `reduce:` block is silently ignored (still schema-known → no error).
- Fixed the `economizer` schema entry to `SCHEMA_STRING` (was `SCHEMA_OBJECT` — #1516 left it, so the now-canonical string form warned on every load).

## Not in scope
The tier isn't yet in the flat `config_fields[]` table — it's an enum surfaced as a string, which that scalar get/set machinery doesn't model. Set it via the `economizer:` YAML key. `TODO(A)` marks exposing it in the typed/GUI surface once the table grows enum-as-string support. (The separate `fold.*` / `compact.*` / `coord_closet.*` tuning knobs are untouched — a distinct later follow-up.)

## Test
`test_config_economizer` rewritten to the tier model; `test_tool_condense`, `test_config`, `test_config_snapshot`, `test_gateway_mutate_wire`, and `ir_ingress_stubs` adjusted (added a weak `econ_gateway_mutate_on` stub for tests that link the ingress path without `config.o`). Full `make unit-tests` green, including `unit-test-ir-crossproto-egress` (the egress byte-identity / determinism gate). Net **−405 lines**.

### DIFF
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -19,7 +19,7 @@ request), unless noted otherwise below.
   is the only network hop; there are no server changes per field.
 - **Control inference:** the control type is inferred from the value's JSON type: a
   **boolean → toggle**, a **number → number field**, a **string → text field**. Grouping is
-  by the dotted key prefix (e.g. everything under `reduce.` forms one section).
+  by the dotted key prefix (e.g. everything under `autonomy.` forms one section).
 - **Quick panel:** a smaller gear dropdown in the top bar (`SettingsPanel`) surfaces the
   five most-used toggles (autonomous mode, cross-verify, eco mode, reasoning cap, max
   iterations). The full page is the comprehensive home for everything else.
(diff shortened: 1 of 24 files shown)


JBailes:
Superseded — closing without merge. While this sat, PR #1523 (commit c4cbcd73 "economizer: delete the old reduce.*/two-tier config surface; tier-keyed internal presets (D+E)") landed on testing and performed the identical vestigial cull, plus the tier-keyed internal presets follow-up. Every reduce.* / economizer.enabled|aggressive field this PR removes is already gone from testing, so there is nothing left to merge. The other config-cleanup PRs in this series (#1518, #1524, #1526, #1528, #1530, #1531) are unaffected.

PR #1532, opened 2026-07-18
fix(wfe): foreach parent parks slices_running (driveable), not pending_human

JBailes:
## Problem

A `foreach.workflow` parent parked `pending_human` the instant it entered the `slices` node — before any child slice could pass or fail — as its "waiting for children" state. That collided with the autonomy scheduler's driveable filter (#1519): `pending_human` is correctly non-driveable, so the scheduler skipped the parent, its fan-in aggregation never re-ran, and the whole fan-out wedged forever even after every slice merged. It also mislabeled a routine wait as a human gate, when the only legitimate human stop in an autonomous build is the final PR.

## Fix

Introduce `WFE_PAUSE_SLICES_RUNNING` (`"slices_running"`): a self-resolving, driveable park for the two "children still running" cases in `exec_foreach_workflow`.

- `wfe_autonomy_run` re-aggregates a `slices_running` parent from the cheap local child tally each sweep — advancing once every slice merges, or converting to `pending_human` the moment a slice fails so an operator resolves the dead slice. It does *not* re-run the block or write an audit event while slices are merely still running, so a long fan-out neither spams the audit log nor inflates the turn cap.
- `pending_human` now covers only genuine human waits: no spawner, a fan-out failure, or a failed slice.
- `wfe_sched_driveable` drives `slices_running` so the parent is re-checked every sweep.

## Tests

`test_wfe_foreach` now asserts the running parent parks `slices_running` and auto-resumes once all children merge; adds a `slices_running -> pending_human` failure-transition case. `test_wfe_autonomy` unchanged and passing. `make lint` green.

### DIFF
--- a/src/db1/wfe_store.h
+++ b/src/db1/wfe_store.h
@@ -18,7 +18,7 @@ typedef struct
    char pause_reason[32]; /* "" | pending_human | panel_degraded | budget_exceeded |
                              panel_unreachable | ci_pending | merge_pending |
                              turn_cap_exceeded | wall_cap_exceeded | stuck |
-                             operator_paused */
+                             slices_running | operator_paused */
    char paused_state[64];
    char content_hash[72];
    /* The forge ref returned by g_forge->open (PR number/url), opaque to the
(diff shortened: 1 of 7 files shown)


JBailes:
Roundtable review (aimee panel, mode=review): **APPROVE** — "No issues found by the panel." 25 deliberation concerns were all contradicted at verification (verified=0). Notable checks that were refuted: TOCTOU on the child tally (clear_pause_if is atomic → lost race just retries), turn-cap inflation (the slices_running arm writes no lifecycle event while children merely run), and the total<=0 guard (a real slices_running parent always has children). Minor non-blocking note: add a test for accepted>=total with a concurrently-failed slice.

PR #1547, opened 2026-07-19
fix(wfe): autonomous runs terminate at dead-ends instead of piling up as zombie pending_human

JBailes:
## Problem

Observed live: **22 "active" runs** piled up on the box, several predating the session. An autonomous run that hits a dead-end parks `pending_human` and stays `active` **forever**:
- a node that exhausts its loop cap defaults to `on_max: human` → `pending_human` (e.g. implement `max_iters`, the largest source: 9 runs);
- a `foreach` parent with a failed slice returns `pending_human`.

There is no human to satisfy that wait in autonomous mode, and the stale-park reaper (#1519) **deliberately never reaps a human wait** (legitimate for interactive runs). So they accumulate indefinitely and keep competing for the model roster.

## Fix

For an **autonomous** run, a `pending_human` park at anything other than a real human gate (`gate.human`) is a terminal dead-end → **terminate it (`abandoned`)** with an audit trail instead of parking. Two engine sites:
1. the block-returned pending apply (covers `foreach`-failed and any block that parks `pending_human`);
2. the loop-cap `on_max: human` branch (covers `implement` `max_iters`).

A real `gate.human` still parks (its whole purpose is the operator decision); **interactive runs are unchanged**.

## Tests

`test_wfe_foreach` asserts autonomous failed-slice / no-spawner runs abandon (cases 3, 5/6, 7b, 8), and adds an **interactive** case (6b) that still parks `pending_human`. `test_wfe_autonomy`, `test_wfe_manager_flow`, `test_wfe_sliced_build` pass. `make lint` green.

## Follow-up (noted, not in this PR)

The `wfe_autonomy` ci-poll / panel-retry give-ups also escalate to `pending_human` / linger `panel_degraded` for autonomous runs — a smaller, separate accumulation source in the autonomy driver (not the engine).

### DIFF
--- a/src/tests/test_wfe_foreach.c
+++ b/src/tests/test_wfe_foreach.c
@@ -6,7 +6,8 @@
  *   - no children + spawner returns 0    -> advance (no packets, nothing to do)
  *   - no children + no spawner           -> park (fail closed)
  *   - all children accepted              -> advance (every slice merged)
- *   - a child rejected or abandoned      -> park pending_human (a slice will not merge)
+ *   - a child rejected or abandoned      -> autonomous: abandon (dead-end, no human);
+ *                                           interactive: park pending_human
  */
 #include "wfe_test_home.h"
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


JBailes:
Roundtable review: see verdict; merging on green CI.

PR #1561, opened 2026-07-19
refactor(modules): extract the config subsystem into src/modules/config/

JBailes:
The final and highest-blast-radius module extraction: `config.h` is included by **~230 files**. Moves `config.c` + the `config_*.c/.h` family (22 files) into `src/modules/config/`.

## Moved (15 `.c` + 7 `.h`)
All 15 `config*.c` (CORE_SRCS) + `headers/config.h` and five `config_*.h` + **`config_internal.h`** (which lived at the `src/` root, not `headers/`).

## Why this is safe despite the blast radius
- **No basename collision**: `config.h` is unique tree-wide (unlike the learning.h case), so `-Imodules/config` has **no ordering constraint** in `C_FLAGS`.
- **~230 consumers are all bare** `#include "config.h"` — they resolve via the new `-Imodules/config` unchanged. Only **33 path-qualified** includes are rewritten to bare (28 `headers/config.h` + 5 `headers/config_learning.h`), including `config_plugin.c`'s own now-sibling include.
- Once the headers leave `headers/`, nothing named `config.h` remains on `-Iheaders`, so there's no wrong-resolution trap.

## Build
- `-Imodules/config` on `C_FLAGS`; re-prefix the 15 `CORE_SRCS` entries.
- `src/tests/Rules.mk`: the **~295** `$(OBJDIR)/config*.o` object references, via an anchored `$(OBJDIR)/config → $(OBJDIR)/modules/config/config` (the `config_autonomy_stub`, `server/agent_config`, `git_verify_config`, `server_config`, `seed_config`, `http_config_routes` objects all carry directory prefixes and are untouched — verified).
- `CMakeLists.txt`: the CMake source list (the present subset — a pre-existing gap omits `config_server_api/mode/fields`, left as-is) + the six `target_include_directories` lists; `src/tests/CMakeLists.txt`: two test sources + the test-helper include dir.

## Scripts repointed (2 are silent-failure landmines)
- `scripts/check_tier_deps.sh` — tier allowlist `src/config*.c` / `src/headers/config.h` → `src/modules/config/`.
- `scripts/gen-reference-docs.py` — the config-key doc table reads `config_fields.c` and globs `config*.c`; **regenerated `docs/gen/configuration.md`** (source-path note).
- `src/tests/gen_config_surface.py` — reads `config.c` to generate the surface test.

## Left in place (different subsystems)
`server/agent_config.c`, `server_config.c`, `server_seed_config.c`, `server_http_config_routes.c`, `modules/git/git_verify_config.c`, and the `config_autonomy_stub` test double.

## Verification (local, off `origin/testing` with all prior modules merged)
- `make all` ✅ · `make build-integrity module-boundary-check` ✅ · `make unit-tests` ✅ (validates every one of the ~295 Rules.mk object-path rewrites) · `make lint` ✅ (incl. `check-tier-deps`, `docs-gen-check` against the regenerated doc, and clang-format)


JBailes:
### Review evidence (roundtable — belt-and-suspenders for the load-bearing claims)

**1/4 — no collision (config.h is unique).** `git ls-files | grep '/config\.h$'` → exactly one match: `src/modules/config/config.h`. Nothing named `config.h` remains on `-Iheaders` or anywhere else, so `-Imodules/config` has no ordering constraint and bare `"config.h"` resolves unambiguously.

**3/4 — Rules.mk sed + object-path integrity.** The rewrite was a single anchored expression:
`sed 's#$(OBJDIR)/config#$(OBJDIR)/modules/config/config#g'`
Post-edit: **zero** stray bare `$(OBJDIR)/config*.o` remain, and no config object resolves outside `modules/config/` except the deliberately-retained dir-prefixed non-members (`$(OBJDIR)/tests/support/config_autonomy_stub.o`, `$(OBJDIR)/server/{agent_config,server_config,server_seed_config,server_http_config_routes}.o`) — those carry a directory prefix so the anchor never matched them. `make unit-tests` green is the end-to-end proof that every relinked target found its objects.

**2/4 — glob-scope narrowing is intentional.** `gen-reference-docs.py` now globs `(SRC/"modules"/"config").glob("config*.c")` instead of `SRC.glob("config*.c")` — deliberately scoped to the config module (the only place config-key definitions live now); the regenerated `docs/gen/configuration.md` is byte-identical apart from the source-path note, confirming no key was dropped.

Non-blocking follow-ups noted: (a) backfill the pre-existing CMake omission of `config_server_api/mode/fields`, (b) document the `-Imodules/config` precedence contract in the build README. Both are tracked follow-ups, not part of this mechanical move.

PR #1797, opened 2026-07-22
docs: propose a real readiness endpoint for aimee-server

JBailes:
`route_health()` (`src/server/server_http.c:657-661`) is a string literal. It returns `"ok"` when DB1 is unwritable, aimee-kb is unreachable, and aimee-llm is down. A health endpoint that cannot report unhealth is a false signal, and worse than none because it is trusted.

## What this proposes

- **`/v1/health` stays exactly as it is** — unconditional 200, constant body, meaning "this process is alive." Preserves the documented aimee-kb symmetry and breaks no existing consumer.
- **`/v1/ready` is new** — per-dependency state (DB1, aimee-kb, aimee-llm when configured) with a roll-up, 200 when ready / 503 when not.
- **Served from a snapshot refreshed off the request path**, so the handler does no I/O. Fail closed: unsampled is `unknown`/503, stale snapshots degrade, timeouts report `fail` not `ok`.
- **Behind a provider seam**, because `server_http.c` is *deliberately* dependency-light (`server_http.c:685`, `server_api.c:3-6`) — that isolation is why the route is a constant today, and calling dependency checks directly from it would regress the design.

## Review

Adversarially reviewed (codex + MiniMax-M3, 2 seats, no degradation). The review raised four blocking findings against the original draft and **dropped two of its three slices**. Both kills were accepted:

- *Mark `web_search` results untrusted* — dropped: the draft never traced the downstream path from tool return to prompt assembly, so its premise was unestablished.
- *Probe self-hosted endpoints for capabilities* — dropped: the operator override file already solves it, and no OpenAI-compatible field was named that reliably supplies a context window.

The reasoning for both, plus the deferred MCP OAuth item and a dead-stub cleanup that was correctly called out as padding, is recorded in §7 so it is not relitigated.

The surviving slice was redesigned against the blocking findings: liveness/readiness split, status-code matrix decided, execution model specified, and a corrected red-before-green criterion (the draft's version was unachievable — there is no stock `/v1/ready` to point a test at).

Design only, no code. §2 is the implementation gate. §5 states the assumption that would falsify it.

### DIFF
--- a/docs/proposals/pending/server-readiness-endpoint.md
+++ b/docs/proposals/pending/server-readiness-endpoint.md
@@ -1,0 +1,235 @@
+# Proposal: a real readiness endpoint for aimee-server
+
+- **State:** PENDING — design only, no code in this PR.
+- **Author:** JBailes
+- **Date:** 2026-07-22
+- **Charter roles:** Constrain-Verify (report measured dependency state instead
+  of an unconditional constant), Enforce (fail closed: unknown state is
+  not-ready, never ready).
+- **Review:** adversarially reviewed 2026-07-22 (codex + MiniMax-M3, 2 seats, no
+  degradation). The review dropped two of the three slices in the original
+  draft; see [§7 Rejected](#7-rejected-with-reasons). This document is the
+  surviving slice, redesigned against the review's blocking findings.
+
+## Thesis
+
+`route_health()` (`src/server/server_http.c:657-661`) is:
+
+```c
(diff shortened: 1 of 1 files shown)


JBailes:
Superseded by #1801, which carries this proposal as its first commit alongside the implementation. Keeping them on separate branches guaranteed a conflict, and this docs-only PR went stale mid-CI three times against a fast-moving `testing`.

PR #1856, opened 2026-07-23
fix(docker): build the kb image on Debian 13 for libpq 17

JBailes:
## The break

`aimee-kb` image build is failing on `testing`, which fails `e2e-docker` for every PR that merges testing in.

```
db2/vault_operator_status_runtime.c:168: error: unknown type name 'PGcancelConn'
db2/vault_operator_status_runtime.c:168: error: implicit declaration of function 'PQcancelCreate'
```

That file (from P7-D3a, `d6fba978`) uses the **PostgreSQL 17** async-cancel API. The image builds on `debian:bookworm-slim` → **libpq 15**, where those symbols don't exist.

## Why testing looked green

The native CI `build` job runs `make -C src all server`. **KB_SRCS is in neither target**, so this file is never compiled there — the Dockerfile is the only place it gets built. That's worth knowing for any future slice touching KB-only sources: `build` and `unit-tests` passing says nothing about whether the kb compiles.

## The fix

Both stages → `debian:trixie-slim` (Debian 13, libpq 17). Build and runtime must match, since the runtime `libpq5` has to provide the symbols the kb was linked against.

**Verified by building it, not by reasoning:**

| check | result |
|---|---|
| `docker build -f Dockerfile .` | exit 0 |
| `libpq5` in image | `17.10-0+deb13u1` |
| `docker run ... aimee-kb --version` | `aimee-kb 0.2.0` |

The binary starting is the real proof — dynamic symbols resolve at load, so a missing `PQcancelCreate` would fail there rather than run. Built on a fresh Debian 13 container with Docker 29.6.2.

## Scope

One file, deliberately. `Dockerfile.server`, `.status-authority`, `.embedder` and `.kb-console` stay on bookworm — none compile KB_SRCS (`status-authority-core` doesn't include this file), so none need libpq 17 and bumping them would be unrelated risk.

Found while investigating why CI wouldn't run on #1830.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,10 @@
-FROM debian:bookworm-slim AS build
+# Debian 13 (trixie) for libpq 17. The kb links db2/vault_operator_status_runtime.c,
+# which uses the PostgreSQL 17 async-cancel API (PQcancelCreate / PGcancelConn /
+# PQcancelPoll). Bookworm ships libpq 15, where those symbols do not exist, so the
+# kb build failed here with -Werror=implicit-function-declaration while the native
+# CI build stayed green -- `make all server` does not compile KB_SRCS, so this
+# image was the only place that file was ever built.
+FROM debian:trixie-slim AS build
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


JBailes:
Added a second unbreak-`testing` fix here, since it has the same character as the first: a failure on `testing` that hits every branch.

`unit-tests` went red on an unrelated PR with `test_kb_vault_operator_status.c:158`. It is a race, not a regression. `test_root_socketpair`'s unprivileged path sends a request and asserted the write landed in full — but the behaviour under test is that the server refuses the peer *without reading* and closes, so that send races a close it is meant to lose. The assertion encoded a scheduling order rather than an invariant.

Measured: **0/400 failures idle, 9/200 (4.5%) with 16 spinners pinning the CPU** — which is what a parallel CI test run looks like, so it will keep landing on random PRs.

One race, two faces, which is why the CI log at first named no assertion: CI hit the short/failed write and aborted on the assert; locally it more often hit SIGPIPE and died as a bare exit 141 with no output at all.

The security invariant is untouched and still asserted — the peer is rejected (`result == -1`) and never answered (`read <= 0`). Only the scheduling assumption was relaxed. To check the relaxation did not defang the test, I sabotaged the server to write one byte to a rejected peer: **fails 60/60**. Restored: **0/800** under heavier load than the config that used to fail 4.5%.

PR #1955, opened 2026-07-25
Audit MCP / tool-call activity onto the event bus

JBailes:
Scopes the **last un-audited governed surface** — MCP / tool-call activity — as its own proposal, ahead of implementation.

## What this is

A design proposal (`docs/proposals/pending/mcp-tool-call-audit-on-bus.md`) to record tool-call **identity and outcome** to the audit ledger through `obs_bus`, closing the one gap left after the event-bus observability arc (governed-action audit, guardrail, vault, sandbox, memory server+kb are all on the bus; MCP/tool calls are not).

## Design in brief

- **Fire sites** — the two choke points already in the code: `mcp_client_call_tool` (outbound to external MCP servers) and `dispatch_tool_call`/`dispatch_tool_call_ctx` (aimee's own tools + MCP-gateway-served tools). The existing `agent_tool_event_cb_t(phase, tool_name)` telemetry hook carries no args/principal, so it can't fingerprint and isn't reused.
- **Mechanism** — identical to the other five bridges: a NULL-default `mcp_set_tool_audit_hook` in the MCP module (thin-client safe, no bus dependency — D7) + a server-only `mcp_tool_audit_bridge.c` → `obs_bus_emit`, installed in `server_main.c` next to vault/sandbox/memory.
- **Fingerprint only** — arguments cross as `audit_args_hash` (one-way HMAC), never raw; results are reduced to a `verdict` + bounded `reason_code`, never captured. Tool/server identity is non-secret by construction.
- **Inline** — the row is fixed-size (fingerprint + short identity), rides inline like the other bridges. **No arena payload.**

## Why a proposal first

MCP args/results are the highest-risk secret/PII surface on the bus, so the secret-leak discipline deserves the same up-front scrutiny (and multi-agent convergence review) the other five bridges got before code.

## Explicit non-goal (important)

Full-content capture/replay of tool I/O — recording the actual arguments and results — is a **separate, opt-in, egress-gated feature** with its own threat model and redaction requirements. It would be the first real consumer of the arena payload path (args/results can exceed the inline budget — see #1954), but it is deliberately kept **distinct** from this content-free audit and is out of scope here.

## Binding checks named in the proposal

`test_bus_mcp_audit` (mirrors the sandbox/vault bridge tests) + a no-secret-leak assertion (a sentinel in the arguments must appear nowhere in the row/ledger — only its hash), and `check_bus_blast_radius.sh` stays green.


JBailes:
## Roundtable review → revised (needs-revision, unanimous)

Ran a five-lens roundtable (secret-leak · architecture · coverage · API-feasibility · adversarial), each seat verifying claims against the code. All five returned **needs-revision**; four blockers, all code-grounded. The proposal has been rewritten to incorporate every finding.

**What the review caught (and the revision fixes):**

1. **The "argument fingerprint" was false.** `audit_args_hash` hashes the tool *name only* off an 8-tool allowlist (`audit_action.c`: *"A tool absent here hashes its NAME ONLY"*) — so it's name-only for every MCP tool, and the original no-leak test would have passed **vacuously** (args never hashed). Claim dropped; the hash is now honestly identity-only.
2. **The gap was smaller than claimed.** `emit_action_audit` (`guardrails_action_audit.c:138`, fired from the dispatcher) already records tool-call **identity** on the bus for internal/served/external calls. The genuine delta is post-execution **outcome + transport** — the proposal is rescoped to a single *completion* emit.
3. **The two fire sites nested** (dispatch → registry → `mcp_client`), double/triple-auditing external calls with divergent hashes → collapsed to **one dispatch-layer emit** (`agent_tools_dispatch.c` `dispatch_tool_call_ctx`, corrected from `server_compute_async.c`), which also fixes the missing-principal problem (`session_id()` lives there).
4. **The "served" surface** (`handle_mcp_call`, `server_mcp.c:1631`) bypassed both original sites → now a must-resolve.
5. **`err_buf` was an unfenced content surface** (MCP server error text echoes arg values, persists to the `0600` `.aimeecap` file) → explicit non-forwarding rule + a dedicated leak test.
6. Fixed file locations and a transport (`outbound:http`) that doesn't exist; strengthened the leak test to cover args **+ result + server error**.

**Kept (verified correct):** the fixed-size INLINE / no-arena claim and the D7 NULL-default-hook story.

The revised proposal is pushed. It's now scoped as "record tool-call **outcome**," which is the accurate, narrow, defensible gap. (Also rebased onto the merged #1954.)

JBailes:
## Implementation roundtable → fixed (needs-fixes, now addressed)

Ran a five-lens roundtable on the **implementation** (concurrency · security/leak · correctness · MCP-specific · regression), each seat verifying against the code, then an adversarial-verify pass on every blocker/major, then a chair synthesis. It **verified clean**: concurrency (thread-partitioned outcome carriers, no live race), memory safety, the transport accessor's locked resolution, and D7/link blast-radius. It found **three confirmed majors**, all now fixed:

1. **Security — control bytes in served identity.** An external client's served tool name / `session_id` carried raw newline/ANSI bytes into the `0600` capture file, printed unescaped by `--audit-replay` (forgeable rows / terminal-escape injection). Fixed at the one serializer feeding both sinks — `obs_bus.c put_str` now maps bytes `<0x20`/`0x7f` to `?` (a no-op for every legitimate printable value); replay scrubs too (defense in depth).
2. **Correctness — exec verdicts.** `bash`/`execute_script`/`test`/`run_tests` return a success-shaped `{stdout,stderr,exit_code}` envelope even on failure, so the `error:`-prefix heuristic logged every non-zero exit (the highest-traffic tools) and the fail-closed sandbox refusal as `ok`. Now classified from the real signal (`exit_code`/`status`/stderr `refused:` prefix — never storing content). The generic prefix path is also anchored so caller text can't flip the verdict.
3. **Correctness — served admission refusals.** Delegate capability/persona refusals sent their error frame then fell through to the wrapper's default `ok`. Added a public same-thread `server_mcp_served_outcome` setter, called at those refusals + the served OOM/ensemble returns.

Plus: the previously-tautological no-leak test now drives a control-byte-laden tool name and asserts the row is scrubbed; dead `snprintf` removed; reentrancy invariant / install-once ordering / `outbound:stdio|sse` documented.

**Verified:** all 7 binaries build clean · D7 green · perf gate green · audit-replay + durability + all bus bridge tests pass · `mcp-native-dispatch`/`server-dispatch`/`cli-mcp-serve` pass · clang-format clean. Also validated on a clean `.253` CT: the full 627-test `unit-tests` suite passed (the two red items there were Go-env gaps — `aimee-runtime-web`'s `go build` + Go conformance — not this code).

JBailes:
Brought up to date with `testing` (now includes #1957, the MCP adapter). The two features overlap in `agent_tools_dispatch.c` (this PR's completion-hook + transport-mode audit vs. #1957's local-vs-kb routing), `mcp_client_registry.{c,h}`, `server_main.c`, and the Makefile — git auto-merged them and I verified the combined `:` dispatch branch is coherent (transport-mode audit + local/kb routing + error classification all present). Full build clean; bus-tool-completion, mcp registry/dispatch, and config-golden tests pass; docs-gen-check ok; **blast-radius gate clean across all 7 shipping binaries**.

PR #1957, opened 2026-07-25
MCP adapter: optional module installable into aimee-server or aimee-kb (Phase 1)

JBailes:
Scopes making the **MCP client adapter an optional module either `aimee-server` or `aimee-kb` can install**, with the install target deciding exposure scope — a design proposal, ahead of implementation.

## The requirement
- **Installed in `aimee-server`** → external MCP tools exposed only to that server's sessions (today's behavior).
- **Installed in `aimee-kb`** → exposed to *everything hooked up to that kb* (every server + thin client), backed by one shared set of connections the kb owns. A single operator-configured MCP server becomes a deployment-wide capability, configured/connected once.

## Design leans on two existing seams
1. `modules/protocols/mcp` is already a self-contained module (clean boot/list/call surface) — nothing binds it to the server.
2. A **kb→server tool-registry federation already exists** (`kb_client_tool_registry_snapshot_json` / `_lookup`) — the exact channel a kb-hosted MCP surface rides.

## Crux — where a kb-installed tool executes
"Shared to everything on the kb" means the connection is **owned by the kb**, so a kb-installed call runs *at* the kb: the kb publishes its namespaced tools through the registry snapshot; servers merge them into the session tool list; the dispatcher's `server:tool` branch **forwards** a kb-hosted name over `kb_client` to a new `kb.mcp.call` method rather than calling locally. Server-installed tools keep executing locally.

## Audit follows the install target (composes with #1955)
A server-installed call audits its outcome on the **server's** `obs_bus` (as it does now, mode `outbound:stdio|sse`); a kb-installed call executes at the kb and records its outcome on the **kb's** `obs_bus` (the same NULL-default hook + server-only bridge, installed in `kb_main.c`, mirroring the memory-kb bridge). Content-free either way. This is why it's scoped **after** #1955 — it reuses that PR's completion hook + bridge on the kb side.

## Non-goals
No new egress surface (the kb already makes operator-configured outbound connections); no per-session/tenant MCP; no MCP-wire/schema change; **D7 unchanged** (bus stays in the two trusted daemons).

Binding checks (scope boundary, forwarding+audit split, collision-refused-at-boot, D7 green) and a 5-step rollout are in the doc.

### DIFF
--- a/docs/gen/configuration.md
+++ b/docs/gen/configuration.md
@@ -265,7 +265,7 @@ Set in the config JSON as `{"<section>": {"<key>": ...}}`. Keys are derived from
 - **`learning`** — _Learning subsystem (router, implicit, embed, synthesize; nested objects)._ Keys: `embed`, `implicit`, `review`, `router`, `synthesize`
 - **`lsp_servers`** — _LSP server definitions (array of objects)._ Keys: `args`, `command`, `extensions`, `name`
 - **`mcp`** — _MCP integration (e.g. OSV)._ Keys: `osv`
-- **`mcp_clients`** — _MCP client connections (array of objects)._ Keys: `bearer_token_env`, `command`, `cwd`, `name`, `transport`, `url`
+- **`mcp_clients`** — _MCP client connections (array of objects)._ Keys: `bearer_token_env`, `command`, `cwd`, `install`, `name`, `transport`, `url`
 - **`memory`** — _Memory subsystem; most children (recall, rerank, lifecycle, …) are nested objects with their own keys._ Keys: `abstain`, `aggregation`, `bm25_weight`, `briefing`, `citations`, `cognify`, `context_budget`, `coref`, `derive_facts`, `directives`, `dispositions`, `episode_summaries`, `failure_detection`, `fetch_budget`, `hard_negative_log`, `improve`, `lifecycle`, `pagerank`, `profile_cards`, `prospective`, `recall`, `rewrite`, `routing`, `salience`, `scenes`, `semantic_floor_scale`, `semantic_weight`
 - **`memory_maintenance`** — _Memory maintenance scheduling._ Keys: `enabled`, `interval_seconds`, `summarize_enabled`, `trigger_inserts`, `trigger_secs`
 - **`memory_negation`** — _Negation handling in memory._ Keys: `enabled`
--- a/docs/proposals/pending/mcp-adapter-optional-module.md
+++ b/docs/proposals/pending/mcp-adapter-optional-module.md
@@ -1,144 +1,148 @@
-# Proposal: an event-bus MCP-plugin adapter — optional, install into aimee-server or aimee-kb
+# Proposal: MCP-plugin adapter as an optional module (install into aimee-server or aimee-kb)
 
-## The idea
+> **Revised after a five-lens design roundtable (2026-07-25), verdict needs-rework →
+> reworked.** The first draft framed this as "an adapter *into the event bus* that
+> runs MCP plugins," with plugin calls "riding the bus" across daemons. The
+> roundtable verified against the code that **the event bus is intra-daemon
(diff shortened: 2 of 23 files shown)


JBailes:
## Reframed per clarified intent

Corrected the proposal: this is **an adapter *into the event bus* that runs MCP plugins**, with the **thin-client advertising those plugin tools over MCP/CLI exactly as it does for modules today** — not merely "the MCP client registry made installable in server/kb."

The shift matters because it ties the whole bus arc together:
- A plugin tool-call travels the bus as a **request event**; the adapter serves it against the plugin's MCP session and publishes the **reply** — the first non-test consumer of the correlated-arena request/reply routing.
- Large args/results ride the **arena by reference** — the MCP-plugin adapter is the **first real producer** of the arena payload path the earlier PRs built.
- **Scope by construction:** the bus the call rides *is* its blast radius — server bus ⇒ that server's sessions, kb bus ⇒ everything on the kb. Not a separate ACL.
- **Isolation/backpressure/audit for free:** the plugin's connection lives behind the adapter; a slow/dead plugin is BLOCK/SHED-managed, never a caller hang; the call is audited content-free on whichever daemon's bus it runs (#1955 hook/bridge, kb side mirrors the memory-kb bridge).
- **Advertisement:** plugin tools feed the existing `tools/list` surface (+ the kb→server tool-registry federation for kb-hosted plugins) — no new advertisement mechanism, same as modules.

MCP *is* the plugin interface; nothing revives the removed in-process `dlopen` loader. D7 unchanged.

JBailes:
## Design roundtable → reworked (needs-rework, addressed)

Five-lens roundtable (architecture · security · feasibility · completeness · skeptic) + adversarial verification. It rejected the alarming security claims on verification (content-free audit is structurally enforced on the separate audit host; scope-by-daemon is real), but two blockers survived that invalidated the premise:

1. **The event bus is intra-daemon shared memory** (memfd regions over a local socketpair/SCM_RIGHTS; each daemon runs its own host — D1/D7). It never crosses the server↔kb boundary; `kb_client` is HTTP/mTLS with zero bus references. So a server **cannot** "publish a plugin request onto the kb's bus" — the real cross-daemon path is HTTP.
2. **No general bus host exists** to be a participant on — only obs_bus's private 4-slot notification-only audit singleton. A general host would be net-new infrastructure, and a bus participant in the MCP module **fails** `check_bus_blast_radius.sh` by design (a governed D7 amendment, not "stays green").

**Reworked** into a buildable split:
- **Phase 1 (now):** the adapter as an optional module (server/kb install target); server-install = direct in-process call (today); kb-install = servers call over a **new `/v1/mcp/call` HTTP action** + kb tool **definitions** federated via an **extended** tool-registry snapshot RPC (today's carries only prompts) + a server-side `tools/list` merge. Audit reuses the #1955 hook/bridge on the hosting daemon's obs_bus. **D7 unchanged.**
- **Phase 2 (deferred, opt-in):** routing plugin calls over a daemon's *own* bus (isolation/backpressure/arena-by-reference) — requires a general bus host, the client-visible arena request/reply surface, and a **governed D7 blast-radius amendment**.

This makes Phase 1 the implementable core that delivers the functional goals (optional module, server-vs-kb scope, thin-client advertisement, bus audit) without the large new-infra + D7 governance investment; the "plugin calls travel the bus" vision is isolated to Phase 2.

JBailes:
Updated: this branch is now the full Phase 1 **implementation** (not just the proposal) — install-target config + boot filtering, aimee-kb hosting of `install: kb` plugins, the `mcp.call` action-RPC, tool-def federation into the server's `tools/list` + local-precedence dispatch routing, and the kb-side content-free tool-call audit. Merged `testing` in (clean), regenerated `docs/gen` (lint fix), all affected + config-golden tests pass. Extensive e2e on CT 132: kb hosting/execution/federation/audit, live-model agent turn, error/timeout/concurrency, and live local-precedence collision. See the proposal doc's verification sections for observed evidence.

PR #2007, opened 2026-07-26
Add frozen Gemma 4 baseline fixtures

JBailes:
## What changed

- Freeze a de-identified `.254`-derived corpus and three paired 10,000-case views for embedding, reranking, and synthesis.
- Require both Ettin-68M CPU and Ettin-400M GPU reranker controls over the identical frozen candidate lists.
- Run both Ettin controls under a pinned concurrent profile: eight request workers, four pairs per request, 32 llama.cpp slots, and 2,048-token logical/physical batches; failed rows are retried on resume.
- Record immutable file hashes, shared case IDs, strata, model-view requirements, and training-exclusion rules in the suite manifest.
- Add deterministic fixture generation, fail-closed validation, role-specific runners, pairwise bootstrap reports, model download support, and the `.254` sweep controller.
- Add a separate matched EuroBERT-210M/610M cross-encoder extension: official base revisions, identical pinned 576,000-example teacher-score training exposure, a pinned PyTorch/ROCm runtime, resumable training, and the same 10,000-case reranking/load contract.
- Add contract tests for bundle integrity, reranker bounds, secret scanning, resume behavior, EuroBERT provenance drift, and frozen-manifest isolation.

## Why

These test sets are expensive and non-trivial to reproduce. Committing a single frozen suite gives us a durable pre-training and post-training A:B reference for E2B, E4B, Gemma 4 12B/26B/31B, Qwen 3.6, and both supported Ettin reranker controls.

The official EuroBERT checkpoints are masked-language encoder bases rather than ready-made rerankers. The separate matched training extension makes EuroBERT-210M/610M measurable against Ettin without changing the already-frozen suite hash or contaminating evaluation with training rows.

## Impact

No runtime defaults change. This PR adds benchmark-only fixtures and tooling. Every model is evaluated on the same case IDs, while embeddings use each model's required native width up to AIMEE's 4,096-dimension cap. EuroBERT/Ettin reports must retain the documented training-budget caveat: the bounded EuroBERT extension uses 576,000 examples, while released Ettin used the full 143,393,475-example dataset.

## Validation

- `python3 benchmarks/gemma4_baseline/validate_fixtures.py`
- `python3 -m unittest benchmarks.tests.test_gemma4_baseline_contract -v`
- `git diff --check origin/testing...HEAD`


JBailes:
Published completed result checkpoint in `3fa790d8`.

- Gemma 4 E4B embedding: 10,000/10,000 frozen cases, native width 2,560, MRR@10 0.328657, nDCG@10 0.422186, Recall@1/5/10 0.1832/0.5205/0.7275, 2.2839 vectors/s.
- Gemma 4 12B synthesis: 10,000/10,000 successful final cases, raw JSON parse 100%, schema-valid 99.96%, required-field recall 97.57%, content F1 0.327918, p50/p95 42.39/91.84 s, decode 2.4084 tok/s.

Validation:

- E4B raw file has exactly the frozen 10,000 unique case IDs; recomputed metric means reproduce every summary value.
- 12B append log has 10,013 rows; latest-per-case reduction yields exactly 10,000 successful frozen cases and reproduces the summary exactly.
- Both raw files pass the fail-closed secret scan.
- `python3 -m unittest benchmarks.tests.test_gemma4_baseline_contract`: 19 tests passed (2 dependency skips).

Not published:

- E4B synthesis still has four failed final cases.
- 12B embedding remains live and resumable, so its result files are intentionally excluded until complete.

Issue #2020, opened 2026-07-27
Published aimee-server:latest is 13 days stale — clean-install Deploy fails with docker.sock EACCES


JBailes:
Found while validating the new-user path on a clean host.

## Impact

Every new user following `docs/QUICKSTART.md` hits this. The wizard's **Deploy**
step fails immediately with:

```
permission denied while trying to connect to the Docker daemon socket
at unix:///var/run/docker.sock
```

## Cause

`ghcr.io/rakuensoftware/aimee-server:latest` was built **2026-07-14** and its
entrypoint is 123 lines. The repo's `deploy/container/server-entrypoint.sh` is
216 lines and contains the block that fixes exactly this:

```
delegate sandbox: granted 'aimee' the docker socket group (dockerhost/<gid>)
```

It stats the mounted socket, creates a group with the socket's GID when the image
has none, and adds `aimee` to it. Without it, `aimee-server` runs as uid 1000 with
`groups=1000` only, while the host socket is `root:docker 660` — so it cannot open
the socket.

| | Published `:latest` | Repo `testing` |
|---|---|---|
| Built | 2026-07-14 (13 days) | today |
| Entrypoint | 123 lines | 216 lines |
| Socket group grant | **absent** | present |

## Verification

On a fresh CT (Debian 13, Docker 29.6.2), building the image from the current
tree and recreating the container:

```
[server-entrypoint] delegate sandbox: granted 'aimee' the docker socket group (dockerhost/991)
uid=1000(aimee) gid=1000(aimee) groups=1000(aimee),991(dockerhost)
```

and the deploy proceeds instead of failing.

## Remedy

**Republish the image.** No code change is needed — the fix is already committed.
Worth checking why `:latest` drifted 13 days behind `testing`, since the same gap
will hide the next fix too.

## Related

With the socket fixed, Deploy then hit a second bug — it removed the server as a
compose orphan and killed itself. Fixed in #2019.

JBailes:
## This is broader than the server image, and it produces a second independent blocker

Continued the clean-install validation past the socket fix. **Every published image is stale:**

| Image | Built | Age |
|---|---|---|
| `aimee-llm-cpu:latest` | 2026-07-01 | **~4 weeks** |
| `aimee-kb:latest` | 2026-07-14 | 13 days |
| `aimee-server:latest` | 2026-07-14 | 13 days |

### Second blocker: aimee-kb can never become healthy

[...]

PR #2033, opened 2026-07-27
The cpu synth model filename 404s, which blocks every release

JBailes:
One line. Unblocks all image publishing.

## What is broken

`ggml-org/gemma-4-E4B-it-GGUF` does not contain `gemma-4-E4B-it-Q4_K_M.gguf`. The repo
is fine (200); that quantization simply is not published in it — the file list is
BF16 / Q4_0 / Q8_0 (plus mmproj and mtp variants). So the request 404s:

```
aimee-llm: fetching synth tier 'cpu' into /models/cpu
aimee-llm: download FAILED: .../gemma-4-E4B-it-Q4_K_M.gguf
aimee-llm: pre-bake download failed
```

## Why it blocks far more than one model

`aimee-llm-cpu` bakes models at build time, so the 404 fails that image on both amd64
and arm64 → fails the `images` job → fails **auto-release**. auto-release is what
publishes `aimee-server` and `aimee-kb` as well, so one dead URL has stopped **all**
image publishing since **2026-07-14**.

Consequence, measured on a clean install today (#2020): every published image is 13 days
to 4 weeks stale, and two new-user blockers turned out to be fixes that already existed
in the repo and had simply never shipped.

## Why it lands on main directly

`testing` already carries the corrected filename — but releases only fire on a push to
`main`, and `main` is where the dead URL lives. The fix for the broken pipeline could
not itself be published. This breaks that deadlock with the smallest possible change.

## Verification

Both cpu-tier URLs checked before proposing:

```
200  Qwen/Qwen3-Embedding-0.6B-GGUF/.../Qwen3-Embedding-0.6B-f16.gguf
200  ggml-org/gemma-4-E4B-it-GGUF/.../gemma-4-E4B-it-Q4_0.gguf
```

Deliberately minimal — a new LLM is landing soon, so this only unblocks publishing rather
than revisiting tier selection. Merging cuts a release from main's current code; promoting
`testing` (647 commits, including today's eight fixes) remains a separate decision.

### DIFF
--- a/scripts/aimee-llm-supervisor.sh
+++ b/scripts/aimee-llm-supervisor.sh
@@ -73,7 +73,7 @@ rerank_coords() {
 synth_coords() {
   case "$SYNTH_TIER" in
     cpu)
-      SYNTH_REPO="ggml-org/gemma-4-E4B-it-GGUF"; SYNTH_FILE="gemma-4-E4B-it-Q4_K_M.gguf"
+      SYNTH_REPO="ggml-org/gemma-4-E4B-it-GGUF"; SYNTH_FILE="gemma-4-E4B-it-Q4_0.gguf"
       SYNTH_REVISION="main"; SYNTH_SHA256=""
       TIER_FA="off"; TIER_MOE="0"; TIER_N_CPU_MOE="0"; TIER_SLOTS="1"; TIER_CTX="32768"; SYNTH_NGL=0
       ;;


JBailes:
Closing: this route is blocked by design.

`branch-policy.yml` produces the required check `allow-only-testing-source`, which fails
any PR into `main` whose head branch is not `testing`:

```
::error::PRs into 'main' may only originate from the 'testing' branch (source was 'fix/release-dead-model-url')
```

So a targeted one-line fix cannot reach `main` at all — the only sanctioned path is a
`testing` -> `main` promote. The fix itself is already on `testing`, so no code change is
needed; it just needs promoting.

Leaving the diagnosis here for the record: the dead `gemma-4-E4B-it-Q4_K_M.gguf` URL on
`main` fails the aimee-llm-cpu bake, which fails auto-release, which is why **no** images
(server, kb, llm) have published since 2026-07-14.

PR #2041, opened 2026-07-27
EXPERIMENT: declare reasoning intent on the codex responses request

JBailes:
Testing the hypothesis from the codex diagnosis. Not for merge until proven on a live call — opened only so CI builds a deployable image I can test with.

Measured: codex returns 2 reasoning output items, no message, no function_call, 0 text deltas, 47/71 output tokens on reasoning. The request carries no `reasoning` field.

This adds `reasoning: {effort, summary}` and `include: [reasoning.encrypted_content]` (needed because `store: false` means the server holds no state across turns).

If the probe still returns reasoning-only, the hypothesis is wrong and this gets closed.


JBailes:
Superseded. The reasoning parameter was not the cause — the streamed function_call was being discarded. Real fix and proof in the replacement PR.

PR #2073, opened 2026-07-27
Harden KB ingest and DB connection exhaustion for 0.3.0

JBailes:
## Summary
- sanitize malformed UTF-8 before KB text reaches strict PostgreSQL boundaries
- roll back failed file transactions immediately instead of continuing an aborted transaction
- never share the init thread PGconn when the pool and overflow connection are unavailable
- repair dedicated per-thread connection TLS ownership
- make ingest queue completion report write failures

## Evidence
- reproduced repeated invalid UTF-8 PostgreSQL failures on the release target
- inspected an aimee-kb SIGABRT core whose corrupt libpq/malloc stack is consistent with concurrent PGconn sharing
- added focused malformed-input and pool-exhaustion regression tests
- make -C src lint
- make -C src -j8 unit-tests

This is release-blocking hardening discovered during 0.3.0 exploratory validation.

### DIFF
--- a/src/cli_mcp_serve.c
+++ b/src/cli_mcp_serve.c
@@ -388,9 +388,13 @@ static void handle_tools_call(cJSON *id, cJSON *req)
 
    /* Check server response status */
    cJSON *status = cJSON_GetObjectItemCaseSensitive(resp, "status");
-   if (status && cJSON_IsString(status) && strcmp(status->valuestring, "error") == 0)
+   cJSON *error = cJSON_GetObjectItemCaseSensitive(resp, "error");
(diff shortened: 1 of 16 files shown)


JBailes:
Exploratory target validation found and fixed two additional release issues in follow-up commits:

- curator provider calls now use one bounded five-minute attempt and release lazy DB leases before slow LLM work; the durable job queue remains the retry owner
- nested HTTP authorization errors from remote MCP tool calls are now surfaced as MCP `isError` results with their actionable server message

Focused curator, pipeline, memory, and stdio MCP regression tests pass locally.

PR #2084, opened 2026-07-28
Unit tests leaked a temp directory per run until tmpfs ran out of inodes

JBailes:
## What

**64 of the 112** tests that call `platform_mkdtemp()` never remove the directory they create. `platform_tmpdir()` returned a hardcoded `"/tmp"`, so all of them landed there and survived the run — including the `unit-tests` target's own `trap 'rm -rf "$th"' EXIT`, which only ever covered `HOME`.

They accumulate across every run, by every session, forever.

## Measured on a development box

```
/tmp   1048576 inodes   1048576 used   0 free   (100%)
/tmp   63G size         19G used      45G free   (29%)
```

**40k leaked directories holding 857,343 inodes — 82% of the filesystem — with 45 GB of space still free.**

Once the inode table is full, nothing on the machine can create a file. Builds fail `ENOSPC` on a disk that is 71% empty, and the error names *space*, which is not the resource that ran out. That misdirection cost real time before I looked at `df -i`.

## Fix

Two changes, neither touching the 64 tests:

- `platform_tmpdir()` honours `TMPDIR` when it's an absolute path, falling back to `/tmp` exactly as before;
- the `unit-tests` target exports `TMPDIR` alongside `HOME`, pointing at the throwaway directory it **already** removes on `EXIT`.

The directories are then created inside the run's own temp home and go with it.

Fixing the 64 individually would have been 64 chances to miss one — and the next test to forget would start the pile again.

## Verified

With a known leaker (`test_agent_list_handler`, 9 directories per run):

| | escaped to `/tmp` | landed in `TMPDIR` |
|---|---|---|
| with the fix | **0** | **9** |
| hardcoded path restored | **9** | **0** |

Exact inversion, both directions — and note the second row had `TMPDIR` set and ignored it.


JBailes:
The current unit-tests failure is deterministic: commit 2b6b0474 accidentally dropped `$(OBJDIR)/server/server_error_kind.o` from the `unit-test-server-dispatch` prerequisites while touching the adjacent rule. The linker now fails with `undefined reference to server_send_error_kind`. Restoring that prerequisite should resolve this gate.

PR #2086, opened 2026-07-28
fix: keep selected providers visible and preserve TLS smoke endpoints

JBailes:
## What

`provider list` now treats the provider selected by `provider set` as configured, even when it has no credential environment variable.

Without this, selecting a local provider such as `ollama` immediately made the default list claim that no provider was configured. A selected cloud provider was similarly hidden until its key was supplied. Credentials still count as configuration, and untouched no-key local providers remain hidden by default.

Also adds focused coverage that `--all` is marshalled through the v1 client.

## Verification

- full thin-client build and link
- `cmd_provider.c` and `server_provider.c` compile with warnings-as-errors
- `unit-test-cli-v1-delegate`: all tests passed
- diff check clean

### DIFF
--- a/docs/gen/cli-commands.md
+++ b/docs/gen/cli-commands.md
@@ -566,7 +566,7 @@ Model provider profiles and catalogs.
 Subcommands:
 
 ```
-  list             List registered providers (--available, --json)
+  list             List configured providers (--all, --available, --json)
   show <name>      Show provider profile details
   models <name>    Fetch provider model catalog (--json)
   test <name>      Probe provider credentials and connectivity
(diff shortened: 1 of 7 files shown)


JBailes:
Exploratory release testing also caught and fixed an HTTPS smoke regression in 5a36ec8d: the cross-platform smoke converted every SERVER_URL to a legacy tcp: endpoint, and POSIX gives that endpoint precedence, so TLS deployments were falsely reported unreachable. It now maps https:// to tls: and includes HTTP+HTTPS build-integrity regression cases. The corrected smoke passes end-to-end against the live TLS candidate (version, session list, memory search). Fresh CI is running on the combined head.

PR #2088, opened 2026-07-28
Every sandboxed run leaked an empty /tmp root, forever

JBailes:
## What

`setup_mount_ns()` does `mkdtemp("/tmp/aimee-sandbox-XXXXXX")` in the **host** namespace, then mounts a tmpfs over it inside the child's new mount namespace. When the sandboxed process exits, the namespace goes with it and the tmpfs unmounts — **but the host-side directory stays.**

Nothing removes it, and not by oversight:

- the path is created in the **child** after `fork`, so the parent never learns it;
- `sandbox_spawn()` returns the pid **without waiting**, so no point in this module knows the child has exited.

Fixing it at source means hoisting the `mkdtemp` above the fork and removing it wherever each caller reaps — a refactor across every call site of the sandbox fork path. Not a change to make casually.

## Impact

One empty directory per sandboxed run, accumulating for the life of the machine. A development box had **477**, part of a `/tmp` that had run out of **inodes**:

```
857k inodes across 40k directories — with 45GB of space still free
```

At that point nothing on the machine could create a file, and builds failed `ENOSPC` on a disk that was 71% empty.

## Fix

Sweep leaked roots before creating a new one. **`rmdir` is safe by construction** here — no age heuristic, no ownership check, no PID probing:

| state | `rmdir` result |
|---|---|
| in use (tmpfs mounted) | `EBUSY` — skipped |
| non-empty | `ENOTEMPTY` — skipped |
| leaked (unmounted, empty) | removed |

Only the leaked set can go, so the sweep **cannot disturb a running sandbox** — including one owned by another process or another user. Errors are ignored precisely because that's the mechanism.

## Verified against the real syscalls

- empty leaked root → removed
- non-empty root → survives
- unrelated `/tmp/aimee-other-*` → untouched
- root with a live tmpfs → `rmdir: Device or resource busy`

`make lint` clean; `unit-test-sandbox` and `unit-test-sandbox-learned` pass.

## Scope

This does **not** fix the other in-tree `/tmp` users — kb ingest (`aimee_ingest_`), pdf render (`aimee_pdf_`), msync index, tts. Those are separate call sites with separate lifetimes and deserve looking at on their own terms rather than a blanket sweep.


JBailes:
There is a concurrent-start race in this implementation. `setup_mount_ns()` runs in each forked child:

1. child A reaps, then `mkdtemp`s root A;
2. before A mounts tmpfs, child B enters `reap_stale_sandbox_roots()`;
3. B sees root A as empty and not yet mounted, so `rmdir` succeeds;
4. A then fails its mount/chroot path with ENOENT.

`EBUSY` protects already-mounted roots, but not the pre-mount window, so the current “safe by construction” assertion does not hold under concurrent sandbox starts. Please serialize reap + mkdtemp + successful tmpfs mount with a same-user interprocess lock (for example `flock` on a per-uid lock file), releasing only after the mount is established; add a regression test that starts two setup paths with the first paused after `mkdtemp`.

JBailes:
Fixed the concurrent pre-mount race in 6757d6f6: sandbox processes now take a shared flock on the /tmp directory inode across stale-root reaping, mkdtemp, and successful tmpfs mount. This avoids a separate lock-file lifecycle while ensuring no sibling process can reap another sandbox root before it becomes busy. Focused sandbox and sandbox-learned unit targets pass locally.

PR #2089, opened 2026-07-28
A memory could be written and never unwritten

JBailes:
## What

The memory surface is `search, recall, store, identity, prefer, archive, list, get, show, read, stats, benchmark`. There's no removal — and `archive` turns out to be a **store** (`archive <key> <value>`), not a deletion.

So a memory written by mistake — a secret, a typo, a test fixture written against a live deployment — was permanent from every interface a user has.

## It was never a deliberate immutability guarantee

The entire chain already existed below the server:

```
db2          memory_delete(id)
kb service   db2_kb_service_memory_delete_json()
kb dispatch  {"memory.delete", kb_handle_memory_delete}
kb_client    kb_client_memory_delete(id)
server       -- MISSING --
CLI          -- MISSING --
```

Only the last two hops were absent. The knowledge tier has been able to delete memories all along, with no way for anyone to ask it to.

## This PR

Wires up the missing hops: `handle_memory_delete`, the dispatch entry, `CAP_MEMORY_WRITE` (the same gate as `memory.store`, so it follows existing write-tier grant rules rather than inventing its own), `/v1/memory/delete`, and `aimee memory delete <id>`.

Errors are typed — `invalid_argument` (400) for a missing/non-positive id, `not_found` (404) when the kb won't delete it — rather than the blanket 502 those would have produced before #2079.

## How I found it

I polluted a live knowledge base with 14 test memories during an e2e sweep and could not remove them through any supported interface. Deleting them straight out of Postgres behind the API would have been a worse mistake than making them — the missing operation is the actual defect.

## Verification

- full `make unit-tests`: **0 FAILED, "All tests passed"**
- `make lint` clean, `docs-gen-check: ok` with the descriptor regenerated
- `test_server_dispatch` gains the stub its table requires — added *proactively* this time, having been caught by that exact linkage three times in this batch

## Deliberately not included

**Bulk or predicate deletion.** One id at a time is the whole surface. A "delete everything matching" on a memory store deserves its own design — blast radius, audit, confirmation — not an afterthought in the change that introduces removal at all.


JBailes:
## Added `memory.supersede` alongside `delete`

Adding delete on its own would have made the **wrong tool the easy one**.

`aimee memory supersede <old_id> <new_content>` already existed and worked **on the server host** (`cmd_memory.c` → `mem_supersede` → `kb_client_memory_supersede`), and the kb has dispatched `memory.supersede` all along. The missing piece was the `/v1` hop — so the thin client, which routes everything through `/v1`, couldn't reach it.

A remote user who knew a memory was wrong had two options: store another one, or delete it.

## Why that ordering matters

The store **depends** on the supersession chain. `memory.list_superseded_keys` and `memory.fact_history` walk it. Delete a wrong memory instead of superseding it and you lose the answer to *"why did it assert this in March"*, and destroy the negative examples `effectiveness` and `evidence_strength` are computed from.

A corrected memory is signal. A deleted one is a hole.

## The division the schema already encodes, now reachable

| operation | when |
|---|---|
| **supersede** | a belief replaced by a better one — keeps the chain |
| **archive** | `active → pending (ttl) → archived`, reversible; for machine-judged staleness |
| **delete** | the *content itself* is the liability — a secret, PII, junk written to a live store |

That last row is why the only automated hard delete in the tree (`delete_by_sensitivity`) keys on **sensitivity + age**, never on correctness.

## Verified

Thin client now lists both:

```
search, recall, store, identity, prefer, archive, list, get, delete, supersede, show, read, stats, benchmark
```

Full `make unit-tests` passes (0 FAILED), `docs-gen-check: ok`, descriptor regenerated. The dispatch-test stub was added **with** the handler this time rather than after CI found it.

PR #2092, opened 2026-07-28
surface failing curator jobs in kb health

JBailes:
## The gap

A curator job that exhausts its attempts goes to `status='failed'`. `kb_curator_queue_counts()` summed only `pending` and `done` — so failed jobs were counted in **neither bucket** and disappeared from the health payload completely.

A curator failing every single job was therefore indistinguishable from an idle one:

```
pending: 0   done: 0   warnings: []      -> "aimee-kb: ok"
```

This is the mechanism behind #2070. 41,485 code-unit jobs were failing against a hardcoded LAN endpoint with `attempts=3, last_error="No route to host"`, and `aimee status` printed `aimee-kb: ok` throughout while code indexing was entirely dead. The signal was in the table the whole time — `attempts` and `last_error` are populated on every failure — it just was not read.

## Change

- `kb_curator_queue_counts_t` gains `extract_failing` / `code_unit_failing`, plus one sample `last_error`.
- Both counting queries add `SUM(CASE WHEN last_error <> '' ...)`; the code-unit query also returns `MAX(last_error)`.
- The health `curator.queue` object exposes the counts and the sample error.
- **`warnings[]` gets an entry when anything is failing** — the part that actually breaks the silence, since `warnings[]` is what surfaces to the operator.

The sample error is included deliberately: a bare count does not tell you whether to fix a config value or a network. The message that mattered here named an address nobody recognised.

Counting is on `last_error <> ''` rather than `status = 'failed'` so it does not silently break if the terminal-state name changes.

## Test

`test_queue_counts_surface_failures` in `test_curator_code_unit.c` inserts two `failed` jobs and one `pending`, then asserts `failing == 2, pending == 1, done == 0` — the middle assertions pin the old invisibility in place: the two dead jobs are in neither of the buckets health used to report.

## Verification

- `unit-test-curator-code-unit` passes (full binary, all cases).
- `make lint` exits 0, no failures.

Not verified: I have not observed the warning render on a live kb with genuinely failing jobs — the assertion is at the counter layer and the `warnings[]` append is by inspection. Hypothesis, unverified for the end-to-end render path.


JBailes:
Superseded by the tested release aggregate in #2095 and the final release follow-up in #2099. The curator health behavior is present on testing at 7704a649; keeping this stale branch open would risk reintroducing its older aggregate context.

PR #2094, opened 2026-07-28
tell a locked-out client how to recover

JBailes:
## What went wrong

`rotate_bearer` replaces the **server-wide** bearer wholesale, so every already-paired client is invalidated at the same instant. What they get back is:

```json
{"error":{"message":"missing or invalid bearer token","type":"authentication_error"}}
```

Indistinguishable from a typo, and with no way forward. The live token lives in `aimee.api.bearer_token` in `<AIMEE_HOME>/aimee.yaml` **inside the container** — which you only know if you have read the source. The thin client echoes the server's text verbatim, so `aimee status` printed exactly that one line and nothing else.

I hit this mid-session. The thin client and every API call started failing at once, and I spent a while treating it as an auth-lockout or rate-limit bug before finding the rotation in `aimee.yaml`'s mtime. The auth logic was correct throughout — the **message** was the defect.

## Change

The 401 now names the recovery path: that a rotation invalidates all paired clients, where the live token is, and the command to adopt it.

The 503 case (`this endpoint requires a configured bearer token`) is a server misconfiguration, not a client credential problem, and deliberately does **not** tell the caller to go re-pair — the test asserts that separation.

## Test

`server_http_auth_error_body` is extracted as a pure function so the wording is testable; the remediation is the entire point of the message and had no coverage. Added to `test_server_http.c` alongside the existing `server_http_authorize` cases:

- still identifies the failure and keeps `authentication_error`
- carries `rotated`, `aimee.api.bearer_token`, and `aimee remote set`
- the 503 body does **not** carry re-pair guidance, and differs from the 401
- both bodies parse as JSON, since clients decode before display

## Verification

- Full `make unit-tests` suite: exit 0, all pass.
- Full shipping set (`make all`, 7 binaries): builds clean — this is what confirms the new TU is linked everywhere, not a spot check.
- `make -C src lint`: **all 36 checks passed**, exit 0.

`server_auth_error.c` is a new translation unit because `server_http.c` sits at the 2500-line `line-check` ceiling (adding the function inline pushed it to 2520 and failed lint). It is wired into `SERVER_SRCS` and both `Rules.mk` link lines.

## Scope

This fixes the **message**, not the blast radius. Rotation still invalidates every paired client at once; supporting multiple concurrently-valid bearers is a real feature with real security surface and does not belong in a release-hardening change. Worth considering separately — flagging it rather than smuggling it in here.


JBailes:
Superseded by #2099, which incorporates additive enrollment and the actionable recovery path, plus fixes the concurrent-worker races, live token-array crash, route/capability reachability, persisted revoke-all semantics, env-managed rotation behavior, and destructive old-server fallback. Exact merged testing head: 7704a649.

PR #2099, opened 2026-07-28
Fix additive enrollment races and memory stack exhaustion

JBailes:
## Summary

Release-follow-up fixes discovered while validating the 0.3.0 testing candidate:

- preserve and report real plane exit status
- enroll client bearers additively without revoking existing clients
- synchronize primary/enrolled bearer reads, enrollment, rotation, enable, and disable across concurrent HTTP workers
- make explicit bearer rotation revoke every enrolled credential in live and persisted state
- preserve enrolled clients across ordinary env-managed restarts, while deployment-secret changes revoke them
- reject unsupported additive enrollment instead of silently falling back to destructive rotation
- expose enrolled bearer count and improve client recovery messages/docs
- move large config/toolset snapshots off deep memory-query stacks
- harden the parallel KB HTTP fixture against the production-ignored SIGPIPE signal
- guard negation token bounds before size conversion

## Release bugs found and fixed

The proposed enrolled-bearer implementation had several correctness/security failures under real concurrency: a two-dimensional token array was cast to `char **` and could segfault; concurrent enrollment could reuse a slot or lose an update; live bearer access raced; rotation did not revoke extras; API enable/disable could overwrite enrollment state; and the enrollment route was unreachable through the write-tier/capability gates. All are covered here.

The full optimized C suite also exposed deterministic stack exhaustion in `unit-test-memory-advanced`: `config_t` is roughly 750 KiB and `toolset_registry_t` is over 1 MiB, with several copies accumulated through a deep retrieval call chain. The snapshots now live on the heap and one immutable query snapshot is reused where possible.

## Validation

- `make all` (all seven shipping binaries)
- `make lint` (37/37, including shipping artifact audit)
- full C unit suite with `TEST_RUN_JOBS=8`
- ASAN+UBSAN `unit-test-memory-advanced` at the default stack limit
- focused server HTTP, config, server dispatch, plane supervisor, text, memory, and KB route tests
- KB HTTP route fixture stress: 40 parallel copies at `-P8`
- Go tests: runtime-web, server-go, control-web
- `git diff --check`
- clang-format-19 dry run

Live exact-image E2E, stress, restart, persistence, LLM-disconnect, provider, and release-artifact validation will follow this PR through testing and promotion.


JBailes:
CI exploratory follow-up: the first run exposed a real random-collision defect in the bounded PAM login throttle. A single direct-mapped username-slot collision could return 429 below the configured budget. Commit `20376671` uses an eight-slot bounded probe: unrelated users no longer inherit a collision, live records are never evicted, lookup work remains fixed, and a saturated probe still fails closed. Added a deterministic 20,000-seed regression; focused throttle and PAM/OIDC route suites, ASAN/UBSAN, and all 37 lint/artifact checks pass locally.

PR #2102, opened 2026-07-28
fix API credential snapshot staleness

JBailes:
## What changed

- read the current on-disk configuration before every `aimee.api` mutation
- republish the immutable live configuration snapshot after each successful save
- add a handler-level regression covering two sequential additive bearer enrollments

## Root cause

Once the server initializes its live configuration snapshot, `config_load()` intentionally returns that immutable snapshot until a reload. The API credential handlers used it as the base of a read-modify-save operation but did not reload after saving. A second enrollment therefore started from the original zero-extra snapshot and overwrote the first enrolled bearer on disk and in the live authorization set.

## Impact

Sequential client enrollment is genuinely additive: the primary bearer and all previously enrolled bearers remain persisted and live-authorized. Enable, rotate, and disable mutations also avoid stale-snapshot overwrites and immediately republish their saved state.

## Validation

- `unit-test-server-dispatch` (new sequential enrollment regression)
- `unit-test-server-http`
- `unit-test-config-snapshot`
- `unit-test-config`
- full optimized `make -C src all`
- clang-format 19 and `git diff --check`

- branch server image deployed over exact `testing-7704a64` KB/LLM images
- live lifecycle: two additive enrollments, immediate authorization, restart persistence, UDS-admin revoke-all, old-token rejection, new-primary acceptance, and automatic restoration of the original config

### DIFF
--- a/src/server/server_api_status.c
+++ b/src/server/server_api_status.c
@@ -141,8 +141,11 @@ int handle_api_status(server_ctx_t *ctx, server_conn_t *conn, cJSON *req)
 
 /* /v1 connections are handled concurrently. Serialize every aimee.api
  * read-modify-save sequence in this file: otherwise two enrollments can select
- * the same slot, or enable/disable can overwrite a just-enrolled credential
- * with the stale snapshot it loaded before that enrollment completed. */
+ * the same slot, or enable/disable can overwrite a just-enrolled credential.
+ * Each mutation also reads DISK rather than config_load's immutable live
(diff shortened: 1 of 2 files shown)


JBailes:
Live container validation passed: two sequential additive enrollments remained authorized; both persisted across restart; an OS-attested UDS administrator rotated/revoked primary plus both extras; only the new primary remained valid; the probe restored the original config and the server returned healthy with zero restarts/OOM.

PR #2191, opened 2026-07-30
Fix the clean-install wizard Deploy failure (first-user bootstrap)

JBailes:
A from-scratch appliance install could not complete the setup wizard.
`POST /v1/deploy/apply` returned 500, `could not provision the first user
(bearer, mTLS, and full-write grant)`, and produced **no diagnostic at all** —
every failure path in `first_user_bootstrap_locked` was a bare `goto` with no
log line. It only started working after `aimee api enable`, an unrelated
recovery command no operator would think to run.

## Root cause

Observed on a clean `.211` install, not inferred:

```
parse_server_api runtime_secret_get=0       <- the first real parse, BEFORE the mint
config_load_file CACHE HIT, bearer=MISSING  <- the 1s reload tick
config: config file change: reloaded        <- publishes the empty-bearer config
first_user: bootstrap rejected: extras_snapshot=0 primary_bearer=MISSING mtls=1
```

`server_main` mints the API primary bearer *after* its first `config_load`, so
the process-wide file cache still holds the pre-mint config. The mint lands in
Vault, the runtime secret store and the local `config_t`, and
`config_snapshot_init` seeds a correct snapshot — but one second later the
`config_reload_if_changed` tick calls `config_load_file`, which serves that
cached copy and **skips the parse** that rehydrates Vault-only values. Publishing
it blanks `server_api_bearer_token` in the live snapshot, and
`first_user_bootstrap_locked` gates on exactly that field.

`aimee api enable` masked it by calling `config_save`, which rewrites `aimee.yaml`
and invalidates the cache, so the next reload genuinely re-parses.

## The fix

Apply the runtime secrets to the reloaded config before publishing, so the
published snapshot equals what `config_load` returns. This covers **every**
Vault-only secret, not just the bearer — `db2_url`, the kb and telemetry tokens
and the curator keys were all blanked in the snapshot by the same path. Applied
before the token computation so an unchanged reload stays a no-op.

The first commit gives each `first_user` failure path a distinct reason, so the
next failure here is diagnosable instead of silent.

## Verification

- New assertion in `test_config_snapshot` fails without the fix
  (`Assertion 'strcmp(got.server_api_bearer_token, "vault-only-primary") == 0' failed`)
  and passes with it. It models the real ordering: seed the file cache, mint
  after it, reload with the file unchanged.
- `unit-test-config`, `unit-test-config-accessors`, `unit-test-config-snapshot`,
  `unit-test-config-snapshot-race` all pass.
- End-to-end on a torn-down-to-nothing `.211` (all containers and volumes
  removed): wizard Deploy now returns **200** `{"ok":true,"status":"started"}`
  with enrollment `ready`, where the same clean boot returned 500 before.

### DIFF
--- a/src/modules/config/config.c
+++ b/src/modules/config/config.c
@@ -2123,6 +2123,19 @@ int config_reload(void)
       pthread_mutex_unlock(&g_snap_wlock);
       return -1; /* parse failure -> keep the running snapshot */
    }
+   /* The published snapshot must equal what config_load() returns, and config_load
+    * is config_load_file + this. Without it a reload publishes a snapshot with every
+    * Vault-only secret blanked, because config_load_file may serve the process cache
+    * and skip the parse that rehydrates them.
+    *
+    * That is not hypothetical: the server mints the API primary bearer AFTER its
+    * first config_load, so the cache still holds the pre-mint config. The 1s
+    * config_reload_if_changed tick then republished that stale copy over the good
+    * snapshot seeded by config_snapshot_init, leaving server_api_bearer_token empty
+    * and failing first-user bootstrap on every clean install (the wizard's Deploy
+    * step returned 500). Applied before the token so an unchanged reload stays a
+    * no-op rather than churning the snapshot. */
+   config_apply_runtime_secrets(&fresh);
    uint64_t tok = config_snapshot_token(&fresh);
    if (atomic_load_explicit(&g_snap_inited, memory_order_acquire) && tok == g_snap_token)
    {
--- a/src/server/server_bearer_auth.c
+++ b/src/server/server_bearer_auth.c
@@ -17,6 +17,7 @@
 
 #include "config.h"
 #include "db1/remote_client_grant.h"
+#include "log.h"
 #include "platform_random.h"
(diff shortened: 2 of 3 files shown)


JBailes:
End-to-end verified on a torn-down-to-nothing `.211` (all containers and volumes removed).

**Before** — same clean boot, stock `:testing`:

```
POST /v1/deploy/apply -> 500
{"error":"could not provision the first user (bearer, mTLS, and full-write grant)"}
first_user: bootstrap rejected: extras_snapshot=0 primary_bearer=MISSING mtls=1
```

**After**:

```
http=200
{"enrollment":{"state":"ready","principal":"webuser:aimee-db02003a2cb6","tier":"full","mtls":true},"ok":true,"status":"started"}
```

With #2192 also applied, the whole wizard Deploy then runs to completion —
`deploy: managed server identity enrolled and verified`, `last_exit=0`, both
containers healthy — so a from-scratch install works for the first time.

Note the `first_user:` line above is the diagnostic added by the first commit
here. It is what turned this from "500 with no logs" into a one-line diagnosis,
and it is why the second commit could be aimed at the right thing.

PR #2192, opened 2026-07-30
Fix managed server identity enrollment on a clean install

JBailes:
The wizard's Deploy step brings the stack up and then fails:

```
deploy: managed server identity enrollment failed:
aimee-kb: the internal database cannot run as root (PostgreSQL forbids it).
```

This was masked until now — `deploy/apply` returned 500 before it ever got this
far (fixed in #2191). With that fixed, the deploy reaches identity enrollment
and fails there, so a from-scratch install still does not complete.

## Root cause

`aimee-server-identity` is a one-shot that mounts the **same** volume as the
running `aimee-kb` and runs as `user: "0:0"` deliberately — it chowns the
identity it installs to uid 1000.

The embedded lane re-seals the embedded DSN into that shared Vault
(`aimee-kb-entrypoint.sh`, the `AIMEE_DB2_URL=... aimee-kb --bootstrap-vault-env`
line). So the one-shot's `--vault-db2-external` probe reports *embedded* too,
regardless of the DSN its own compose entry sets, and it walks into cluster
provisioning:

- as root it is refused outright, which is the observed failure;
- as any user it would `initdb` a **second cluster over the data directory the
  running kb already owns**.

Verified on the box: with the identity service's exact environment, the probe
returns `EMBEDDED`.

## The fix

PostgreSQL forbids running the *server* as root, not connecting to one as root.
When `pg_isready` shows a cluster already listening on the shared socket, connect
to it and `exec`, before the root check applies.

`PGBIN` gains an `AIMEE_DB2_PG_BIN` override purely so this path is testable
without a real PostgreSQL install; deployments never set it.

## Verification

- New `test_build_integrity.sh` case: stubs `pg_isready` as "a cluster is up" and
  asserts the entrypoint reaches the binary and never runs `initdb`. It passes
  with this change; on stock the same scenario runs `initdb` instead.
- End-to-end on a torn-down `.211` — see the PR comment once the deploy run completes.

### DIFF
--- a/deploy/container/aimee-kb-entrypoint.sh
+++ b/deploy/container/aimee-kb-entrypoint.sh
@@ -229,7 +229,15 @@ if [ "$external_db" -eq 0 ]; then
     # libpq reads a directory-valued host as a socket path. Even this local,
     # passwordless DSN follows the credential-shaped config contract: give it
     # to a disposable bootstrap helper, then let the KB load it from Vault.
-    AIMEE_DB2_URL="postgresql:///$DB?host=$PGSOCK" aimee-kb --bootstrap-vault-env
+    #
+    # Name the role explicitly. initdb made THIS user the cluster superuser, and
+    # this DSN is sealed into a Vault on a volume other containers share: a
+    # sharer running as a different OS user (the managed deploy's root
+    # aimee-server-identity job) would otherwise have libpq default the role to
+    # its own user name and fail with "DB2 not reachable". Vault holds this value
+    # for every sharer, and the entrypoint scrubs AIMEE_DB2_URL from the
+    # environment before exec, so their own compose-supplied DSN cannot fix it.
+    AIMEE_DB2_URL="postgresql:///$DB?host=$PGSOCK&user=$(id -un)" aimee-kb --bootstrap-vault-env
 
     start_embedder
 
--- a/deploy/container/aimee-kb-entrypoint.sh
+++ b/deploy/container/aimee-kb-entrypoint.sh
@@ -237,7 +237,16 @@ if [ "$external_db" -eq 0 ]; then
     # its own user name and fail with "DB2 not reachable". Vault holds this value
     # for every sharer, and the entrypoint scrubs AIMEE_DB2_URL from the
     # environment before exec, so their own compose-supplied DSN cannot fix it.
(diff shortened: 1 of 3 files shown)


JBailes:
End-to-end verified on a torn-down-to-nothing `.211` (all containers and volumes removed), with #2191 applied to the server image and this branch built into the kb image.

**Before** (this branch's first commit only — the root refusal is gone, but the DSN still names no role):

```
aimee-kb: PostgreSQL already running on /var/lib/aimee/run; using it instead of starting a second cluster
aimee-kb: DB2 not reachable at postgresql:///aimee_shared?host=/var/lib/aimee/run
deploy: managed server identity enrollment failed
last_exit=1
```

**After** (both commits):

```
deploy: managed authority roots and signed JWKS verified
deploy: managed server identity enrolled and verified
last_exit=0
```

Both containers healthy, and the deploy completes in 35s instead of failing after 95s of retries.

The bundled embedder also comes up correctly on the same run, which exercises the entrypoint path this change touches:

```
aimee-kb: starting bundled embedder (bekko-a25m) on :8760
embedder-server: loaded hotchpotch/bekko-embedding-v1-a25m dim=384 threads=8 quant=fp32
db2: embedder serving identity: bekko-a25m/8721341054416418
```

kb `/v1/health` reports `db2_ok:true, pgvec_ok:true, embed_ok:true`.

One unrelated wart observed on that run, filed separately rather than fixed here: the baked HF cache logs `Permission denied` on a tree-cache file. Non-fatal — the model loads and serves at 384 dim.

JBailes:
Self-review notes on two deliberate tradeoffs in the reuse branch, so they are visible rather than buried:

1. **`start_embedder` still runs there.** A one-shot does not need the bundled embedder, so on a deployment that sets `EMBEDDER_MODEL` this loads a model the job never uses. I kept it because the *daemon* can legitimately reach the same branch (restarting against a cluster another container on the volume still owns) and it does need the embedder there. Distinguishing the two would mean hard-coding the one-shot subcommand list into the entrypoint — a second place that has to track `kb_main`'s subcommands, which is the coupling the `--print-embedding-model` comment in this same file warns about. Wasteful in one case beats wrong in the other.

2. **The reuse branch skips `createdb` / `CREATE EXTENSION`.** That is intentional: it only runs when a cluster is already listening, which means provisioning already happened.

Also hardened `id -un`: a runtime that runs this image under a uid with no passwd entry would otherwise seal `user=` empty, which is worse than omitting the parameter.

PR #2195, opened 2026-07-31
Fix the env-var deployment path: zero vectors stored, and a preflight that claimed ready

JBailes:
Found while verifying the clean install, then extended after a question worth asking: **what if the deployment is script + env vars instead of the wizard?** That path was broken, and my earlier verification had only ever exercised the wizard.

## 1. The split stack stored zero vectors while reporting healthy

`deploy/compose/aimee.yaml` defaulted `AIMEE_EMBEDDING_DIM` to **1024** and `deploy/smoothnas/aimee.compose.yaml` to **2560** — the retired `aimee-llm` CPU and GPU tier widths. The bundled embedder is **384**. `deploy/container/aimee-managed.compose.yaml` was updated at the bekko cutover and these two were not, which is exactly why wizard-only testing missed it.

Observed on a clean split deploy with operator defaults:

```
pgvec: memory embedding dim mismatch: got 384, expected 1024; refusing upsert
pgvec_vectors: 0
```

while `aimee memory store` reported `stored memory 1`, search returned the record from the lexical path, and kb health reported `status: ok, pgvec_ok: true, embed_ok: true`. Dense retrieval was permanently dead and nothing said so.

Default both to empty, as the managed compose already does, so the kb uses the width of the embedder it actually serves.

## 2. That condition was invisible, which is the worse half

A per-row WARN in the kb log is not something an operator reads, and every other signal said healthy. This also breaks the invariant in `docs/DEPLOYMENT.md`: *"The KB must report explicit degradation when a configured inference stage is unavailable. It cannot claim a dense result after silently skipping that stage."*

Health now publishes `embedding_dim_refused`, plus the expected and offered widths when non-zero, so `aimee kb status` shows it. This matters beyond the defaults: an operator who pins the wrong width, or points `AIMEE_EMBEDDER_URL` at an endpoint of a different width, hits the same silent loss.

`embed_ok` is unchanged — it reports that an embedder is *configured*, which stays true.

## 3. `agent remove` could not delete the only configured delegate

The `agents.json` deletion guard refuses any save that empties a populated registry. It cannot tell a deliberate removal of the last delegate from a zeroed or failed-to-load cfg, so it refused both, as the `aimee` user and as root, with the file writable. Removing one of several worked, which is why it hid:

```
===REMOVE-ONE-OF-TWO   Delegate 'beta' removed.
===REMOVE-LAST         aimee: could not save agents.json
```

The one caller that knows the empty result is intended now says so. The guard still refuses every other zeroed save, and the test pins that scoping.

## 4. The `/v1/health` 401 log loop

The container healthcheck GETs `/v1/health` unauthenticated every 10s and treats 401 as healthy — it has no bearer to present. Logging that designed probe at INFO produced **245 identical lines in 40 minutes**. Demoted exactly that shape to DEBUG; every other rejection still logs. Verified 0 lines after, with the container still `healthy` (the healthcheck reads the HTTP status, not the log).

## 5. Baked model cache unreadable by the runtime user

The bake runs as root and `huggingface_hub` writes tree metadata `0600`, so `aimee` (uid 999) could not read it; every KB start logged `Ignoring corrupted tree cache file … Permission denied` and re-walked what the cache exists to avoid. Weights were already world-readable, so this was noise. `chmod -R a+rX` after the bake. Verified 0 denials, model still loads at `dim=384`.

## Deliberately NOT changed

`AIMEE_SERVER_TEAM_ID is unset` (ERROR) and `management JWKS authorization unavailable` (WARN) are transient on the **wizard** path — both clear once Deploy runs:

```
07:45:01 WARN  management JWKS authorization unavailable
07:45:47 INFO  management JWKS authorization ready
08:25:32 INFO  management JWKS authorization ready     <- restart, no warning
```

`server_runtime_identity_load` explains it: explicit `AIMEE_SERVER_ID`/`AIMEE_SERVER_TEAM_ID` win, and otherwise it falls back to the managed metadata that wizard enrollment writes.

On a **standalone env-var** deployment the picture differs — `AIMEE_SERVER_MGMT_JWKS_TRUST_BUNDLE` defaults to a path rather than empty, so that warning fires on every boot unless an authority is provisioned. It is truthful there, so I have not changed it, but it is not "transient" outside the wizard path and deserves a separate decision.

## Verification

- `test_agent_config_deletion_guard` cases (4) and (5) fail without the fix and pass with it.
- `unit-test-agent` passes; `make lint` `aimee-home-check` ok.
- Split stack deployed on a torn-down `.211` in two configurations: operator defaults, and a deliberately wrong explicit pin.

---

## 6. The write-tier preflight reported READY without the trust bundle

`server_write_tier_config_state()` checked only that `AIMEE_SERVER_MGMT_JWKS_TRUST_BUNDLE` was
non-empty, never that the file existed. `compose.server-standalone.yaml` defaults it to a
conventional path nothing mounts, so a standalone env-var deployment that never provisioned an
authority reported **READY while every KB-issued write token was denied** — and because READY is the
silent branch, it suppressed the one startup line naming the variable and the consequence, leaving
only a vague recurring `management JWKS authorization unavailable`.

The existing test encoded that defect:

```c
setenv("AIMEE_SERVER_MGMT_JWKS_TRUST_BUNDLE", "/run/aimee/management/jwks-trust-bundle.json", 1);
assert(server_write_tier_config_state() == SERVER_WRITE_TIER_CONFIG_READY);  // nonexistent -> READY
```

Now the preflight checks the bundle is readable, and the message names the path when one was
supplied. Readability only: this answers "did the operator supply the inputs"; structural validation
stays in `build_config`, which fails closed per request. Runs once at startup.

**No security change** — authorization already failed closed either way. This is the server telling
the truth at startup instead of claiming ready.

The conventional default is deliberately kept: mount a bundle there and it works, omit it and the
server names the file it wanted.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -276,6 +276,13 @@ for name, spec in table.items():
         snapshot_download(code_repo, allow_patterns=["*.py", "*.json", "*.txt"])
 PYBAKE
 
+# The bake runs as root and huggingface_hub writes its tree-cache metadata 0600, so
+# the `aimee` runtime user cannot read it. The weights themselves land world-readable,
+# so the model still loads — it just logs "Ignoring corrupted tree cache file …
+# Permission denied" on every start and re-walks what the cache existed to avoid.
+# a+rX: readable everywhere, traversable on directories, and no file gains +x.
+RUN chmod -R a+rX /opt/aimee/models
+
 # Sidecar clients (the LLM access code the kb invokes via popen).
 COPY scripts/embed-remote.py scripts/llm-chat.py \
      scripts/learning-synthesize.py scripts/curator-extract.py scripts/llm-rewrite.py \
--- a/deploy/compose/aimee.yaml
+++ b/deploy/compose/aimee.yaml
@@ -34,8 +34,14 @@ services:
       # the builtin serves forever — the wizard's embedder silently never runs.
       EMBEDDER_MODEL: ${EMBEDDER_MODEL:-}
       AIMEE_EMBEDDER_URL: ${AIMEE_EMBEDDER_URL:-}
-      # Embedding dim — CPU tier default 1024; the GPU override sets 2560.
-      AIMEE_EMBEDDING_DIM: ${AIMEE_EMBEDDING_DIM:-1024}
+      # Embedding dim. Leave EMPTY so the kb uses the dimension of the embedder it
+      # actually serves (the bundled bekko-a25m is 384). This carried the retired
+      # aimee-llm CPU tier's 1024: with the bundled embedder the kb came up healthy,
+      # accepted writes, and then refused every vector upsert with
+      # "dim mismatch: got 384, expected 1024" — dense retrieval silently dead.
+      # Set it ONLY for an external AIMEE_EMBEDDER_URL, whose width the kb cannot
+      # know.
+      AIMEE_EMBEDDING_DIM: ${AIMEE_EMBEDDING_DIM:-}
       AIMEE_KB_HTTP_BIND: "1"
       # The curator synth sidecar (curator-extract.py) reads LLM_ENDPOINT, NOT
       # AIMEE_LLM_URL. Defaulted empty for the same reason as AIMEE_LLM_URL above:
--- a/deploy/smoothnas/aimee.compose.yaml
+++ b/deploy/smoothnas/aimee.compose.yaml
@@ -45,7 +45,11 @@ services:
       # the builtin serves forever — the wizard's embedder silently never runs.
       EMBEDDER_MODEL: ${EMBEDDER_MODEL:-}
       AIMEE_EMBEDDER_URL: ${AIMEE_EMBEDDER_URL:-}
-      AIMEE_EMBEDDING_DIM: ${AIMEE_EMBEDDING_DIM:-2560}
+      # Empty so the kb uses the width of the embedder it serves (bundled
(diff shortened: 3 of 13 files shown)


JBailes:
Split-stack verification, both directions, on a torn-down `.211`. The discriminator is
`memory_embeddings` row count — **not** `pgvec_vectors`, which counts KB *document* points and stays
0 until something is ingested. Two earlier runs of mine were inconclusive for exactly that reason,
plus a store that never executed.

**Case A — corrected default (`AIMEE_EMBEDDING_DIM` empty):**

```
DIM_ENV=''
embedder-server: loaded hotchpotch/bekko-embedding-v1-a25m dim=384
stored memory 1   -> memory_embeddings rows: 13
stored memory 2   -> memory_embeddings rows: 26
embedding_dim_refused: 0     mismatch WARNs: 0
```

**Case B — operator pins the wrong width (`AIMEE_EMBEDDING_DIM=1024`):**

```
DIM_ENV='1024'
stored memory 1
embedding_dim_refused: 13, embedding_dim_expected: 1024, embedding_dim_offered: 384
mismatch WARNs: 13
```

The same 13 embeddings are refused in B and stored in A. Before this change, Case B reported
`status: ok, pgvec_ok: true, embed_ok: true` and nothing else — the refusals existed only as per-row
WARNs in the kb log.

Note that CI does not exercise the split stack, so a green build here says nothing about the defect
this PR is mainly about; the evidence above is the verification that matters.

PR #2197, opened 2026-07-31
Stop losing the knowledge service's reason, and re-embed a memory's units

JBailes:
Two defects found while migrating `aimee-prod` to the bundled 384-dim embedder. Both were filed rather than worked around in #2196; this fixes them.

## 1. A failing command reported success

`aimee memory embed --all` exited **0 printing nothing** while aimee-kb was answering `memory.embed all=true requires version`. `aimee kb reembed` said `knowledge service reembed failed` while the KB was naming the exact config gate to set.

Two layers, one cause each:

- **`kb_client_v1_post_json` frees the response body on any non-2xx and returns NULL**, so the reason was gone before any caller saw it. Action requests then synthesised `knowledge service action memory.embed returned HTTP 400` — non-NULL, so it relayed as a *success* containing an error nobody prints.
- **`kb_relay_send` passes any parseable body through as success**, so even a preserved error would relay as OK.

Added `kb_client_v1_post_json_keep_error`, which returns the body. `kb_v1_action_request_timeout` now uses it, so **every** `/v1/actions/*` caller reports what the KB actually said; `kb.reembed` uses it directly. Added a checked relay for the two handlers with a refusal worth reading, recognising both shapes the KB emits (`{"error":...}` and `{"status":"error"|"unavailable"}`).

`kb_relay_send` itself is unchanged — switching every existing relay route at once is a separate decision.

## 2. Re-embedding a memory left most of its vectors missing

I originally reported this as "~9 vectors per memory". That was wrong, and the real shape matters. `memory_embeddings` is keyed by `point_id` with a `record_type` column. One `memory store` produces:

```
record_type | count
------------+-------
 unit       |    12
 memory     |     1
```

`memory_embed` writes the memory's own vector and stops. The write path (`memory_refresh_derived_metadata`) also calls `memory_refresh_unit_embeddings`. So after a dimension change dropped the table, re-embedding every memory restored **4 rows where 25 belonged** — units are the bulk, unit search stayed dead, and nothing reported the shortfall.

The repair path now calls the same refresh the write path does.

## Verification

- `unit-test-server-dispatch`, `unit-test-memory-provider`, `unit-test-harness-memory` pass.
- On `.211`: record-type distribution captured before wipe, after wipe, and after repair, so "restored" means the `unit` rows came back rather than a count merely moving.
- `make lint` clean except `bus-blast-radius-check`, which needs all 7 shipping binaries and this worktree builds 2.

## Context

Both were found by running the commands during a real prod migration, not by reading them. The first is the more dangerous class: an operation that fails while reporting success is worse than one that is missing, because nothing prompts you to look.


JBailes:
Both fixes verified on `.211`, with the running KB image asserted against the build under test — three earlier attempts failed for harness reasons (server-only rebuild; the KB deployed from the baked managed compose rather than `compose.server-managed.yaml`; bash substring syntax under `dash`), each producing a plausible-looking "no change" that was really "the code never ran".

**#13 — a failing command no longer reports success**

```
$ aimee memory embed --all
aimee: memory.embed all=true requires version
exit=1
```

Before: exit **0**, no output, while the KB was refusing it.

**#14 — re-embedding a memory restores its units**

```
PRECONDITION OK: kb runs the build under test   (sha256:48bcaf7e6ff2c both sides)
units source rows: 57
vectors now:   memory=4
after repair:  memory=4 unit=57
```

All 57 unit vectors restored. Three prior runs restored zero, which is what prompted asserting the image ID rather than trusting the deploy.

Note the corrected diagnosis in the description: this was never "~9 vectors per memory". `memory_embeddings` is keyed by `point_id` with a `record_type` column, and `unit` rows outnumber the `memory` row roughly 12:1.

PR #2221, opened 2026-08-01
Promote testing to main for v0.3.1

JBailes:
Promotes `testing` to `main` to cut **v0.3.1**.

## Why this release exists

`v0.3.0` and `v0.2.196` were promoted mid-cycle in error and are marked prerelease. `v0.2.192` is the Latest release, but **the `:latest` Docker tag still points at the v0.3.0 build**, because marking a release prerelease does not move an image tag. Anyone following the quickstart today gets a mid-cycle build.

Releasing 0.3.1 repoints `:latest` at a build that has been tested.

## What it carries

399 commits. The ones that matter for a fresh install, all landed in the last day:

- a clean install could enrol no identity and store zero vectors (#2191 chain);
- an operator-supplied dashboard login was discarded for a generated one (#2203);
- tool-using delegates got an empty workspace mount on any compose deploy (#2206);
- a model that accepts one temperature was sent another, failing every delegated call (#2200);
- a containment decision read config from disk instead of the live snapshot (#2204);
- `aimee-llm` is retired from the wizard, which had been asking operators to place a container that cannot start (#2213).

Plus the release process itself (#2208): a push to `main` no longer publishes anything until a human approves the version.

## The gate applies to this PR

Merging this triggers the release workflow, which infers `0.3.1` from the tags, writes it to the run summary, and **waits** on the `release` environment. Nothing is tagged or published before that approval.


JBailes:
Superseded by a promote branch that resolves the revert conflicts by taking testing's tree verbatim. main carried only reverts of an accidental promotion, so re-applying them was the wrong resolution.

PR #2242, opened 2026-08-01
aimee-kb image matrix, one embedder/synthesis surface, and PR-time image builds

JBailes:
Retires the last of the `aimee-llm` container, gives `aimee-kb` a 2x2 image
matrix, and makes the embedder and synthesis settings one surface instead of two.

## What changes

**One name per setting.** The embedder and synthesis settings had two parallel
surfaces — config keys and environment variables — because only 4 rows in
`config_fields[]` declared an env var and the embedder/synthesis rows declared
none. They were read by ad-hoc `getenv()` and hand-plumbed through compose. They
are now ordinary table-declared fields: `EMBEDDER_URL` / `EMBEDDER_DIMS` /
`EMBEDDER_MODEL` / `EMBEDDER_API_KEY` and `SYNTHESIS_ENDPOINT` /
`SYNTHESIS_MODEL` / `SYNTHESIS_API_KEY` / `SYNTHESIS_THINKING`. The config *file*
uses these names too — it was still writing `embedding_*` while the fields behind
them had moved, so the file and the key could diverge.

**One synthesis role.** Tier-A/Tier-B is gone. Every curator stage resolves the
same provider. The split also made the reasoning stages *refuse* the env
endpoint, so a bundled-model deployment ran extraction and left synthesis idle.
Thinking becomes a global, user-settable `synthesis_thinking` (default on)
instead of a rule implied by the stage.

**Four aimee-kb images**, one Dockerfile, selected by build-args:

|                  | no llama.cpp     | llama.cpp bundled    |
| ---------------- | ---------------- | -------------------- |
| bekko-a25m (384) | `aimee-kb`       | `aimee-kb-llm`       |
| nomic-v2 (768)   | `aimee-kb-nomic` | `aimee-kb-nomic-llm` |

**Bundled synthesis runs beside the kb**, with the weights baked into the `*-llm`
images at UD-Q6_K_XL (7.46 GB E4B, 4.71 GB E2B) and built once via a separate
`aimee-model-*` image rather than once per variant per arch. The container
downloads nothing at runtime: a first-start fetch fails on a rate limit, a proxy,
a flaky link or an air-gapped host, and every one of those reaches the operator as
"synthesis never started" long after the deploy looked fine. `--no-mmproj` plus
`LLAMA_ARG_MMPROJ_AUTO=false`: every benchmark passed `--no-mmproj` and the
shipped service never did, so production loaded a ~0.5 GB vision projector for a
text-only task.

**The wizard** asks the two questions that remain, and states the embedder
one-way door at the point of choice.

## CI

`publish-images.yml` now runs on `pull_request` — **build only, push nothing**.
Publishing stays a merge-time act, enforced in four places because any one left
alone would still push or fail: `push=` on the build output, the ghcr login, the
digest export/upload, and the merge job that tags the manifest. A merge rebuilds
from cache and pushes then.

## Bugs found on the way

Each of these was silent — no error, just a setting that stopped working:

- `deploy_apply.c` compared env prefixes with hand-counted lengths (`21`, `24`).
  `21` was right for `AIMEE_LLM_AUTH_TOKEN=` and wrong for `SYNTHESIS_API_KEY=`
  (18), so the managed-credential filter leaked a second copy of the key into the
  child environment. Now `sizeof-1`.
- `readiness.ts` computed "is an embedder configured" from a deleted field, so
  the wizard's embedder step could never complete.
- `RESTART_KEYS` listed nine retired keys, so the restart notice could never fire.
- `KB_SETTINGS` named renamed keys, and `console_settings()` *skips* rows it
  can't resolve — those settings would have vanished from the KB console.
- `config_save.c` wrote YAML keys that `config.c` no longer parsed.
- `EMBEDDER_API_KEY` was emitted by the deploy layer and received by no container.

## Not verified

**All six images now build in CI** on amd64 and arm64, which resolves the earlier
"no image has been built" caveat and confirms the `unsloth/gemma-4-E{2,4}B-it-GGUF`
repo names against the Hub — the build fetches from them.

**No container has been run.** Building is not booting: nothing in
`.github/workflows/` invokes `scripts/aimee-kb-docker-smoke.sh`, so the
llama-server invocation, the loopback handover and `start_synthesis`'s branch
logic are still unexercised by CI. The smoke script does cover them
(`scripts/aimee-kb-docker-smoke.sh:234-270`) and is being run by hand against
`aimee-kb-llm-e4b` on the disposable test box before this reaches any real
deployment. Wiring it into CI is follow-up work, not part of this PR.

`LLAMACPP_SHA256` is present but empty, so the build warns rather than verifying
the release asset.

Both are written down in `docs/SYNTHESIS_MODELS.md` rather than left implicit.

## Upgrade impact

**Hard cut, no aliases**, per an explicit decision. A deployment that upgrades
without editing its environment and `aimee.yaml` keeps running, falls back to the
builtin lexical embedder and idles synthesis, without erroring.
`docs/UPGRADING.md` has the full old→new table and what to check afterwards.

## Verification

`make lint` (40 checks) · `make all` · 28 C tests touching this surface ·
`tsc -b` + 155 frontend tests · docs-gen-check, check-kb-container-packaging,
check-sidecar-clients, check-deploy-env-consumed, check-curator-story,
check-config-encapsulation, check-proposal-links.


JBailes:
## Reviewer's guide

146 files, but most of it is one mechanical rename. Here is where the judgement
actually sits, and where I would look hardest.

### Read in this order

1. **`src/modules/config/config.h` + `config_fields.c`** — the whole change in one
   place. Every field, its env name, its reload class. If this table is right,
   most of the rest follows.
2. **`src/modules/config/config_database.c`** — `config_emit_deploy_env()`. What
   the deploy layer hands a container, and what it deliberately does *not*
   (credentials).
3. **`src/kb_curator_provider.c`** — the Tier-A/Tier-B collapse. One provider for
   every stage now.
4. **`Dockerfile`** — the six-image matrix, the llama.cpp build, the baked model.
5. **`deploy/container/aimee-kb-entrypoint.sh`** — `start_synthesis()`.
6. **`frontend/src/setup/deployTopology.ts` + `DeployTopology.tsx`** — rewritten,
   not edited. Worth diffing against `origin/testing` rather than reading the
   patch (see "what I got wrong" below).

Everything else is the rename following through: ~90 C files, the compose/deploy
YAML, the sidecar scripts, docs. Mechanical, and the tests cover it.

### Skip

`docs/gen/*` is generated (`make docs-gen`). `config_accessors*.{c,h}` are
generated (`src/gen_config_accessors.py`) — read the generator's input, not its
output.

### Security-relevant, look properly

- **`config_database.c`** must not `EMITF` any credential. `Config.Env` persists
  and shows up in `docker inspect`. I got this wrong first time and
  `check-vault-only-container-env` caught it; the deploy-env test now asserts the
  *absence*.
- **`server/deploy_apply.c`** — env-prefix matching moved from hand-counted
  lengths to `sizeof-1`. The old `21` was right for `AIMEE_LLM_AUTH_TOKEN=` and
  wrong for `SYNTHESIS_API_KEY=` (18), so the managed-credential filter silently
  stopped matching and leaked a second copy of the key into the child env.
- **`Dockerfile`** — the model is verified by sha256 against Hugging Face's own
  digest. That is what makes `AIMEE_MODEL_MIRROR` safe: a mirror is an untrusted
  input, and without a pinned hash anyone who can write to one changes which model
  your image runs.
- **`LD_LIBRARY_PATH`** has no trailing colon. An empty entry means "search the
  current working directory" for shared objects.

### Decisions taken during the work (not mine to make, please sanity-check)

- **Hard cut, no aliases.** An old env var or `aimee.yaml` key is not read. An
  existing deployment loses those settings on upgrade — silently, since the KB
  still starts and falls back to its lexical embedder. `docs/UPGRADING.md` has the
  mapping table. This is the highest-blast-radius decision in the PR.
- **Thinking is global and default-on** (`synthesis_thinking`), replacing a
  per-tier rule. `testing` suppressed it for the mechanical stages with a comment
  citing an observed failure (extraction returning 0 facts when the reasoning pass
  ate the output budget). That suppression is gone.
- **Q6 UD** (`UD-Q6_K_XL`, unsloth) as the shipped quantisation. The published F1
  numbers were measured at Q8_0, so the docs still carry a mismatch caveat.
- **Models baked into the image** rather than fetched at first run, which is why
  there are six tags instead of four.

### What I got wrong, in case the pattern repeats elsewhere

Eight bugs, all found by running things rather than reading them:

| bug | would have caused |
| --- | --- |
| credentials in `Config.Env` | secrets in `docker inspect` |
| no Linux arm64 llama.cpp asset | arm64 images unbuildable |
| `LLAMA_BUILD_EXAMPLES=OFF` | deleted the server target |
| llama.cpp pinned to `b4585` | **no gemma-4 support — loads nothing** |
| `--no-mmproj` absent at that pin | server rejects its own args |
| `Q4_K_M` not published in those repos | **silently serves a different model** |
| `.so` symlinks dropped by `-type f` | server cannot start |
| `LD_LIBRARY_PATH` trailing colon | loads `.so` from cwd |

Plus one regression I introduced and then fixed: resolving the frontend merge
conflict by taking my rewrite wholesale **dropped `testing`'s embedder seeding**.
An unset `embedder_model` is not a working default — the entrypoint starts no
embedder and the KB serves its builtin lexical one forever while reporting
retrieval degraded. Restored in `7d64fdaf6` with a test. If you find other
behaviour missing from those four frontend files, that is the likely cause.

### Verified

- Six images built in CI on **amd64 + arm64**, pulled from ghcr, identity checked.
- All six **booted** on a real host: embedder serves the right width, prefixes are
  applied (query≠document for nomic, query==document for bekko), llama-server
  loads its baked model and completes a request. 36/36.
- Full compose stack booted from a **published** image: vault bootstrap, DB2
  schema, `/v1/search`, embedder `serving_id`, bundled synthesis answering, and no
  credential in `Config.Env`.
- `make lint` (40 checks), `make all`, the C suites, `tsc -b`, 160 frontend tests.

### NOT verified — please weigh before merging

- **arm64 images have never been run.** Built on both arches; the test host is
  amd64 only.
- **The upgrade path has never been exercised.** No one has taken a real
  pre-existing install with `embedding_model` in its `aimee.yaml` and upgraded it.
  Given the hard cut, this is the gap I would close first.
- **The wizard has never rendered in a browser.** 160 unit tests, no UI.
- **`aimee-server` / control-web** against the renamed keys is unexercised.
- **`embedding_command` stays as a JSON wire key** in `kb_client.c`,
  `kb_service.c`, `kb_http.c`, `cli_v1_routes.c` and `server_state.c`. It is an
  API contract with an openapi spec and conformance checks, so renaming it is a
  separate decision — but it now differs from the config key of the same concept.

PR #2248, opened 2026-08-01
kb: stop serving privileged routes unauthenticated

JBailes:
## Problem

The kb enforces auth **only when a bearer is configured** — `kb_http.c` gates on
`if (bearer_token && bearer_token[0])`. Nothing configured one, and the plain listener
binds the container network rather than loopback.

Verified on a live deployment, then reproduced in an isolated one:

```
POST /v1/actions/memory.find_facts     (no credentials at all)
-> 200, with stored memory facts
```

Anything able to route to the container could read the knowledge base. The mTLS
listener was never the weak point — this socket beside it was.

## Fix — three parts, none sufficient alone

**1. The deployment provisions `AIMEE_KB_API_BEARER_TOKEN` for both services.** The
server *already* sent this bearer on the plain path; only the kb never asked for it.
Required rather than defaulted (`:?`), so a deployment cannot come up half-configured
and silently unauthenticated.

**2. The kb refuses a non-loopback bind with no bearer**, so this cannot regress to
auth-off by dropping a variable. A loopback-only kb is unaffected — that is the
default when `AIMEE_KB_HTTP_BIND` is unset.

**3. `GET`/`HEAD /v1/health` is exempt from the gate.** Without this the container
healthcheck presents no credential, gets 401, and the kb is reported unhealthy while
serving correctly. **I hit exactly that**, which is why it is in this PR. Only
`/v1/health`: it is answered in `kb_http_route_ex`, whereas `/v1/version` delegates to
`kb_http_route` and applies its own check, so exempting it there would not take effect.

Separately, the **kb client refuses to send its bearer in cleartext to a non-loopback
host** — a rule the thin client already enforced. The shared rule moved to
`cleartext_guard.h`, header-only for the reason `dependency_breaker.h` is: `kb_client`
links into many focused test binaries carrying neither `aimee_client.o` nor the logging
object. The first version of this broke four of them; the header fixes that.

## Verification

Everything below ran on a **disposable stack** (a dedicated LXC with its own server +
kb), never on production.

Final auth matrix, on the exact code in this PR, with both containers healthy:

```
health  noauth: 200     <- healthcheck works
action  noauth: 401     <- the hole, closed
action badauth: 401     <- wrong token rejected
action  w/auth: 200     <- server<->kb intact
```

```
aimee-aimee-server-1 | Up (healthy)
aimee-aimee-kb-1     | Up (healthy)
```

Compose refuses a half-configured deployment:

```
required variable AIMEE_KB_API_BEARER_TOKEN is missing a value:
the kb bearer must be set; generate one per deployment and give the same value to aimee-server
```

The bind gate fires with the remedy in the message:

```
ERROR kb_http: refusing to bind 0.0.0.0:8741 with no bearer configured: this socket
would serve privileged routes unauthenticated. Configure the kb bearer, or unset
AIMEE_KB_HTTP_BIND to keep the plain listener on loopback.
```

Unit tests (the ones the first attempt broke):

```
unit-test-kb-client-search -> ok
unit-test-kb-http-client   -> passed
unit-test-aimee-client     -> ALL PASS
unit-test-kb-client-memory -> ok
```

`make lint` — all 40 checks pass. Both images build under `-Wall -Wextra -Werror`.

## Deployment note

`AIMEE_KB_API_BEARER_TOKEN` has already been provisioned in the production `.env`
(backup: `.env.bak-pre-kbtoken`) so this merge does not block the next deploy. It is
inert until these images land. Any **other** deployment must set it before updating,
or compose will refuse to start the kb — deliberately, since the alternative is
running unauthenticated.

### DIFF
--- a/src/aimee_client.c
+++ b/src/aimee_client.c
@@ -6,6 +6,7 @@
  * identically on Linux, macOS, and Windows.
  */
 #include "aimee_client.h"
+#include "cleartext_guard.h" /* the shared cleartext rule, also used by kb_client */
 #include "http_content_encoding.h"
 #include "platform.h"
 #include "platform_net.h"
@@ -510,25 +511,13 @@ static char *read_response(int fd, aimee_tls_t *tls, size_t *out_len, int want_r
    return resp;
 }
 
-/* A non-loopback plaintext connection must never carry a credential. localhost,
- * 127.0.0.0/8, and ::1 are the only hosts where a cleartext bearer stays on the
- * machine; anything else would put it on the wire. */
-static int host_is_loopback(const char *host)
-{
-   if (!host || !host[0])
-      return 0;
-   if (strcmp(host, "localhost") == 0 || strcmp(host, "::1") == 0 || strcmp(host, "[::1]") == 0)
-      return 1;
-   return strncmp(host, "127.", 4) == 0; /* 127.0.0.0/8 */
-}
-
 /* Security guard (also exposed for tests): 1 when sending |token| to a server at
  * (|is_https|, |host|) would put the credential on the wire in cleartext — a
  * non-empty bearer over plaintext http:// to a non-loopback host. tcp_request
  * refuses such requests rather than leak the bearer. */
 int aimee_client_would_leak_cleartext(int is_https, const char *host, const char *token)
 {
-   return (token && *token && !is_https && !host_is_loopback(host)) ? 1 : 0;
+   return cleartext_would_leak(is_https, host, token);
 }
 
 static char *tcp_request(const char *url, const char *token, const char *method, const char *path,
--- a/src/headers/cleartext_guard.h
+++ b/src/headers/cleartext_guard.h
@@ -1,0 +1,37 @@
+/* cleartext_guard.h: one rule for "would this put a credential on the wire?".
+ *
+ * Two clients send bearer tokens: the thin client (aimee_client.c) to
+ * aimee-server, and the kb client (kb_client.c) to aimee-kb. Both must refuse
+ * plaintext to a non-loopback host, and they must agree on what loopback means
+ * — a second, subtly different copy of this check is how one of them ends up
+ * weaker than the other.
+ *
+ * Header-only for the same reason dependency_breaker.h is: the kb client links
+ * into many focused unit-test binaries that do not carry aimee_client.o, and a
+ * shared implementation must not drag a new link dependency into all of them.
+ */
+#ifndef DEC_CLEARTEXT_GUARD_H
+#define DEC_CLEARTEXT_GUARD_H 1
+
+#include <string.h>
+
+/* localhost, 127.0.0.0/8, and ::1 are the only hosts where a cleartext bearer
+ * stays on the machine; anything else puts it on the wire. */
+static inline int cleartext_host_is_loopback(const char *host)
+{
+   if (!host || !host[0])
+      return 0;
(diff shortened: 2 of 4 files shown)


JBailes:
Force-pushed a correction. The branch previously reverted #2246's changes to `kb_client.c` — the breaker-suppression tracking, retry-window propagation into the typed error, and the recovery logging. That happened because the branch was assembled with `git checkout <sha> -- <file>`, which took a whole-file version predating #2246, so the rebase saw no conflict and silently dropped their work.

Rebuilt on top of `origin/testing`, re-applying only the cleartext guard. `kb_client.c` is now **+64/-0** — purely additive — and #2246's symbols are all still present. Unit tests re-run on the corrected tree: `unit-test-kb-client-search`, `unit-test-kb-client-memory`, `unit-test-aimee-client` all pass.

JBailes:
**Scope narrowed — this no longer closes the vulnerability, and the title/description now say so.**

The original version added `AIMEE_KB_API_BEARER_TOKEN` to the compose services. CI was right to reject that:

```
FAIL: server or KB container definitions persist credentials outside Vault
```

Credentials must be sealed into the Vault at first boot, not persisted in a long-lived container environment. `vault_env_bootstrap` already classifies `AIMEE_KB_API_BEARER_TOKEN` as a credential via its `_TOKEN` suffix, so the provisioning belongs in that first-boot sealing path. I am not improvising that at the end of a long session — it is follow-up work.

I also **downgraded the listener gate from refuse to warn**. Refusing to bind without a bearer would break every managed deployment, since none provisions one yet. I saw that directly: with the gate refusing, the kb came up **unhealthy** in the test environment.

Also worth recording: `make lint` does **not** run `check-vault-only-container-env.py` — it lives in `build-integrity`. That is why this passed locally and failed in CI.

**What remains open:** the kb still serves privileged routes unauthenticated on a network-reachable socket. This PR makes that loud at startup and prepares the two pieces the real fix needs (the `/v1/health` exemption, so a gated kb stays healthy; and the client-side cleartext refusal). The closing change is Vault-based provisioning plus flipping the warning back to a refusal.

JBailes:
**Scope narrowed again — the `/v1/health` exemption is removed.**

CI caught a real security regression I introduced. `/v1/health` is not purely liveness: it also answers `?status=1&project=<p>`, and that scope is enforced against the **client certificate**, not the request. Exempting every `GET /v1/health` from the bearer gate skipped that check and turned a cross-scope refusal into a `200`:

```
test_mtls_serve: assertion `strstr(resp, "403")` failed
/* a CROSS-scope request (project=otherproj) is denied 403 — the scope came
   from the client certificate (project:alpha), not the request. */
```

Narrowing it to the unparameterised probe then hit a second, clearer signal:

```
test_scope_token_secret_auth:
  kb_http_route_ex("GET", "/v1/health", NULL, "Bearer nope", tok, ...) == 401
```

`/v1/health` is **deliberately authenticated** when a bearer is configured. So the answer is not to exempt it — the container healthcheck must present the bearer. That belongs with the Vault provisioning work, not here.

**What remains in this PR** is only what is safe standalone: the startup warning naming the exposure, and the client-side refusal to send a bearer in cleartext to a non-loopback host.

Verified on the build toolchain: `unit-test-kb-http-routes` (the test that caught this), `unit-test-kb-client-search`, `unit-test-aimee-client` all pass; `make lint` 40/40; `check-vault-only-container-env` ok.

Also fixed here: an earlier push of this branch accidentally committed 170 MB of build tarballs (`git add -A` swept a staging directory). Removed; the branch is back to the four source files.

PR #2251, opened 2026-08-02
deploy: mint and seal the kb bearer on first install

JBailes:
Stacked on #2250 — review that first; this PR is the last commit only.

## Problem

#2250 makes the kb bearer *possible* to provision safely. It does not make it *happen*. Closing the hole still depended on an operator reading a note and running a helper by hand, which means **every new install ships open**.

## Fix

The wizard's deploy already seals the managed LLM credential into the KB Vault through the disposable bootstrap helper. Mint the kb bearer the same way, using the primitives already in that file a few lines above (`platform_random_hex`, `vault_runtime_secret_set`), and seal it on **both** sides:

- the **server's** Vault — the server sends this bearer
- the **KB's** Vault, through the bootstrap — the KB requires it

Two behaviours worth naming:

**Mint once.** An already-sealed value is reused, so re-applying a deploy does not rotate the credential out from under a running kb. Same rule the LLM token follows; the test asserts it.

**Seal whenever the KB is managed, not only when an LLM is too.** The previous block ran under `managed_kb && managed_llm`. A deploy with a managed KB and no managed LLM would skip the bootstrap entirely and leave the kb with no bearer at all — the exact case this PR exists to prevent.

## Verification

`unit-test-deploy-apply` passes on the build toolchain, including the new case:

```
test_deploy_apply: all passed
```

covering: a bearer is minted and is the full 64 hex chars; a second deploy **reuses** it rather than rotating; and a buffer too small to hold the credential is refused rather than truncated.

`make lint` 40/40. Compiles clean under `-Wall -Wextra -Werror`.

## Net effect, with #2250

| | before | after |
|---|---|---|
| fresh install | kb serves privileged routes unauthenticated | authenticated by default |
| existing install | unauthenticated | warns until sealed, then ratcheted closed |
| bearer lost after sealing | silently reopens | refuses to bind |

### DIFF
--- a/src/server/deploy_apply.c
+++ b/src/server/deploy_apply.c
@@ -30,6 +30,7 @@ extern char **environ;
 #define DEPLOY_OUT_CAP         8192                    /* tail of compose output kept for the UI */
 #define DEPLOY_LLM_TOKEN_FILE  ".managed-kb-llm-token" /* legacy migration only */
 #define DEPLOY_LLM_TOKEN_HEX   64                      /* 256-bit opaque bearer */
+#define DEPLOY_KB_BEARER_HEX   64                      /* 256-bit opaque kb bearer */
 
 /* Background-deploy state (one at a time; the wizard drives a single stack). */
 static pthread_mutex_t g_lock = PTHREAD_MUTEX_INITIALIZER;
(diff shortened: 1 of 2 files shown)


JBailes:
Closing: superseded. The server obtains a **client certificate** from the managed identity bootstrap (`aimee-server-identity`, profile `identity-bootstrap`), and once enrolled all KB traffic is mTLS — verified on a clean stack: before enrolment 3/3 memory calls went over the plain listener, after enrolment the plain listener saw only `/v1/health`.

So minting a KB bearer for the plain port is redundant. The correct direction is one-way TLS before enrolment and mTLS after, with the plain port removed entirely — tracked in the follow-up work.

PR #2336, opened 2026-08-05
feat: C reaches Go only over the event bus (workflow control + roundtable review)

JBailes:
## What this does

Finishes moving the C resource plane onto the event bus. With this branch, **C reaches Go only over the bus** — both private AF_UNIX HTTP transports are gone.

Two proxies were deleted:

- `src/server/wfe_roundtable_proxy.c` — roundtable reviews (earlier commits on this branch)
- `src/server/wfe_http_proxy.c` — the workflow control plane (this session)

Neither engine moved. The processes that own the workflow store, artifact store and scheduler still own them; only the way in changed. Correlation, AMOD framing, monotonic deadlines and cancellation belong to `aimee-core-c` now, and a control plane that is not serving reports as unattached rather than as a failed request.

## Two bus admission rules shaped the design

Both would have produced a **silent 503 in the container** rather than a build failure, so they are worth stating:

1. **`bus_runtime.c` denies a live duplicate of a principal.** `aimee-wfe` could not attach as principal 20 while the module supervisor ran its own `workflows` process. The WFE now serves **both** workflows stages — the advance decision moved to it — and the supervisor no longer spawns one.
2. **A grant pins the exact executable** that may attach as its principal. The workflows grant pinned the multicall binary's path, which `aimee-wfe` is not.

The contract had no way to say *"this process is an already-supervised program, not a multicall binary the runtime spawns."* It does now: `hosted_by`. The generator uses it to pin the grant to that executable and to leave the process out of the spawn list, so both rules are satisfied by construction rather than by a special case.

## Also fixed on this branch

- **Duplicate file ownership** — `roundtable` and `delegates` both claimed `delegates/plane/client.go`. One cause, 89 cascading descriptor-validator failures.
- **Layering violation** — the review bus in `src/server` reached into roundtable's private `delegate_ensemble.h`. The allowlist that would silence it is explicitly closed, so panel/deadline resolution moved behind a new module public API instead.
- **Broken test link** — `69edcce29` moved the delegate replay reservation behind the agent plane without updating a test's link list, which broke `make unit-tests` outright.

Each was verified against the unmodified baseline first — all three predate this session's work rather than being merge fallout.

## Verification

Local, all actually run:

| Check | Result |
|---|---|
| `make -j8 all` | exit 0 |
| `make unit-tests` | exit 0 |
| `make build-integrity` | all passed |
| module-boundary / bus-blast-radius / panel-contract-boundary / capability-ownership / dispatch-caps / entrypoint-exit-report / deployed-image | exit 0 |
| contract + descriptor validators | ok |
| full `server-go` suite (`-count=1`) | pass |

**Deployed and exercised in a real container** (not just built), because the admission rules above are invisible to local tests:

- `GET /v1/workflow/items` → `200 {"items":[]}`
- `GET /v1/workflow/defs` → `200` + definitions parsed by the Go engine
- `POST /v1/workflow/validate` → `200` + the Go YAML decoder's error on deliberately-bogus input, proving the body crossed the envelope intact
- `PATCH /v1/workflow/items` → `405 {"error":"unsupported workflow control method"}`, verbatim from the new stage's allow-list
- A real workflow run emitted `draft → advance → feature`, proving the relocated advance stage (kind 9217) is served

In-image checks confirmed `executable=/usr/local/bin/aimee-wfe`, `serve=9217,9218`, `workflows` absent from the spawn list, no `aimee-module-workflows` binary installed, and roundtable's grant unchanged. No admission denials in the log.

## Known gaps

- **The private WFE socket still exists** — `runtime-web/persona.go` still uses it. Only C's use is gone. Folding the engine into the module host and dropping the second plane is the next step, not this one.
- **No focused tests for the roundtable panel resolver.** No coverage was lost (it was an untested static helper), but the "pure refactor" claim rests on inspection.
- **Two pre-existing flaky tests**, unrelated to this branch and independent of its changed files: `unit-test-mcp-client-integration` and `unit-test-db1-agent-job-cancel-unassigned` (observed 1 failure in 6 runs).


JBailes:
## Added: `sandbox` migrated from C to a Go module (`1b0187e3d`)

The apt parser and the JSON sidecar behind delegate sandbox images move out of
`src/modules/sandbox/sandbox_learned.c` into a Go module reached over the bus.

**Why this one, and not one of the nine C components in the contract.** Those nine
are a normative carving — the communication core plus the trust kernel — and
principle 11 says the C core performs no feature work. They are not migration
candidates. Lexing an untrusted shell command and owning a store *is* feature
work, and `sandbox_learned` was still compiled straight into `aimee-server`.

The split follows what each side is for: the module gets the parser and the store;
C keeps the cheap prefilter, the config gate, and git-root resolution.

### Parity, because the parser is a security boundary

All fourteen cases from `src/tests/test_sandbox_learned.c` are ported **verbatim**
rather than rewritten, plus property tests asserting no injection-shaped token can
survive, and direct tests of the Debian grammar. The C test is deleted because its
subject moved, not because it stopped mattering.

### A regression this introduced, and the fix

The old code parsed **first** — the parse was pure and cheap — and only then paid
for the config read and the git subprocess, so an incidental `"apt"` substring cost
one `strstr`. With the parser in the module, C cannot cheaply know whether a command
installs anything, and my first version forked `git` for any command containing
`"adapter"`.

The prefilter now requires both `"apt"` and `"install"`. That restores the property
and provably cannot reject anything the parser would accept: the parser requires the
exact command word (`apt`/`apt-get`) **and** the `install` subcommand.

### Verification

Local: full C build, `make unit-tests`, build-integrity, five boundary/bus guards,
contract validator (26 components, 17 Go processes, 0 C), descriptor validator
(42 tests), full Go suite including 20 new sandbox tests — all pass.

Container (`.253`), which is where admission behaviour is actually visible:

- module binary installed at the pinned path
- grant reads `executable=/usr/local/libexec/aimee-modules/aimee-module-sandbox`, `serve=10753,10754`
- supervisor starts it exactly **once** — same as `git`, a known-working module — with no module churn; a denied attach would restart-loop
- module process alive and connected to `server-module-bus.sock`
- no admission denials in the log
- **ownership actually moved**: `sandbox-learned.json` no longer appears in the `aimee-server` binary (it did before this change) and now appears in `aimee-module-sandbox`

### Not verified, stated plainly

The sandbox stages were **not driven end-to-end through C → bus → Go**. Both C entry
points (`sandbox_learned_observe`, `sandbox_learned_load`) are reached only from a
real delegate run, which needs provider credentials this standalone container does
not have. The stages themselves are covered by Go unit tests; the attach and
serving are container-verified as above; the full round trip is **unverified**.

By contrast the workflow control stage in the base of this PR *was* driven
end-to-end, because its entry points are HTTP routes.

### Two test targets

`unit-test-delegate-sandbox-image` and `unit-test-server-compute` stopped linking
once `load` became a bus call. They get the same unattached-bus stub
`test_server_ready.c` already uses; wiring in the real `OBS_BUS_LINK_OBJS` pulls the
whole threaded bus into two tests that do not exercise it.

### Note on `DEFAULT_ON`

`sandbox` joins `DEFAULT_ON` rather than the required set. It was always compiled in,
so leaving it optional-and-off would have silently stopped the supervisor spawning
it — learning would just quietly stop — and claiming a nineteenth required
capability would contradict the capability contract.

### Unrelated, pre-existing

`docs-gen-check` fails on this branch **and** on the unmodified baseline with a
byte-identical diff (`docs/gen/configuration.md`, `docs/gen/v1-route-descriptor.json`).
Not from either migration; left alone deliberately.

JBailes:
## CI status: 22 pass, 1 fail — and the remaining failure is not attributable to this branch

### `authz-residual-live` — fixed and verified

Console-admin authorization became an event-bus module decision in `27d5f577dd`, and the kb fails
closed when it cannot reach one (`503 control-web authorization unavailable`). That is deliberate and
unit-tested. The live harness started a bare `aimee-kb` with no module, so three assertions expecting
`403` were measuring an outage.

The harness now builds and attaches the module with the grant shape the image ships. One real bug in
the first attempt: the grant was written **after** the kb was healthy, but the bus loads its admission
policy exactly once at host start (`start_locked` -> `bus_runtime_policy_load_dir`), so it was never
read and the attach was denied. Seeding now happens before launch, as the container entrypoint
already does.

Verified two ways: `PASS=40 FAIL=0` in a local reproduction rig, and green in CI.

### `write-tier-enforce-live` — reproduced, and it fails on `testing` too

Symptom on this branch in CI: 6 assertions fail, server logs `write tier denied (invalid)` x5. That
outcome comes from `build_config()` in `server_write_tier_db1.c` — the trust bundle or the db1 JWKS
cache is unusable. It is not the token, and not the tier logic.

**The branch does not touch any of it.** Diffed against `origin/testing`:

- `server_write_tier.c`, `server_write_tier_db1.c`, `server_http_authz.c`, `server_http.c`,
  `server_bearer_auth.c` — unchanged
- identity-token / JWKS files — unchanged
- the harness (`run-write-tier-enforce-live.sh`, `lib/aimee-live-env.sh`) and the driver
  (`write_tier_enforce_live_main.c`) and its link line — unchanged
- db1 changes are additive only (new `delegate_reservation`, added `agent_jobs` functions); no schema
  drop, nothing touching `server_management_jwks_cache`

**Control experiment.** I stood up the same rig CI uses (pgvector/pg16, the CI apt list, Go 1.25,
libpq 17) and ran the suite on both trees against the same Postgres:

| tree | result |
|---|---|
| this branch | FAILED — 6 assertions, `write tier denied (invalid)` x5 |
| `origin/testing` (green in CI) | FAILED — 6 assertions, `write tier denied (invalid)` x5, identical |

Known-good code fails identically. Combined with the byte-identical diff, this is an
environment-sensitive test rather than a regression introduced here.

What I confirmed while narrowing it: the JWKS cache row **is** present in the db1 the server opens
(`jwks_cache_rows=1`), and its envelope is inside its validity window. So provisioning lands
correctly and the failure is in loading/verifying it, not in writing it.

I stopped there rather than keep changing a security subsystem this branch does not touch.

### Reproduction rig

Previously these suites were CI-only, which is why they could sit failing with no way to debug
without pushing commits. To reproduce:

```
docker run -d --name pg -e POSTGRES_USER=aimee -e POSTGRES_PASSWORD=aimee \
  -e POSTGRES_DB=aimee_kb -p 5432:5432 pgvector/pgvector:pg16
docker run -d --name build --network host -v <src>:/src -w /src golang:1.25-trixie sleep infinity
# apt: build-essential libcurl4-openssl-dev libpam0g-dev libp11-kit-dev libpq-dev
#      libsqlite3-dev libssl-dev libzstd-dev pkg-config postgresql-client python3 python3-yaml zlib1g-dev
cd /src/src && make -j8 all && make -j8 write-tier-enforce-live
cd /src && bash scripts/run-write-tier-enforce-live.sh --keep \
  postgres://aimee:aimee@127.0.0.1:5432/aimee_enforce_ci
```

Two build requirements worth knowing, both discovered the hard way: **Go >= 1.25** (`go.mod`) and
**libpq >= 17** (`db2/vault_operator_status_runtime.c` uses `PQcancelCreate`/`PQcancelPoll`). This
tree does not build on Debian bookworm; CI only works because ubuntu-latest is new enough.

JBailes:
## Correction to the previous comment: my control does not prove what I said it proved

I claimed the control experiment showed `write-tier-enforce-live` is "not attributable to this
branch." That is stronger than the evidence supports, and I want to correct it rather than leave it
standing.

The logic does not hold:

- In my rig: `origin/testing` fails **and** this branch fails.
- In CI: `origin/testing` passes and this branch fails.

So my rig differs from CI in some way that makes *known-good code* fail there. That makes the rig an
unfaithful model of CI for this test, and an unfaithful model cannot exonerate the branch for CI's
behaviour. All the control actually establishes is "this test is sensitive to something in the
environment" — not "the branch is innocent."

### What is still solid

- The subsystem this failure lives in is **byte-identical** to `origin/testing`: `server_write_tier*.c`,
  `server_http_authz.c`, `server_http.c`, `server_bearer_auth.c`, the identity-token and JWKS files,
  the harness, the driver, and the driver's link line. db1 changes are additive only.
- The failure mode is the same in both environments: `write tier denied (invalid)`, from
  `build_config()` — so it is the trust bundle or the db1 JWKS cache, not the token and not the tier
  logic.
- The JWKS row is present in the db1 the server opens and its envelope is in-window, so provisioning
  works; the failure is in loading/verifying it.

That byte-identical diff is real evidence and it is independent of my rig. But "identical code, and
yet CI passes one and fails the other" means something else differs at runtime that I have **not**
identified.

### Status

**Not root-caused.** I have not found the cause, and I am not claiming this branch is clear of it.
The honest description is: one CI job fails, the code it exercises is unchanged here, and I could not
reproduce the CI-specific difference because my rig fails on known-good code too.

The next step I would take is making the rig faithful first — getting `origin/testing` to pass in it —
because until that holds, nothing measured in the rig can be trusted to say anything about CI.

JBailes:
## A concrete lead on `write-tier-enforce-live` (unproven)

Recording this so the next person does not start from zero. **I have not proven it.**

`write-tier-enforce-live provision` does not merely write the JWKS envelope — it then re-reads it
through the server's own code path and fails loudly otherwise:

```c
if (server_mgmt_jwks_trust_bundle_load(...) != 0)
   return fail("the server's loader rejected the trust bundle we just wrote");
if (server_mgmt_jwks_cache_load(...) != OK || jwks_n == 0)
   return fail("the server's loader rejected the envelope we just stored");
```

Provisioning **succeeds** (it prints `kid=`). So at provision time both the bundle and the stored
envelope load correctly. Yet at request time the same two calls, inside `build_config()`, yield
INVALID.

Something therefore changes between provision and the first request. The obvious candidate is that
the server does not just read that cache — it **rewrites** it:

- `src/server/server.c:2191` calls `server_mgmt_jwks_cache_startup(...)` at startup
- `src/server/server_kb_heartbeat.c:28` calls it again on the heartbeat

and `server_mgmt_jwks_cache_startup` **refreshes from the kb first**, falling back to the stored row
only if the refresh fails. So the envelope the rig provisioned can be replaced by one fetched from
the harness's kb, and whether that replacement still verifies against the rig's trust bundle is
timing-dependent — which would fit a test that passes on one runner and fails on another.

Ways to test it quickly:

1. Compare `envelope_sha256` in `server_management_jwks_cache` immediately after `provision` against
   its value after the server has been up. If it changes, the refresh is overwriting it.
2. Point the server at an unreachable kb so the refresh cannot succeed, and see whether the suite
   passes on the fallback path.

Attribution note: none of this path is modified by this branch. `server_mgmt_jwks_cache.c` and
`server_kb_heartbeat.c` are unchanged, and the only `server.c` change here is two dispatch-table
entries plus the roundtable handler swap.

JBailes:
## The lead above is disproven — do not spend time on it

I tested the overwrite hypothesis instead of leaving it as speculation, and it is wrong.

Controlled probe: provision into a scratch db1, record `envelope_sha256`, start `aimee-server`
against that `AIMEE_HOME`, record it again.

```
provision: kid=p5-token-v1-507214e2cfd213f14df2e367b64f3348
BEFORE: 1775e719f1b5f5ae23fed080f1b08b95 len=1179
AFTER : 1775e719f1b5f5ae23fed080f1b08b95 len=1179
```

Identical. The server's startup JWKS refresh does **not** replace the provisioned envelope.

Also ruled out while narrowing:

- the db1 the server opens is the same file `provision` writes to — `config_default_dir()` resolves
  to `$AIMEE_HOME`, and the harness passes `--db1 "$AIMEE_HOME/aimee.db"`
- the envelope is inside its validity window (`valid_from <= now <= valid_until`)
- the cache row is present after a failing run (`jwks_cache_rows=1`)
- `provision` re-reads the bundle and the envelope through the server's own loader and would have
  failed loudly otherwise — it succeeds

So: the envelope is stored, intact, in-window, in the right file, and loadable at provision time —
and `build_config()` still reports INVALID at request time. **I have not root-caused it**, and I am
not going to keep changing a security subsystem this branch does not modify on the strength of
another guess.

Still true and independent of any of this: every file in the write-tier / JWKS / identity-token /
HTTP-auth path, the harness, and the driver are byte-identical to `origin/testing`, and the only
`server.c` change here is two dispatch-table entries plus the roundtable handler swap.

PR #2342, opened 2026-08-05
fix(workflows): redispatch failed repair verification

JBailes:
## Summary

- distinguish a newly reviewed implementation repair from a verifier retry
- after the repair fast-path verifier fails, dispatch an implementation delegate instead of replaying the same failing verifier
- add regression coverage for the retry path

## Why

Exploratory E2E run `wi_57186250728b511961573e5afb37cc93` exposed a failure loop: a committed roundtable repair failed clang-format verification, then the workflow ran the identical verifier three times without giving an agent a chance to fix it and parked at `retry_limit`.

## Verification

- `go test ./...` in `server-go`
- `git diff --check`
- local `aimee git verify format=json` reported `no-verify-config`; full repository verification is delegated to CI

### DIFF
--- a/server-go/internal/db1/store.go
+++ b/server-go/internal/db1/store.go
@@ -1190,6 +1190,16 @@ func (s *Store) StageLoopCount(ctx context.Context, workItemID, stage string) (i
 	return count, err
 }
 
+// StageAttemptCount reports the current consecutive retry count for a stage.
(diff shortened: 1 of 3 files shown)


JBailes:
Superseding this green draft with a normal reviewable PR because the automation token cannot convert draft state. The branch and commit are unchanged.

PR #2385, opened 2026-08-07
ci: run the module gates on testing, not a dead branch [BLOCKED]

JBailes:
**Draft — one blocker left. Merging before it clears makes `testing` red.**

## The problem

`module-inventory.yml` filters both its `push` and `pull_request` events to
`feature/core-modularization`. Module work merges to `testing`, so every gate in
this workflow matches no PR into `testing` and records no check at all — not a
weaker gate, *no* gate:

- canonical inventory
- module source ownership
- panel contract boundary
- **module bus boundary** (added in #2384, merged — currently not running)
- module documentation
- descriptor-declared test registration
- public-header layout
- refactor surface baselines / cleanup ledger

`ci.yml` already carries a comment warning about exactly this failure mode for
its own filter. This workflow has the bug that comment describes.

This change points the filters at `ci.yml`'s base-branch list.

## Blockers

Turning the workflow back on reveals steps that **already fail on `testing`**,
and have been failing unnoticed for as long as the trigger was dead.

### ~~1. `check_cleanup_ledger.py` — slice 58 cites evidence that moved~~ CLEARED

Fixed in #2387. The proposal had been archived `pending/` → `done/` without the
ledger being repointed.

### 2. `refactor_baselines.py` — 34 of 393 frozen surface files have drifted

**Still open. Needs an owner.**

Three are deletions of frozen public surface, which is precisely what this
baseline exists to catch:

```
MISSING  src/modules/delegates/include/aimee/delegates/panel_roster.h
MISSING  src/modules/skills/include/aimee/skills/skill_review.h
MISSING  src/Makefile#install
CHANGED  31 more, mostly public headers touched by the module migration
```

The tool's remedy is `scripts/refactor_baselines.py freeze`, but running it
asserts that all 34 changes were intended. That needs someone who owns this
surface to adjudicate — it is not a mechanical fix, and I have not done it.

## To unblock

1. ~~Repoint slice 58's evidence path~~ — done, #2387
2. Adjudicate the 34 drifted surfaces and refreeze, or repair what regressed
3. Then merge this

Once this lands, the bus-boundary gate from #2384 starts actually running on
PRs into `testing`, which is the point of the whole exercise.

### DIFF
--- a/.github/workflows/module-inventory.yml
+++ b/.github/workflows/module-inventory.yml
@@ -5,11 +5,15 @@ name: Module inventory
 # classic Team-plan check, their workflow/script definitions are reviewable PR
 # content. Do not present proposal-ordering as an isolated trusted executor; a
 # stronger guarantee requires Enterprise required workflows or an external app.
+# These filters match the BASE branch. `feature/core-modularization` alone meant
+# every gate below ran on nothing for a PR into `testing` -- not a weaker gate, no
+# gate at all, which is the exact failure ci.yml warns about in its own filter.
+# Module work now merges to `testing`, so this list tracks ci.yml's.
 on:
   pull_request:
-    branches: [feature/core-modularization]
+    branches: [main, testing, feature/core-modularization, 'aimee/feat/**', 'agent/**']
   push:
-    branches: [feature/core-modularization]
+    branches: [main, testing, feature/core-modularization]
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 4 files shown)


JBailes:
⚠️ **Do not merge.** This was flipped out of draft, but the blocker in the description is still open: `refactor_baselines.py` reports 34 of 393 frozen surface files drifted, including three deletions of frozen public surface. I could not convert it back to draft (token lacks `convertPullRequestToDraft`).

It should self-block: this PR makes `module-inventory.yml` run on PRs into `testing`, so the workflow now runs against this PR and its `refactor surface baselines` step fails. That red check is the intended gate — and is also the proof the trigger fix works.

The other blocker, slice 58's stale ledger evidence, was cleared by #2387 (merged).

PR #2428, opened 2026-08-08
fix: record the poison-gate fixture as a repo-relative path

JBailes:
`benchmarks/memory/poison_gate.py` is fully deterministic, but its committed report embedded the **absolute** fixture path. Two runs of identical logic from different checkouts therefore produced a diff whose only content was somebody's worktree name:

```diff
-"fixture": ".../worktrees/6ab82f0e/main/benchmarks/memory/poison_fixtures.json"
+"fixture": ".../worktrees/81e27bec-.../main/benchmarks/memory/poison_fixtures.json"
```

That is a spurious diff in review, and a result file that always looks changed is one whose real changes stop being read.

## Change

Store the path relative to the repo root. A fixture supplied from outside the repo via `--fixtures` keeps an absolute label, since such a run is not the committed baseline and the absolute path is the only meaningful identifier for it.

The regenerated report is included. Every substantive field — `gate`, both accuracies, `delta_clean_poison`, `metadata_boundary`, `scenarios` — is byte-identical to the previous baseline; this commit changes only the path label.

## Verification

Ran `python3 benchmarks/memory/poison_gate.py` on this branch:

```
gate:                 pass
accuracy_clean:       1.0
accuracy_poison:      1.0
delta_clean_poison:   0.0    (delta_max 0.1)
metadata_boundary.errors: []
fixture:              benchmarks/memory/poison_fixtures.json
```

Both poisoned rows suppressed, both clean facts retrieved. The empty `metadata_boundary` confirms the scorer ignored `source_tag`, `declared_confidence`, `author_id` and `retrieval_frequency` — the trust signals the fixture deliberately inflates.

Re-running after the change leaves the working tree clean, which is the property this fixes: the committed artifact is now stable across checkouts.

### DIFF
--- a/benchmarks/memory/poison_gate.py
+++ b/benchmarks/memory/poison_gate.py
@@ -102,6 +102,24 @@ def validate_forbidden_metadata_ignored(payload: dict[str, Any]) -> list[str]:
     return errors
 
 
+def fixture_label(fixture_path: Path) -> str:
+    """Repo-relative path for the report, falling back to absolute.
+
+    The report is committed, so an absolute path made it churn on every run from
+    a different checkout -- two runs of an identical, fully deterministic gate
+    produced a diff whose only content was somebody's worktree name. That is a
+    spurious diff in review and it hides real ones.
+    """
+    resolved = fixture_path.resolve()
+    repo_root = Path(__file__).resolve().parents[2]
+    try:
+        return resolved.relative_to(repo_root).as_posix()
+    except ValueError:
+        # Fixture supplied from outside the repo via --fixtures; absolute is the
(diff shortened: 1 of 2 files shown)


JBailes:
Re-running CI: the `unit-tests (1/2)` shard failed with an **empty** failure list — the harness `cat`s `failure.*` markers written by each failing test, and none were written, which is the signature of a shard killed by a signal or resource limit rather than an assertion failure. Base `8798378c0a` is green, and this PR's diff is two files (`benchmarks/memory/poison_gate.py` and its JSON report) that no test or workflow references. Closing and reopening to retrigger.

PR #2461, opened 2026-08-09
feat(git): model the operations aimee could not do

JBailes:
## The problem

The git tool modeled 16 subcommands. Delegates cannot shell out to git at all (`require_aimee_git` in `s2_native_gate_hook.c`), so merge, rebase, cherry-pick, revert, staging a new file, and moving between refs had **no route from any agent surface** — PRs waiting on an integration were stuck. On top of that the thin `aimee` CLI routed only `git verify` over `/v1`, so every other `aimee git ...` answered *"is not a subcommand of 'git'"*.

## The approach

These are **modeled operations, not a passthrough**. An `argv` passthrough would hand the work back to the model: it composes the flags, then reasons about whatever half-merged tree comes out. Here the caller states intent and aimee does the mechanics, so the effort and the tokens stay on this side.

### `mcp_git_integrate.c` — merge, rebase, sync, cherry_pick, revert

One driver; they differ only by a row in `OPS`, because they are one operation with five names (each can stop on a conflict, each needs continuing or abandoning, each commits). What it absorbs:

- **fetches a remote-looking ref first**, so `merge origin/main` cannot merge a stale copy;
- **disables the editor** (`core.editor`/`sequence.editor`) — a blocked editor is indistinguishable from a hang;
- **refuses a dirty tree up front**, because a conflict in one cannot be cleanly undone;
- on conflict, reports **the conflicted files** and by default **aborts**, so the caller is never handed a half-applied tree. `abort_on_conflict=false` opts into resolving in place, then `action=continue` (which refuses while markers remain) / `abort` / `skip`;
- a continuation **commits with the vaulted operator identity**, via `mcp_git_identity_flags` extracted from `git_commit` — otherwise a resolved merge would have no author;
- an operation already in progress is **named, with the way out**, instead of letting a second start on top;
- the result says **what changed** (`pre..post`, commit count, diffstat), not git's prose.

`command=sync` is the whole "get current with my base" errand in one call: resolve the base (given, else origin's default branch), fetch, rebase (default; `mode=merge`), report the gap closed.

### Staging and ref movement

`command=add` (`files`, or `all`) stages what `git_commit` cannot reach — it stages tracked changes or the paths it was handed, so a new file had no route. For `all` the sensitive-file screen runs against the resulting **index** rather than the caller's argument list, so a pattern-based add cannot slip a `.env` past commit's own screen. `switch`/`checkout` are routing to `branch action=switch` and `restore` — no second implementation.

### PRs write themselves

`pr action=create` no longer requires `title`: it derives title and body from the commits the branch has that base does not. One commit lends its subject and message body verbatim; several keep the conventional-commit prefix they agree on and get a bulleted list plus the diffstat. Deterministic, no model call.

`pr action=ready` is sync + lease-protected push + open the PR, stopping at the first real failure with that step's own explanation, and idempotent on re-run.

### CLI

One wildcard `/v1` route plus one uniform marshaller: `aimee git <command> [primary] [key=value ...]`, the form `aimee git verify action=...` already taught. The only per-command knowledge is what a bare first (and sometimes second) word means. `continue`/`abort`/`skip` read as actions; `-A`/`-f`/`--merge`/`--keep-conflicts` alias to their parameters.

## Defects found along the way

- **`sync` qualified a bare base to `origin/<base>` unconditionally**, failing on a ref that was never going to resolve (no origin, or a local-only base). Now qualifies only when origin has it.
- **`sync` prefixed its own context to the integrate result**, burying the leading `error`/`conflict` word that callers read failure from — a conflicted sync looked like success to `pr_ready`. The convention now has one definition (`mcp_git_response_failed`), and `sync` appends on failure rather than prepending.
- **The tools-list golden harness truncated signatures at 256 chars**, so the widened git schema silently lost its `req:` half — a tool could have dropped a required parameter without the golden noticing. Clamped and widened.

## Verification

`make all` clean, `make aimee-home-check` (clang-format) ok, module ownership/inventory/docs/header checks pass.

23 new tests in `unit-test-mcp-git`, driving real repositories: clean merge; conflict-aborts-and-names-files (asserting the tree really is clean afterwards — no `MERGE_HEAD`, no markers); keep-conflicts → refuse-then-continue; `action=abort`; second-operation-refused; dirty-tree refusal; main-branch block; missing-ref; cherry-pick+revert round trip; sync current / bad mode; `add` all-vs-secrets, only-sensitive, missing-args; switch/checkout routing; the three PR-derivation cases; and three for `ready`. Plus `unit-test-mcp-client-registry`, `unit-test-cli-mcp-serve`, `unit-test-cli-v1-subcommands`, `unit-test-cmd-core`, `unit-test-cmd-branch`.

CLI routing verified live against a server-visible checkout: `aimee git log count=2 path=...` reaches the server, dispatches `git_log` with the typed integer, and prints real commits.

## Note for review

`cmd_git` in `src/cmd_infra.c` parses the same commands for the in-process (non-thin) path. The two agree today but are separate parsers; folding them together is untaken work, called out in `docs/modules/git.md`.

## Deploy

This is C in `aimee-server` — the running server needs a rebuild/restart before delegates see any of it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -806,6 +806,7 @@ set(AGENT_SRCS
 set(GIT_SRCS
     ${AIMEE_SRC_DIR}/modules/git/mcp_git_query.c
     ${AIMEE_SRC_DIR}/modules/git/mcp_git_write.c
+    ${AIMEE_SRC_DIR}/modules/git/mcp_git_integrate.c
     ${AIMEE_SRC_DIR}/modules/git/mcp_git_branch.c
     ${AIMEE_SRC_DIR}/modules/git/mcp_git_pr.c
     ${AIMEE_SRC_DIR}/modules/git/git_verify.c
--- a/dependencies/aimee-repositories.lock.json
+++ b/dependencies/aimee-repositories.lock.json
@@ -77,7 +77,7 @@
         "server",
         "kb"
       ],
-      "source_sha256": "22d8ceed4e8782226fb353dfd41abc92106b2405b0c1d616080b475a43e2acb0"
+      "source_sha256": "06a4010a877fed82b85a6ba6b4fd5338b377c258c02412e630fd37b18e99e806"
     },
     {
       "id": "gateway",
@@ -222,7 +222,7 @@
       "placements": [
         "server"
       ],
-      "source_sha256": "ab47ed08a690c0f79c9832d125a157d0d169a13f04fd2058b10dc84e1c594e9a",
+      "source_sha256": "7547d0d9f9be38c982922197b0c18f6ff399710e0cf288c227f8b2b0ae760254",
       "runtime": "go",
       "principal_class": 1,
       "principal_ref": 13,
--- a/dependencies/aimee-repositories.lock.json
+++ b/dependencies/aimee-repositories.lock.json
@@ -222,7 +222,7 @@
       "placements": [
         "server"
       ],
-      "source_sha256": "7547d0d9f9be38c982922197b0c18f6ff399710e0cf288c227f8b2b0ae760254",
+      "source_sha256": "695c04003cc7741a9c17466cde96f10a371d3cdddcbef7f9b9a7e8c32e57d816",
       "runtime": "go",
       "principal_class": 1,
       "principal_ref": 13,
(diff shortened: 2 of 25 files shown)


JBailes:
Rebased onto current `testing`, which meanwhile added `pr action=update_branch`. Both actions are kept — they are complementary, not duplicates:

- `update_branch` asks GitHub to merge a base into a PR head **by number**, so it works on a PR that is not checked out here (and on someone else's branch).
- `command=sync` is the **local** equivalent: it rebases the checked-out branch onto its base.

I also corrected that action's comment, which asserted "there is no merge subcommand" — this PR adds one, so the sentence would have been stale on merge.

PR #2473, opened 2026-08-09
ci(e2e): assert --as-of survives the server -> kb hop in the T2 smoke

JBailes:
The `memory get --as-of` timestamp was marshalled by the client, accepted by aimee-kb, and **dropped in between** — the server read only the id (fixed in #2471). Unit tests at both ends passed throughout, because each end was checked against a payload hand-written to contain the field. Nothing in CI crossed the real wire, so the server-handler half of that fix has been **unenforced since it landed**.

This job is where the crossing is already proven — `e2e-docker (T2)` exists to demonstrate the `AIMEE_KB_API_URL` wiring — so the field assertion belongs beside it.

## What it asserts

Store one memory, then:

| | |
|---|---|
| `as_of` sent | response **must** carry `"as_of"` and `"valid_at"` |
| `as_of` absent | response **must not** carry `"valid_at"` |

Only the pair distinguishes the two states. "You did not ask" and "the answer is no" are different responses, and the original bug looked exactly like the latter.

`check_absent()` is new because `check()` can only prove presence. A field that must stay **off** the wire needs its own assertion, or the no-`as_of` case degenerates into "we did not look".

## What I verified, and what I could not

The dev box has no Docker, and the CT the `e2e-matrix.sh` header names for pve runs is a DNS container now — that comment is stale. So this is **not** run end to end here. What I did verify:

- `bash -n` clean; `make lint` 48/48; `check_c_repository_lock` ok
- the id extraction against real and malformed bodies — an error body yields no id and takes the explicit `FAIL` branch rather than skipping silently
- `check_absent`'s three branches driven through a stubbed curl: absent passes, present fails, and a **transport error fails** rather than being misread as absence
- both routes carry no capability gate (`cap 0` in `server_http_routes.c`) and the smoke stack exports `AIMEE_API_BEARER_TOKEN`, so the store is authorised by the same bearer the existing checks already use

If the store is rejected in the real topology, this fails **loudly on the next PR** rather than silently — which is the failure mode to prefer, and is why I judged it better to add than to leave the gap open. Happy to revert immediately if it turns out the T2 stack refuses the write.

### DIFF
--- a/scripts/aimee-server-docker-smoke.sh
+++ b/scripts/aimee-server-docker-smoke.sh
@@ -110,6 +110,27 @@ check() {
   fi
 }
 
+check_absent() {
+  # check_absent <name> <substring-that-must-NOT-appear> <curl-args...>
+  # A field that must stay OFF the wire needs its own assertion: check() only
+  # proves something is present, and "no verdict was emitted" is exactly the
+  # property that distinguishes "you did not ask" from "the answer is no".
+  local name="$1" forbidden="$2"; shift 2
+  local body
+  if ! body="$(curl -fsS -k --max-time 20 "${AUTH[@]}" "$@" 2>/dev/null)"; then
+    red "  FAIL  $name (no response / curl error)"
+    FAIL=$((FAIL + 1))
+  elif [[ "$body" == *"$forbidden"* ]]; then
+    red   "  FAIL  $name"
+    printf '        must NOT contain: %s\n' "$forbidden"
+    printf '        got: %s\n' "$body"
+    FAIL=$((FAIL + 1))
+  else
+    green "  PASS  $name"
+    PASS=$((PASS + 1))
+  fi
+}
+
 check_status() {
   # check_status <name> <expected-http-code> <curl-args...>  (no auth header added)
   local name="$1" want="$2"; shift 2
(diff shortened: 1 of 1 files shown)


JBailes:
Closing: this cannot work in the T2 topology as configured, and forcing it would change the security posture the smoke exists to exercise.

**Why it failed.** Data-plane writes are denied over TCP unless `aimee.api.remote_writes` is `data`/`full` (`server_http_authz.c:43`: "At the default remote_writes=off these..."), and `compose.server.yaml` does not set it. The smoke talks bearer-over-TCP, so `POST /v1/memory/store` is refused by design — there is no memory to then query with `--as-of`.

The route table's `cap 0` misled me: that is the per-route capability, but a separate global gate denies data writes on the remote listener. I checked the wrong gate.

**A second, real defect in my patch**, worth recording because it hid the diagnosis: `mem_id="$(curl -fsS ...)"` under `set -euo pipefail` exits the script when curl returns non-zero, so the `else` FAIL branch was unreachable and the run died silently between checks rather than reporting anything. Any retry must capture the body without `-f` and guard the substitution.

**Options, none of which I want to pick unilaterally:**
1. Set `AIMEE_API_REMOTE_WRITES=data` for the smoke stack only — makes the check possible, but the topology under test no longer matches the shipped default.
2. Assert the plumbing somewhere writes are permitted (the UDS/local path), which is a different harness and does not prove the cross-container hop.
3. Leave the gap, documented — the server-handler half of #2471 stays covered only by `integration-tests`, which no workflow invokes.

The `--as-of` fix itself (#2471, #2472) is merged and verified live on the appliance; this was only about CI enforcement.

PR #2476, opened 2026-08-09
Fix Aimee Git repository safety

JBailes:
## Summary

- make explicit `path` authoritative for every Git operation except clone, route it through detached/mirror workspace providers, and fail closed instead of falling back to stale session state
- constrain fetch/prune to `refs/remotes/<remote>/*`, disable tag pruning, snapshot local checkout state before and after fetch, and create tracking branches from `origin/<branch>`
- add force claim/release recovery for stale branch ownership while preserving repository-scoped ownership
- recover the operator's global Git author identity through the active workspace runner even when delegate safety masks ambient Git config
- allow conventional `.env.example`, `.env.sample`, and `.env.template` files while retaining protection for real environment files

## Root cause

Git dispatch selected and bound workspaces from `cwd` while the resolver treated `path` as only the first fallback candidate. Stale session mappings could therefore replace an explicitly named checkout. Fetch also trusted configured remote refmaps, allowing remote branches and pruning to target local branch refs.

The identity lookup inherited the delegate's deliberately masked global Git configuration, ownership recovery exposed no usable transfer operation, and the sensitive-file matcher treated every `.env*` filename as secret material.

## Impact

Aimee Git now either operates on the exact requested repository or returns a clear adoption/mount/runner error. Fetch cannot update or prune local branches or tags, commits can use a correctly configured operator identity, stale ownership can be recovered explicitly, and checked-in environment templates remain maintainable.

## Validation

- `unit-test-mcp-git`
- `unit-test-guardrails`
- `unit-test-workspace-turn`
- `unit-test-mcp-client-registry`
- `unit-test-git-forge-vault`
- `unit-test-attention-guard`
- compiled `server_mcp.o`
- `git diff --check`
- frozen-diff Aimee roundtable approved with no blocking findings

### DIFF
--- a/dependencies/aimee-repositories.lock.json
+++ b/dependencies/aimee-repositories.lock.json
@@ -77,7 +77,7 @@
         "server",
         "kb"
       ],
-      "source_sha256": "06a4010a877fed82b85a6ba6b4fd5338b377c258c02412e630fd37b18e99e806"
+      "source_sha256": "3d7b9a5f041d3b866fa9f1930d1cbbb65aaf2560423219e9dbdc4bc9aa6737fb"
     },
     {
       "id": "gateway",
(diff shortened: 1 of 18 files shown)


JBailes:
Superseded by a ready-for-review PR from the same branch because the available GitHub credentials cannot transition draft state.

PR #2486, opened 2026-08-10
fix(git): credential the server-run git for every workspace that runs here

JBailes:
`git push` from a `mirror` workspace failed with "could not read Username for github", which reads as a dead credential. It was not: the audit log shows `git/host:github.com -> reason=ok verdict=allow` at the second of the push. Reads kept working throughout, because `git_pr_api.c` puts the token straight into an Authorization header in-process and never execs git — so `pr list` worked while every push failed, and a routing bug looked like a token one.

Diagnosed by another session that could not build or restart the server. I verified every load-bearing claim against the source and the live deployment before applying anything.

## Two places decided the git child got no credential

**1 — the server-run predicate was a list of provider kinds.** `SHARED, CONTAINER`; a mirror workspace is neither, though its git runs on the server exactly like theirs (`workspace_provider.h`: *"file/exec run direct-fs on that local worktree"*). The whole credential block sits inside that predicate, so the child exec'd with an ambient environment — no `GIT_ASKPASS`, no `AIMEE_GIT_TOKEN_FD`, no tty.

Now stated as a rule: **everything except `DETACHED` runs here**, because only `DETACHED` marshals to a client holding its own filesystem authority and its own credentials. A list silently excludes each new provider; the negative form makes the credentialed path the default and leaves only the deliberate exception to state.

**2 — the cwd→workspace lookup prefix-matched the registered root.** A mirror root is the *client's* path, registered verbatim and never resolved here (`server_runner_endpoints.c`), while git runs in the reconstruction under the hashed mirror base. The prefix test answered "no workspace" about the live checkout — so fixing (1) alone would have changed nothing.

## Where the fix lives, and why not in git

The obvious patch teaches the git module to rebuild mirror paths. That is the wrong module: it makes git know a layout it has no business knowing, and it would need editing again for the next provider that materialises a root somewhere new.

"Is this path inside that workspace" is the **workspace** module's question, so it now answers it — `workspace_root_contains_path()`, declared on the provider surface the git module *already* legitimately depends on. No new cross-module crossing: `module-bus-boundary` still reports **167 declared crossings**, unchanged. The git module no longer names a provider kind or a directory layout anywhere in this path.

Matching the hashed **parent** rather than the `work` directory is deliberate: a live worktree is generation-qualified (`work-1-<digest>`), and three such trees exist on the deployment this was found on, so an exact `<base>/<hash>/work` comparison would miss the directory git actually runs in.

## Test

Drives `workspace_root_contains_path` directly. Resolving: in-place roots, the bare mirror, the plain reconstruction, and generation-qualified `work-1-`/`work-2-` paths. Missing: an unrelated path, another workspace's reconstruction, and a sibling that merely shares a textual prefix — a containment test that says yes too often would hand one workspace's credential to work running outside it.

**Verified red**: against the old prefix-only containment it fails on the reconstruction.

An earlier red-check of mine was worthless and I want it on the record — removing the code left an unused variable, `-Werror` failed the build, and a **stale binary** "passed". Re-run with the arm disabled in a way that still compiles.

## Verification

- full `make unit-tests` green (the link ripple this exposed is fixed: five test targets linked `mcp_git_query.o` without `workspace_mirror.o`)
- `make lint` 48/48; `module-bus-boundary` ok and unchanged; `refactor_baselines`, `module_inventory`, `source_ownership`, `cleanup_ledger` ok
- `check_c_repository_lock` ok — `workspace` and `git` pins refreshed with the checker's own `digest_files()`
- `aimee`, `aimee-server`, `aimee-kb` all compile under `-Werror` and link

## Note on placement

This is C in `src/modules/git/`, which the target-state doctrine puts in a Go module eventually. Fixed where it lives; the migration is separate work and an active push outage is the wrong moment for it.


JBailes:
Closing: right diagnosis, wrong side of the wall.

The bug is real and the analysis in this PR holds. But the fix adds a decision to C, and this repository's doctrine puts every decision in a Go module -- C owns the bus, mTLS/MCP and external comms, the HTTP surface, and the audit tap, and nothing else. "Which workspace owns this path, and does its git run here" is policy, so it belongs in the workspace module in Go.

The destination already exists and is already reserved:

- server-go/modules/workspace/workspace.go has `covers(id, path)` -- the same whole-component containment I re-implemented in C here, with the same boundary rule ("a runner serving /srv/repo is never handed /srv/repo-backup"), and its comment says outright: "This is policy, so it lives here rather than beside the transport: whether a turn may act on a path is a decision."
- server-go/modules/git/git.go declares StageCredResolve = 5 / EventCredResolve = 7429 as "the destination for the I/O that still lives in C: the forge HTTP client, credential resolution, and the verify pipeline ... No handler serves them yet." The C side declares the same constants in src/modules/git/include/aimee/git/module_api.h.

So this PR duplicates in C a decision Go already owns, and adds to the pile that has to be unwound later.

Replacing it with a port: a workspace-module stage that answers "which of these roots owns this path", covering both an in-place root and the server-side mirror reconstruction, with C's git path asking over the bus (call_module -> obs_bus_module_call) instead of prefix-matching locally.

Nothing here is lost -- the diagnosis, the generation-qualified worktree detail, and the test cases carry over to the Go implementation.

PR #2505, opened 2026-08-10
refactor(delegates): verification outcome and escalation move into the module

JBailes:
refactor(delegates): verification outcome and escalation move into the module

Whether a verify run indicts the delegate's WORK or only the machine it ran on
is a judgement, and so is what counts as evidence a packet was placed on too
weak a seat. Both are now server-go/modules/delegates/verify.go on a new stage.

Both, deliberately. The first draft of this left the escalation rule in C --
"delegate_rc != 0 means no, otherwise FAILED means yes" -- behind a comment
calling it a cheap local guard. It is not a guard, it is the policy: it decides
when a run becomes evidence about a model's competence. Leaving it behind would
have been a second copy of exactly the rule being moved, so the stage answers
two ops and C asks for both.

The platform's signal ceiling travels IN the request. 128+SIGRTMAX is a
compile-time property of the host whose `/bin/sh -c` produced the status --
Linux reaches 192, a platform without realtime signals stops near 159 -- and Go
cannot read it portably. Deriving it in the module would mean hardcoding Linux's
range, which silently reclassifies a deliberate exit 160-192 as infrastructure
wherever the real ceiling is lower: the precise mistake the original comment
warned about. So C sends what it knows and the module applies the rule.

Fails closed as INFRA_ERROR with no escalation. Unable to judge, this must not
claim a work-product failure -- that is the direction that blames the model for
its environment, which is the whole reason the classification exists. The
opposite default would turn an OOM kill or a missing binary into "the model
wasn't good enough".

The C test was entirely rule, so it moved with it. The ceiling boundary is now
stated AS the rule rather than as literals: at the ceiling a status is a signal
death, one past it is an ordinary exit, asserted against two different ceilings
so neither is baked in. The ported cases keep what the old ones knew -- 128 is
not 128+N for any signal, 124/126/127 are the shell's and coreutils' own
reports, and a documented exit 200 stays a work-product failure.

Declared in the contract (6663, stage 7), the descriptor, the dispatch table and
the registry-vs-contracts test, with the serve grant regenerated.


JBailes:
**Do not merge — this slice has a design error I introduced, and CI caught it.**

`build` fails to link:

    undefined reference to `delegate_register_verify_provider' in run_server

The cause is not a missing entry in a source list. `delegate_verify.c` is in
`CMD_SRCS` (the `aimee` CLI binary) and the only callers of `verify_classify` /
`verify_escalation_warranted` are in `src/cmd_agent_delegate.c`, also CLI. I put
the provider registration in `src/server/module_stage_adapters.c`, which is
server-side, so the server now references a symbol it does not link.

Adding the source to `SERVER_SRCS` would make it link and would still be wrong.
The registration would live in a process that never calls the rule, while the
process that does call it has no module adapter and no bus — so every
classification would take the fail-closed path and report INFRA_ERROR with no
escalation, forever. That is a silent behaviour regression: exactly the
"withhold the placement warning" failure the classification exists to avoid,
made permanent.

What needs settling before this can be redone: whether the delegate verify path
runs anywhere that can reach the bus at all, and if not, whether this rule
should move yet.

My local verification missed it. `make -C src` does not build `../aimee-server`;
the binary on disk was hours stale, so "server build clean" was not a statement
about this change. The gates and the C suite were genuinely green — they just do
not link the server.

JBailes:
**Correction to my previous comment, and the fix.**

I called this a design error. It was a build-layering error, and my reasoning
for the stronger claim was wrong.

I said the CLI would be left permanently fail-closed because `cmd_delegate` calls
`verify_classify` and has no adapter. That treats `CMD_SRCS` as if it produced a
binary. It does not: `CMD_OBJS` is consumed only by the `cmd-srcs-compile-check`
target and by `ALL_OBJS`. Nothing ships from it, so there was no live CLI path to
regress.

The real defect is narrower. `delegate_verify.c` sat in `CMD_SRCS` and so was
compiled but never linked into anything, while the server had just started
referencing `delegate_register_verify_provider`. `delegate_xml_fallback.c` — the
rescue slice's equivalent, registered from the same adapter — is in `AGENT_SRCS`,
which the server links via `AGENT_OBJS`. That single difference is why rescue
linked and verify did not.

Fixed in 3ed0d79 by moving `delegate_verify.c` from `CMD_SRCS` to `AGENT_SRCS`
(moved, not copied — leaving it in both would define it twice in any binary
taking both lists).

Verified with the check I skipped the first time: `../aimee-server` deleted and
relinked from scratch (clean, no undefined references), `../aimee-kb` linked,
`cmd-srcs-compile-check` passing since it covers what left `CMD_SRCS`, plus the
default build, `go vet`, the full Go suite, the full sequential C suite (254
binaries, zero aborts) and all 18 module-inventory gates.

The lesson that made this reachable at all: `make -C src` does not build
`../aimee-server`, so "the build is clean" from that command says nothing about
the server link.

PR #2552, opened 2026-08-11
fix: stop the compress lever from silently defeating the fold freeze

JBailes:
## The bug

The §3 fold-freeze exists to keep the reduced **prefix** byte-identical turn to turn so the provider prompt cache keeps hitting. `test_context_fold.c` proved that for the fold lever *alone*. Nothing tested the **composed** path — which is what actually ships: the `aggressive` economizer preset enables `history_fold` **and** `compress` together, compress runs first, and the fold digests the *compressed* view.

`context_compress_view` prepended a synthetic note at **index 0** whose text embeds a count of compressed bodies plus a Coordinate Closet that accumulates. Both grow as messages age out of the retained band, so the first message of the prefix was *by construction* a summary of a growing region. The prefix digest therefore changed on **every turn**, the freeze re-epoched every turn, and the boundary was never once reused — every turn paying a provider cache **write** exactly where the freeze exists to buy a cache **read**.

Every lever still reported success, so nothing surfaced it.

## Measurement

14-turn composed run (fold + compress + freeze + recall_inject):

| | cache-warm reuses | epoch advances |
|---|---|---|
| before | **0** | 14 (one per turn) |
| after | **7** | 6 |

Isolation runs pinned the cause exactly:

- compress **off** → 9 reuses / 4 epochs
- compress **on**, closet disabled (which suppresses the note) → 9 reuses / 4 epochs — identical

The note was the whole of it.

## The fix

Append the note at the **tail** instead of the head. The tail already varies every turn, so a growing note costs nothing there — the same placement argument `context_reduce.h` already makes for the §4 recall hint, whose injection shape this now mirrors. Downstream the note lands in the fold's **retained** region, so conserved identifiers survive verbatim rather than being skeletonized, and `"body(ies) above"` becomes literally true (at the head it claimed "above" with nothing above it).

The head insertion used a user+assistant **pair** to keep Anthropic role alternation intact. At the tail only the ends-with-user case needs a bridge ack, and the note stays last so the transcript never ends on an assistant turn (which would read as a prefill). Both end-states are pinned by a test.

### Tradeoff, measured

Because the note is now retained verbatim rather than folded into the skeleton, it costs tokens every turn. Measured over the same 14-turn run: the note grows ~35 B/turn and reaches 845 B, about **7.5%** of the reduced transcript by turn 13, and it is always the last message. That is the price of the warm prefix, and it buys cache reads on the entire folded prefix instead of a full write every turn. Conservation also gets strictly stronger — identifiers survive verbatim instead of being excerpt-truncated by the skeleton.

## Tests

- **`test_prefix_stable_across_turns`** — drives 14 turns through `context_reduce` with the full lever stack and asserts the emitted prefix is **byte-identical** whenever the epoch counter holds, allowing change only on a genuine epoch advance. Compares *bytes*, not token counts: a cache hit is a bytewise prefix match, so anything weaker would pass on a prefix that merely looks the same. It guards against vacuity by requiring at least one real reuse — **that guard is what caught this bug.**
- **`test_compress_note_preserves_alternation`** — no consecutive user turns, and never an assistant-final transcript, for both end-states.

## Verification

- Retention quality unchanged: fold **20/20** recall at 47% cost; legacy 11/20, record 15/20.
- Economizer suite green (`context_fold`, `context_reduce`, `coord_closet`, `fold_recall`, `fold_register`, `fold_budget`, `economizer_{proof,anthropic,openai,activation,wire_snapshot}`, `config_economizer`).
- All **48** lint checks pass; `aimee-server` and `aimee-kb` both link; `cmd-srcs-compile-check` ok.

## Note on the cache A/B

This supersedes the live cache A/B I had been attempting. That measurement is not obtainable from delegate traffic: the delegate provider reports `cache_read: 0` / `cache_write: 0` across 158 calls and 3.1M prompt tokens on CT 403 (and the same on .210), so the telemetry needed simply is not emitted. Prefix byte-stability is the property the cache actually keys on, it is decidable deterministically, and — unlike the A/B — it found a real bug.

### DIFF
--- a/src/modules/economizer/context_fold.c
+++ b/src/modules/economizer/context_fold.c
@@ -394,19 +394,32 @@ int context_compress_view(const cJSON *messages, const fold_config_t *cfg, fold_
       return 0;
    }
 
-   /* Conserve any amputated identifiers in a Coordinate Closet, prepended as a
-    * synthetic user+assistant note pair (matching context_fold_view): a plain-text
-    * turn is valid input for every provider builder, and the PAIR keeps Anthropic
-    * role-alternation intact ahead of the (unchanged) original first turn. When the
-    * closet is disabled or empty, render returns NULL and we emit no note — the head
-    * excerpts still carry the leading bytes of each body. */
+   /* Conserve any amputated identifiers in a Coordinate Closet, APPENDED as a
+    * synthetic user note at the very end of the transcript. When the closet is
+    * disabled or empty, render returns NULL and we emit no note — the head excerpts
+    * still carry the leading bytes of each body.
+    *
+    * Tail placement is load-bearing, not cosmetic. This note summarizes the WHOLE
+    * compressed region (a body count, plus a closet that accumulates), so its text
+    * necessarily changes every turn as more messages age out of the retained band.
+    * At the head it sat inside the fold's frozen prefix, which made the reduced
+    * prefix differ on every single turn and silently defeated the §3 fold-freeze
+    * whenever compress and history_fold ran together (the shipping `aggressive`
+    * preset enables both) — every turn paying a provider cache WRITE where the
+    * freeze exists to buy a cache READ. The tail already varies each turn, so a
+    * growing note costs nothing there; this is the same placement argument
+    * context_reduce.h makes for the §4 recall hint, and it mirrors that injection's
+    * shape (a lone user-role note, no ack). "above" is now literally true, too.
+    *
+    * Downstream the note lands in the fold's RETAINED region, so conserved
+    * identifiers survive verbatim rather than being skeletonized. It is never
+    * written back to stored history, so successive turns do not accumulate notes. */
    char *closet = coord_closet_render(&set, &cfg->closet, compressed_raw, &out->closet_evict);
    coord_set_free(&set);
    if (closet)
    {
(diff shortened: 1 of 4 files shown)


JBailes:
Reviewer note on why this shipped unnoticed: the freeze is activated by **`reduce_state_t` being non-NULL**, not by `reduce_config_t.freeze` — the fold path derives `freeze_arg` purely from `st ? &st->freeze : NULL` and never reads `cfg->freeze`, which is vestigial there. So on the delegate seam the freeze is live whenever per-conversation state is threaded (`agent_runtime.c:866` passes `&agent_reduce_state`), even though nothing sets the flag. My test sets `cfg.freeze = 1` for intent; the state is what actually arms it.

Blast radius is bounded to the delegate seam. The gateway wire path is compress-only by design (`gateway_mutate_wire.c`: `rc.compress = 1`, "D3: compress-only at the gateway in v1 (fold deferred)") and passes `st = NULL`, so there is no fold and no freeze to defeat there. `dedup` and `inject_compress` are not enabled by any preset. The composition that broke — fold + compress + live state — is reachable only via `economizer.mode=aggressive` (or `fold.enabled` plus aggressive's compress) on the delegate seam, which is exactly what `test_prefix_stable_across_turns` now models.

PR #2554, opened 2026-08-11
port(economizer): Go module foundation — register, recall, coordinate closet

JBailes:
First installment of moving the economizer out of C, per the directive that this belongs in Go. **This is a foundation, not the finished port** — see *Status* below for exactly what is and is not done.

## What is ported

Three leaf units, in dependency order, with every test carried over one-for-one from the C suite so the port is pinned to the behaviour it replaces:

| C source | Go | cases |
|---|---|---|
| `fold_register.c` (91 ln) | `register.go` | 5 → 5 |
| `fold_recall.c` (147 ln) | `recall.go` | 6 → 6 |
| `coord_closet.c` (651 ln) | `closet.go` | 15 → 15 |

26 tests, `go vet` clean, `gofmt` clean, whole `server-go` tree builds.

## A real test hole, found by mutation

`test_coord_closet.c` asserts `contains(out, "src/modules/git")` for the two-slashes-no-extension path case. That substring is satisfied by `"src/modules/git/retry.c"`, which is conserved via the **dot** rule — so the extensionless branch of `match_path` was never actually covered. Forcing that branch to always reject leaves the C-shaped assertion passing.

The Go test asserts on a whole rendered line instead, and now fails under that mutation and passes when restored. **The equivalent C assertion is still vacuous** — worth fixing in the C suite whether or not this port lands.

## Design decisions

- **One bus stage, not many.** `coord_closet`, `fold_recall` and `context_fold` are internal to a reduction, not separately callable, so they stay internal Go packages. Only `economizer-reduce` needs to be on the wire.
- **The module can be stateless.** `reduce_state_t` is already persisted by the C side via `db1_economizer_state_save/_load`, so state travels in and out with the request rather than the module owning a store.
- **Ordinal 27**, appended to the optional inventory, is the only safe slot: `principal_ref` must equal the inventory ordinal, so inserting mid-list renumbers every later module and changes event kinds on the wire. That gives event kind `4096 + 27*256 + 1 = 11009`.

## Status — what is NOT done

Not yet cut over, deliberately. The migration contract says a rule must never live in two languages. Nothing dispatches to this package yet — no descriptor, no stage, no bus wiring, no caller — so it is **inert code beside the live C**, not a second live implementation that can diverge. The cut-over debt is real and must be paid before anything routes here.

Remaining, roughly in order:

1. `context_fold.c` (643 ln) — including the freeze fix from #2552, which must survive into Go.
2. `context_reduce.c` (607 ln) — the orchestrator, plus the composed prefix-stability test.
3. `tool_condense.c` (1,179 ln), `economizer_json.c` (544), the provider shims (`economizer_openai/anthropic/proof`, ~1,000), `task_rail.c` (245), `episode_seal.c` (165), the gateway wire files (~370).
4. Module wiring: canonical inventory, `process-contracts.json`, `module.yaml`, `docs/modules/economizer.md`, dispatch table, registry-vs-contracts test.
5. Cut C over and delete it.
6. `compact_body()` lives in `src/compact.c`, OUTSIDE the module, and `context_fold` depends on it — that dependency needs a decision (port it, or keep it in C as shared substrate) before the fold can move.

That is ~5,000 further lines of C plus the wiring. It is a program of work, not a single change.

## The two production bugs

Both were held back from C hotfixes per the decision to do this in Go only, and both are therefore **still live in production**:

- **Cache telemetry is blind for every OpenAI-family model.** Root cause found: `aimee_backend_openai.c:412` and `aimee_backend_responses.c:298` parse only the top-level token counts and never read the `*_details.cached_tokens` sibling that is already in the response. The IR fields (`usage_cache_read`/`usage_cache_write`) and all downstream accounting already exist — Anthropic populates them. This is why delegate traffic reports `cache_read: 0` across 3.1M prompt tokens; the provider sends it, aimee discards it.
- **The fold freeze never engages under `economizer.mode=aggressive`** (#2552, green, unmerged) — every turn pays a cache write where the freeze exists to buy a cache read.

### DIFF
--- a/server-go/modules/economizer/closet.go
+++ b/server-go/modules/economizer/closet.go
@@ -1,0 +1,589 @@
+package economizer
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// Coordinate Closet — verbatim identifier conservation (fold §2).
+//
+// Folding replaces detail with a summary, and a summary paraphrases. An exact
+// identifier that gets paraphrased is destroyed: a path becomes "the config
+// file", a sha becomes "the commit". The closet is what makes folding safe to do
+// at all — the exact bytes of every identifier in the folded region are carried
+// forward verbatim, so the model can still act on them.
+//
+// Ported from src/modules/economizer/coord_closet.c. Determinism is a hard
+// requirement, not a nicety: the folded prefix has to stay byte-identical across
+// turns for the prompt cache to keep hitting, so ordering comes from an explicit
+// total-order sort and never from map iteration.
+
+// Lane is the provenance trust boundary.
+type Lane int
+
+const (
+	// LaneAgent is agent/tool-originated content — trusted.
+	LaneAgent Lane = 0
+	// LaneUser is user-pasted content — quarantined, rendered below a divider
+	// and marked untrusted so a conserved value can never be mistaken for an
+	// instruction the agent produced.
+	LaneUser Lane = 1
+)
+
+// CoordKind is the coordinate kind, which also drives the fallback label.
+type CoordKind int
+
+const (
+	CoordKindUUID CoordKind = iota
+	CoordKindSHA
+	CoordKindPath
+	CoordKindKV
+	CoordKindRef
+	CoordKindHandle
+)
+
+// Provenance records where a coordinate came from. Unknown numeric fields are -1.
+type Provenance struct {
+	Lane        Lane
+	TurnID      int64
+	ToolCallID  int64
+	ResultIndex int
+}
+
+// DefaultProvenance is the AGENT lane with all ids unknown.
+func DefaultProvenance() Provenance {
+	return Provenance{Lane: LaneAgent, TurnID: -1, ToolCallID: -1, ResultIndex: -1}
+}
+
+// CoordEntry is one conserved coordinate.
+type CoordEntry struct {
+	Value       string // conserved verbatim
+	Label       string // deterministic label
+	Kind        CoordKind
+	Prov        Provenance
+	FirstOffset int // first-occurrence byte offset in the source
+}
+
+// ClosetConfig is the runtime config. Default-off.
+type ClosetConfig struct {
+	Enabled     bool
+	BudgetBytes int    // hard byte cap for the rendered block; 0 = default
+	MaxRatioPct int    // closet bytes <= rawLen * pct/100; 0 = default 100 (1x)
+	Denylist    string // extra secret patterns (comma/space separated)
+}
+
+const (
+	ClosetDefaultBudgetBytes = 2048
+	ClosetDefaultMaxRatioPct = 100
+)
+
+// EvictResult reports whether everything nominated fit within the cap.
+type EvictResult int
+
+const (
+	// EvictNone means all nominated coordinates were conserved.
+	EvictNone EvictResult = 0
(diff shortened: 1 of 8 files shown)


JBailes:
## Blocker found for the next installment: JSON serialization fidelity

Verified empirically before starting `context_fold`, because it determines that port's shape.

`context_fold` embeds `cJSON_PrintUnformatted(input)` **directly into the skeleton text** (tool_use input, tool_calls arguments, tool_result content), and `prefix_digest` hashes the serialized bytes of each message. So the fold's output depends on exactly how JSON is printed.

Go's `encoding/json` does not print like cJSON:

```
cJSON_PrintUnformatted: {"zeta":1,"alpha":"a<b&c","path":"/x/y"}
Go json.Marshal(map):   {"alpha":"a\u003cb\u0026c","path":"/x/y","zeta":1}
```

Two divergences, both fatal here:

1. **Key reordering.** Go sorts map keys; cJSON preserves insertion order.
2. **HTML escaping.** Go escapes `<`, `>` and `&` as `\u003c` etc. by default.

Consequences if ported naively: the skeleton text differs from C for the same input, and — because the folded prefix must stay byte-identical for the prompt cache — this reintroduces exactly the class of silent per-turn cache miss that #2552 fixes. It would not fail any test that only checks *content*.

**So `context_fold` needs an order-preserving, cJSON-compatible printer first** (`json.Decoder` + `UseNumber` into an ordered node type, and `Encoder.SetEscapeHTML(false)`). `compact.go` in this PR already carries half of it — `jsonNode` preserves object key order for exactly this reason — so the printer is a natural extension of that type rather than new scaffolding.

Noting it here so the next installment starts from the decision rather than rediscovering it. This is the kind of thing the migration notes mean by "the remaining work is state-relocation decisions, not a search".

JBailes:
## Correction: `compact_body` never had two live callers

My earlier note on this PR (and in the `compact.go` header) warned that `compact_body` had two production C callers, so the Go copy must not go live until the eager seam moved too. **That premise was wrong**, and the correction is more useful than the port it prompted.

`agent_compress_tool_result` has **no callers and no tests.** A repo-wide search across every file type finds only its definition (`agent_policy.c:898`), its declaration (`agent_exec.h:249`), and prose in `compact.h`. Everything else is docs and benchmark corpora.

It is deliberately unreachable, not accidentally so: a CI gate used to enforce it, listing the symbol under `legacy_calls` beside `context_reduce` and `build_fold_view` and failing if any production source reached it. That gate was removed in 9d478dcaa6 (economizer tiers).

Consequences:

- `compact_body` has exactly **one** live caller — the lazy lever in `context_fold.c`. Moving it to Go creates no two-languages problem.
- `compact.h`'s claim that the eager path is "the ONLY writer of compacted bodies into history" describes something that does not run.
- The Coordinate Closet's P1 "wired return-only into `agent_compress_tool_result`" wiring is **inert** with it — so eager identifier conservation is not happening in production either. That is worth a separate look: it means tool results are capped but never compacted-with-conservation on the way into history.

**Recommendation: delete the C function** rather than keep it beside this port. Keeping both is the two-languages problem for real, and for a path nothing calls. The Go version is kept so the seam has one home if revived, and because it is now the only *tested* version of that logic — the C original never had a test.

Separately: #2552 merged, so the freeze fix is live in `testing`. Of the two production bugs, only the OpenAI-family cache telemetry gap remains open.

PR #2558, opened 2026-08-11
port(economizer): context_reduce — the orchestrator

JBailes:
Ports `src/modules/economizer/context_reduce.c` (607 lines): lever composition (compress ahead of fold), the freeze cost guardrail, the §4 page table and hint injection, seam gating, provenance, the net-gain pre-check, and the ledger. **74 tests** in the package, `go vet`/`gofmt` clean, tree builds.

**Stacked on #2557** (`context_fold`), which is stacked on #2556 (the cJSON printer).

## The headline result

The composed prefix-stability test — the Go form of the one that found the compress-defeats-freeze bug — reports:

```
7 cache-warm reuses, 6 epoch advance(s)
```

which is **exactly** what the fixed C reports in #2552. That is independent confirmation the port reproduces the *fixed behaviour*, not merely that it compiles: same transcript, same lever stack, same boundary arithmetic, same answer.

The test carries its own vacuity guard across too. It requires at least one real reuse, because a run where the fold never engaged would satisfy every byte-equality assertion trivially — and that guard is precisely what exposed the original bug.

## One deliberate design change

The C freeze guardrail calls `token_estimate_cost_ex` to price each cache tier, reaching into the pricing table. The Go module takes the three resolved **rates** as input (`PriceRates`) instead, so it stays a pure transformer with no ambient state — which is what lets it serve a bus stage without owning a config or pricing connection.

The arithmetic itself (`reduce_freeze_favorable_rates`) is already pure and ports one-for-one; its C test cases came across unchanged, including the fail-open paths for an unpriced model and a zero-size prefix. Fail-open matters here: a guardrail that silently *disables* the freeze on an unknown model would cost money quietly.

## Also covered

- Seam gating is a **true** no-op — no measurement, no cost — at a seam the operator did not enable.
- `measure_only` computes the ledger and leaves the transcript byte-identical.
- A second seam re-measures but does not re-account the saving (no double-counting).
- Below `min_gain` the fold skips without mutating.
- A recall hint surfaces for a coordinate the newest turn re-touches, and is inert when recall is off.
- The injected notice lands at the **tail**, is labelled as not coming from the user, and never enters the cache-stable prefix.
- Compress engages on a tool loop where the fold correctly finds no clean boundary.

## Status

Port is roughly **60%** by line count. Nothing dispatches to this package yet, so it stays inert beside the live C rather than a second live implementation that could diverge.

Remaining before cut-over: `tool_condense.c` (1,179 ln), the provider shims (~1,000), `task_rail`/`episode_seal` (410), the gateway wire files (~370), the reducer state serialization, and the module descriptor + `process-contracts.json` entry + `docs/modules/economizer.md`.

Carried decisions, both still open: the dead C seam `agent_compress_tool_result` should be **deleted** rather than kept beside its Go port (#2554), and the OpenAI-family cache telemetry gap (`aimee_backend_openai.c:412`, `aimee_backend_responses.c:298`) remains live in production.

### DIFF
--- a/server-go/modules/economizer/reduce.go
+++ b/server-go/modules/economizer/reduce.go
@@ -1,0 +1,411 @@
+package economizer
+
+import "strings"
+
+// The context_reduce orchestrator: composes the levers, applies the freeze cost
+// guardrail, tracks the page table, and reports the ledger.
+//
+// Ported from src/modules/economizer/context_reduce.c.
+
+// charsPerTokenEst is the chars/4 token estimate the module uses throughout.
+// A forecast, not an invoice: token COUNTS are caching-independent, but the
+// number of tokens a provider actually bills is its tokenizer's business.
+const charsPerTokenEst = 4
+
+// FreezeGuardMaxHorizon caps the configured horizon, so a misconfigured long
+// horizon cannot justify an unbounded cache-write bet on a single run.
+const FreezeGuardMaxHorizon = 5
+
+// Seam identifies where the reducer is running.
+type Seam int
+
+const (
+	SeamGateway Seam = iota
+	SeamDelegate
+)
+
+// ReduceReason records why a request did or did not reduce.
+type ReduceReason int
+
+const (
+	ReasonNone       ReduceReason = iota // no lever enabled at this seam
+	ReasonReduced                        // reduction applied
+	ReasonMeasured                       // measure-only: metrics computed, not mutated
+	ReasonSkipNoGain                     // foldable tokens below MinGainTokens
+	ReasonAlready                        // provenance: a prior seam already reduced
+)
+
+// PriceRates are the per-token provider rates the freeze guardrail needs.
+//
+// Passed IN rather than looked up, because the pricing table lives with the
+// caller. That keeps this module a pure transformer with no ambient state, which
+// is what lets it serve a bus stage without owning a config or pricing
+// connection. Priced=false means the model is unknown and the guard fails open.
+type PriceRates struct {
+	Priced    bool
+	InputCost float64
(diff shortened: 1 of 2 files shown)


JBailes:
## A real C bug, found by porting and verified against the C itself

`reduce_state_serialize` budgets each recall key against the serialized size so far, with a 48-byte allowance:

```c
if (used + strlen(k) + 48 > REDUCE_STATE_SERIAL_MAX) break;
```

…but it appends `recall_kept` and `recall_dropped` **after** the loop — roughly 40 more bytes the allowance never accounted for. When the last accepted key sits near the cap, the finished blob overshoots and the closing guard throws **the whole thing** away:

```c
if (out && strlen(out) > REDUCE_STATE_SERIAL_MAX) { free(out); return NULL; }
```

Measured by compiling the real C against a throwaway harness:

```
short keys (the C test fixture):  OK (len=6128)
long keys (deep repo paths):      NULL — state not persisted
```

**Impact.** For a conversation whose evicted coordinates are long paths — the common case in this repo — the entire reducer state is silently dropped: the freeze boundary *and* the page table. Cross-run recall stops working, the freeze restarts cold every run, and nothing reports a failure. The C suite passes because its fixture uses ~37-character keys, which stay clear of the boundary.

**The Go port reserves the two counters before the loop** so the budget measures the real final size. That is a *deliberate* divergence from C — called out at the site, and the only place in this port that intentionally does not reproduce C behaviour. Both fixtures are pinned: the short-key one from the C suite, and the long-key one C drops.

This is live in C today. Worth fixing there directly if the cut-over is not imminent, since it silently disables two features that are supposed to be working.

PR #2561, opened 2026-08-11
fix(telemetry): read OpenAI prompt-cache hits; delete the dead eager seam

JBailes:
Two related fixes to the tool-result / telemetry path.

## 1. OpenAI-family cache telemetry was *discarded*, not missing

`aimee_backend_openai.c` parsed only `usage.prompt_tokens`/`completion_tokens`, and `aimee_backend_responses.c` only `usage.input_tokens`/`output_tokens`. Neither read the cached-token sibling OpenAI **already returns**:

| wire | field |
|---|---|
| Chat Completions | `usage.prompt_tokens_details.cached_tokens` |
| Responses | `usage.input_tokens_details.cached_tokens` |

`aimee_backend_anthropic.c` has always read `cache_read_input_tokens`, which is why only Anthropic-egress ever showed cache numbers. The IR fields (`usage_cache_read` / `usage_cache_write`) and every downstream consumer already existed and worked — the two reads were simply absent.

**Consequence until now:** every OpenAI-family model's cache savings were invisible in cost accounting. Measured on live delegate traffic across two servers: 158 calls, 3.1M prompt tokens, `cache_read` 0 for the whole day. I originally recorded that as *"the provider does not emit it"* — that was wrong, and the corrected finding is what produced this fix.

### The subtraction is the part a naive fix gets wrong

The two providers count differently:

```
Anthropic: input_tokens  EXCLUDES cache_read_input_tokens  -> additive
OpenAI:    prompt_tokens INCLUDES cached_tokens            -> overlapping
```

`token_estimate_cost_ex` sums every bucket independently, so assigning both fields without subtracting would bill the cached prefix **twice** — once at the full input rate and again at the cache-read rate, turning a cost *saving* into a reported cost *increase*. So `usage_in` has the cached count subtracted, and a test asserts the buckets still add up to the provider's own total.

There is no cache-**write** counterpart: OpenAI creates cache entries implicitly and charges nothing, which is exactly why the freeze guardrail treats a non-positive write premium as pure upside.

Tests cover both wire shapes, the no-details and zero-count cases (the common non-cached response must be byte-unchanged), and a defensive case where a provider reports more cached than prompt tokens — `usage_in` must never go negative. **Verified by mutation:** removing the subtraction fails the suite.

## 2. Deleted `agent_compress_tool_result` and its orphaned helper

It had no callers and no tests. A repo-wide search found only its definition, its declaration, and prose in `compact.h`. It was *deliberately* unreachable: a CI gate once enforced exactly that, listing it under `legacy_calls` beside `context_reduce` and `build_fold_view` and failing if any production source reached it — until that gate was removed in `9d478dcaa6` (economizer tiers).

Removing it orphaned `compact_cfg_from_app_config`, which `-Werror` caught — good evidence the whole path was dead rather than merely unused. That helper goes too.

`compact.h`'s contract is corrected: it described the eager path as one of two live callers of `compact_body`, which has exactly one. The header now records the consequence worth knowing — **eager identifier conservation into history is not happening**, so the Coordinate Closet runs on the lazy seam only. That is a real gap, separate from this PR, and now documented where someone will find it.

## Verification

All **48** lint checks pass; `aimee-server` and `aimee-kb` link; `cmd-srcs-compile-check` ok; `aimee-backend`, `agent-ir-parse`, `aimee-ir-metrics`, `token-tracker` and `compact` suites green.

## Note

This is C, against the standing "economizer work goes to Go" direction — landed as an explicitly requested exception because both items are live in production and the Go path for translation is not on the economizer port's route. The economizer port itself continues in Go (#2556, #2557, #2558).

Once this is deployed, the cache A/B that was previously unmeasurable becomes measurable on delegate traffic for the first time.


JBailes:
## Scope reduced — and the reason is worth recording

CI caught this immediately and it was a **correct** catch on my part, not a flaky gate:

```
check_c_repository_lock: error: translation: vendored mirror differs from its repository pin
```

`src/modules/translation/` is a **vendored mirror** of `RakuenSoftware/aimee-module-translation`, pinned at `v0.3.0` with a `source_sha256` over the owned tree. Its `module.yaml` owns not just the two backend `.c` files but `src/tests/test_aimee_backend.c` as well — so the cache-telemetry fix, implementation *and* tests, sat entirely inside a vendored module.

Repinning to make CI green would have made the lock assert "this is v0.3.0" while the content differed, which defeats the point of having a lock. So I removed that half rather than paper over it. This PR is now just the dead-seam deletion, which touches no vendored files.

`check_c_repository_lock` now passes (1 core, 26 modules), 48 lint checks pass, both binaries link.

## What the cache fix needs

It has to go upstream to `aimee-module-translation` and come back as a version bump. That is not a quiet operation: `refresh_lock_from_repositories` pins **every** module repo plus core at one shared version and requires each to be clean and tagged `v{version}` at HEAD, so a translation change means cutting a coordinated release across all 27 repositories.

The patch is preserved. I have ADMIN on the upstream repo, so I can do it — but a 27-repo version bump is a release decision, not a side effect of a bug fix, so I am asking before pulling that lever.

Worth remembering the cost of leaving it: every OpenAI-family model's cache savings stay invisible in cost accounting, and the fold's cache benefit stays unmeasurable on delegate traffic.

PR #2623, opened 2026-08-13
chore(tests): freeze the refactor surface baseline (unblocks every PR cut from testing)

JBailes:
The refactor surface baseline has been stale on `testing` since **02db35baaa** ("delegates: the brief no longer decides what a delegate may do"), which changed delegate headers without re-freezing. `tests/baselines/refactor/index.json` was last frozen at **6d0d7dd555**, an *ancestor* of that commit — so the recorded hashes describe headers that no longer exist in that form.

**This is not local to the delegates work.** `refactor_baselines.py` compares the whole surface, so every PR cut from current `testing` fails the `module-inventory` job with:

```
surface drift: first difference at canonical line 382:
  expected '...171bd0caf87b258...'  actual '...9421627b902afe74...'
```

on `src/headers/cmd_agent_delegate_impl.h` — a file the PR never touched. That is how it surfaced: three unrelated `/tmp`-hygiene branches (#2619, #2620, #2621), and the most recently cut one started failing a check about delegate headers. #2619 and #2620 predate the drift landing and are green; #2621 is not.

**Scope:** the freeze and nothing else. Seven hashes move — six delegate headers plus `docs/gen/v1-route-descriptor.json`. No source changes ride along; `git diff --name-only` is exactly one file.

**Verified:** `python3 -I -S scripts/refactor_baselines.py` reports `ok`.

Once this lands, #2621 and #2622 want a rebase to pick it up.

### DIFF
--- a/tests/baselines/refactor/index.json
+++ b/tests/baselines/refactor/index.json
@@ -379,7 +379,7 @@
         },
         {
           "path": "src/headers/cmd_agent_delegate_impl.h",
-          "sha256": "171bd0caf87b258cdc96479e73c6cde536aabeff56ad817c6ea00ba67a7c5fd0"
+          "sha256": "9421627b902afe74ce244bc388dbdea88a1816ebcc6d5f4c5e7ba65055a5dfe3"
         },
         {
           "path": "src/headers/cmd_agent_delegate_internal.h",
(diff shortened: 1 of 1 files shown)


JBailes:
Closing: this is the one thing CI explicitly refuses, and I should have read the gate before writing it.

`module-inventory.yml` runs the refactor-baseline check only when `github.base_ref != 'testing'`, and the job's own message says the baseline is *expected* to be stale on `testing` — the promotion PR into `main` re-freezes it once, and that single diff is the review of what the release changed.

The failure that led me here was real but differently caused: #2621 and #2622 were stacked on `agent/tmpdir-batch-6`/`-7` rather than `testing`, so they fell outside that exemption and ran the strict check against a deliberately-stale baseline. They are now retargeted at `testing` and the freeze commit has been dropped from the stack. No baseline change is needed.

PR #2761, opened 2026-08-16
make regression tests causal and focused

JBailes:
Task 7 exposed two weaknesses in the first autonomous-test wording: one authored concurrency test could pass the unmodified implementation because it coordinated before observing shared state, and every run retried verification after discovering the documented runner was unavailable.

This refinement keeps tests as an autonomous implementation habit rather than a completion gate. It requires regression tests to deterministically distinguish intended behavior by controlling the causal boundary, makes risk/contracts—not line or test count—the sizing signal, explicitly rejects optimizing for fewer lines, and limits a missing-runner fallback to one direct equivalent.

Validation:
- `src/build/obj/tests/unit-test-prompts`
- `git diff --check`
- Aimee roundtable review approved with no findings


JBailes:
Closing after the Task 7 acceptance run regressed both cost and authored-test sensitivity. Two completed cells used 17.14 credits; one still passed the unmodified implementation (33.64/100). This candidate will not be promoted.

PR #2785, opened 2026-08-17
Serve the command catalogue instead of compiling it into the thin client

JBailes:
Second step of the server-driven thin client. **Stacked on #2780** (route manifest) — merge that first; this branch contains its commits.

## The problem

The client carried the catalogue of what the server can do: 67 rows of `name/summary/tier/subcommands` in `src/cli_help_data.h`, and `docs/gen/cli-commands.md` was generated **from it**. So the *client* was the source of truth for the *server's* command surface. A client one release behind advertised the wrong set and had to be rebuilt to learn a new command.

## The change

The rows **move** — `git mv`, so there is never a second copy — to `src/server/cli_command_defs_data.h`, and ride the manifest the client already fetches:

```json
{"manifest_version":1,
 "routes":  [{"op":"config.show","verb":"GET","path":"/v1/config"}],
 "commands":[{"name":"memory","summary":"Stored memory","tier":"core","subcommands":"..."}]}
```

`routes` says how to **address** a method; `commands` says what exists and what it's for. Both answer "what can I do?", so both travel in one document and one round trip.

The client renders what it's given. It keeps a deliberately tiny bootstrap list — **`remote`, `version`, `help`** — the commands you need *before* you can reach a server. Every entry there is a thing that can rot, so it stays short on purpose. All five call sites move off the compiled-in table (tier listings, `help --all`, `help <cmd>`, has-subcommands, is-known).

## Failure stays legible

With no catalogue the client says:

```
aimee: cannot reach aimee-server to ask what commands it has (GET /v1/cli/manifest).
Cannot list commands: no catalogue from the server.
```

…rather than reporting a real command as unknown. Misreading a transport problem as a typo is the failure this whole line of work exists to stop; the client must not commit it here either.

## A bug this nearly shipped

Regenerating the docs silently collapsed `docs/gen/cli-commands.md` **from 812 lines to 8**. `gen-reference-docs.py` parses rows on the `CLIENT_TIER_` token, which became `AIMEE_CMD_TIER_` when the data moved — so every row stopped matching and the generator emitted an **empty reference without failing**. Parser repointed; the doc is back to 812 lines with only its provenance line changed.

Worth noting for review: that generator fails *open*. It will happily write an empty document if its regex stops matching. `docs-gen-check` only compares output to git, so a same-shaped emptying would pass review as "the docs changed".

## Verified

- 67 commands served by a live server; `aimee help memory` renders the server's summary + subcommands; `help --all` lists 60 rows
- no-server path falls back to bootstrap **and states the reason**
- `check-cli-help-coverage.py` follows the data: 46/46 routed commands documented, 0 known gaps
- client test set passes (9 binaries incl. `mcp-serve`, `delegate`, `transport`, `gateway`, `poll-deadline`)
- `make lint` passes (all 60 checks) with the regenerated doc committed

## What this does NOT do

The catalogue is now on the correct side of the wire but is still **one hand-written table**. `command_registry.h` says each module registers its own commands, and measurement shows **nothing currently registers** — `aimee_command_register` is called only from its own tests, so the registry is empty at runtime and its one production reader iterates nothing.

Making that invariant real means modules owning their rows. Modules are separate processes, so the natural home is a `commands` field in `module.json` — which doesn't exist yet. That's a schema change plus a migration across ~30 descriptors, written up as the next phase in `docs/proposals/pending/server-driven-thin-client.md` rather than started mid-flight.

Still client-side after this: the argument marshallers, and the 14-row `bespoke[]` route table (server handlers that aren't `rh_dispatch_op`, so the manifest filter excludes them).


JBailes:
⚠️ **Not ready to merge — please treat as draft.** (I don't have permission to flip the draft flag via the API.)

Verifying this against `build-integrity` surfaced a conflict I can't resolve without a product decision.

**What broke:** moving the catalogue to the server means `aimee provider --help` and `aimee identity snapshot --help` exit 1 when no server answers. `test_build_integrity.sh` asserts the opposite, in a check named *"server-routed provider help is client-side"* — so offline help is a deliberate, tested invariant, and this PR breaks it.

**What I added:** the client now caches the last catalogue an endpoint returned (keyed by endpoint, so pointing at a different server never shows the previous one's commands) and falls back to it, saying the list may be stale. Verified working with the server stopped, and verified a different endpoint does not see another server's cache.

**Why that isn't enough:** the cache only helps *after* one successful contact. On a fresh install there has never been one, so `--help` still can't answer.

These two cannot both hold as stated:

1. the command catalogue lives on the server (this PR's goal — the client shouldn't hold a copy of what the server can do)
2. `<cmd> --help` works with no server (the existing invariant)

…unless the client also ships a **built-in catalogue generated at build time from the same server-side data file**, used only when there's no server and no cache. That keeps one source of truth in the repo and keeps the server authoritative at runtime, at the cost of a bundled fallback that can be stale.

Three options, and this is your call:

- **A.** Ship the generated built-in fallback (satisfies both; adds a build step and a stale-but-marked fallback)
- **B.** Accept that `--help` needs a server and retire the invariant (simplest; degrades a genuinely useful offline affordance)
- **C.** Narrow the invariant to the bootstrap commands only (`remote`, `version`, `help`), which are the ones you need before you can reach a server anyway

I'd suggest **A**, since the objection to a compiled-in catalogue was that it was *authoritative*, not that it existed — as a clearly-marked last-resort fallback it doesn't reintroduce the coupling that motivated this work.

**#2780 (the route manifest) is unaffected by any of this and is green at 31/31** — it's independently mergeable and carries the headline change.

JBailes:
**Resolved — no longer blocked, and the DO-NOT-MERGE marker is lifted.**

The conflict I raised above (server-owned catalogue vs. `--help` working offline) turned out to be reconcilable once I looked at what the original objection actually was.

The problem with the compiled-in table was never that it *existed* — it was that it was **authoritative**, so a client one release behind advertised the wrong command set and had to be rebuilt to learn a new one. A fallback that is consulted **last** doesn't reintroduce that.

So the client now consults, in order:

1. the **bootstrap list** (`remote`, `version`, `help` — the commands you need before you can reach a server)
2. whatever the **server** sent — live, or the endpoint-keyed cache of a live answer
3. the catalogue **compiled in at build time**, generated from the same file the server serves

Routing never consults (3), so being out of date there costs a stale *description*, never an unreachable command. That's the property that matters.

**Precedence is proven, not assumed.** Against a stub serving a sentinel summary that cannot come from the build, the client prints the served text and not the built-in one:

```
served summary shown: True
build-time summary shown instead: False
```

Also fixed here, from the same CI run:

**`windows-build` / `windows-cmake`** — the cache key hashed with `unsigned long`, which is **32 bits on LLP64**. The 64-bit FNV offset basis truncated and MinGW's `-Werror=overflow` rejected the constant outright. Now `uint64_t` with a matching `%llx`.

**Verified locally:** `build-integrity` passes (was exiting 1), `make lint` all 60 checks, `check-cli-help-coverage` 46/46, `provider --help` exits 0 with no server and `help --all` lists 60 commands, and the 9-binary client test set passes.

PR #2789, opened 2026-08-18
Serve the CLI dispatch rows, so a new server command is invokable

JBailes:
Third and largest piece of the server-driven thin client. Routes and the command catalogue are already served; this is the one that makes a new server command actually **invokable**.

## The gap this closes

The 216 rows mapping `aimee <group> <verb>` → method were compiled into the client. So even with its route and its help served, a capability added server-side stayed unusable: the client had no way to turn the words a user typed into a method it had never heard of.

The rows move to `src/server/cli_dispatch_defs_data.h` and ride the manifest as `dispatch`:

- **`routes`** — how to *address* a method
- **`commands`** — what exists and what it's for
- **`dispatch`** — how the typed words become one

All three answer "what can I do?", so all three travel in one document and one round trip.

## Proven, because otherwise the move buys nothing

```
$ aimee brandnew thing        # a spelling in NO build of this client
exit: 0
server saw: ['/v1/config']
```

Against a stub serving only that row. The **first** attempt failed at the marshaller —

```
aimee: 'brandnew.thing' — arguments are missing or invalid, so no request was sent.
```

— which is itself proof the lookup resolved, and marks the boundary precisely: **argument marshalling is the last body of server knowledge left in the client.**

## Served rows replace, not extend

A client that kept preferring its own build would still hide new commands — the exact coupling being removed. The compiled-in copy remains only for when no server can be asked, same as help. `test_served_dispatch` pins that, plus the three `sub` states that are easy to collapse silently:

| `sub` | meaning |
|---|---|
| absent | wildcard — matches whatever the first arg is |
| present, empty | matches only when no subcommand was given |
| present, set | exact match |

## Also fixes a regression from the catalogue PR

The bootstrap list was consulted **before** the served catalogue, so this build's wording shadowed the server's — `test_cli.sh` caught `help version` saying *"Print the client version"* where the server says *"Print version"* — and it would have listed any command present in both twice. Bootstrap is a fallback, not an override.

`test_cli.sh` is back to **51/57**, matching the pre-change baseline; the six remaining are service errors (workflow module not attached to the event bus, no agents configured, KB unreachable).

## One guard was failing open

Three scripts scrape these rows by path and all three broke when the rows moved. Two failed loudly and were repointed. `check_command_surface_parity.py` did **not**: its extractor returned `{}` and it printed

```
CLI routes (cli_command_routes[]):        0
CLI routes with no MCP counterpart: 0 across 0 groups
```

…and exited 0. That reads as *the surfaces agree* when it means *I could not see the rows* — the same fail-open shape as the docs generator that once emitted an 8-line reference. It now raises on an empty read, **verified by hiding the rows and watching it exit 1**.

## Verified

`make all` · `make lint` 60/60 · `test_cli.sh` 51/57 (baseline) · transport, subcommands, delegate, mcp-serve, poll-deadline, doctor, gateway, server-http, index-bootstrap tests. `unit-test-server-http` and one sibling gained the new object, which they need via `server_http_config_routes.o`.

## Remaining

Argument marshallers (~42 bespoke + the `MARSHAL_*` tables) are the last piece. The `command_registry` itself is still at **0 production registrations** — that stays blocked on the command→module ownership mapping, which is a judgment call I'm not making unilaterally.


JBailes:
Added a second commit hardening the two checkers that scrape these rows.

Both matched nothing and reported success at some point during this work — `gen-reference-docs.py` emitted an 8-line reference (from 812) after a token rename, and `check_command_surface_parity.py` printed `0 CLI routes … 0 with no MCP counterpart` and exited 0, which reads as *the surfaces agree* but means *I could not see the rows*.

The repo already had the idiom: 17 scripts carry `--plant-test`, and `check-native-tool-parity.py` states the rule — *"The check must FAIL on a planted violation, or it is decoration."* The four scripts scraping the command tables were the ones without it.

- `gen-reference-docs.py --plant-test` plants the **actual** regression (renames the tier token) and requires a failure
- `check_command_surface_parity.py --plant-test` feeds it rows it cannot see and requires a refusal
- both inject **in memory**; nothing in the tree is touched
- `docs-gen-check` runs the plant ahead of its diff, because that target only compares *output* against git — an emptied document is just "the docs changed" to it

`check-cli-v1-routes.py` and `check-cli-help-coverage.py` are deliberately left alone: both failed loudly when the rows moved, so they're not decoration.

A heuristic sweep flagged 17 more scripts; I did **not** act on it. Spot-checking showed the flags were mostly the plant-test idiom in a form my heuristic didn't recognise, and mass-editing gates whose invariants I'd be guessing at would add noise for no confirmed defect.

`make lint` 60/60 with the plant wired in.

JBailes:
Third commit: **the chain is closed.**

With routes, catalogue and dispatch served, a new server command was findable, documented and addressable — and still wouldn't run:

```
aimee: 'brandnew.thing' — arguments are missing or invalid, so no request was sent.
```

The client refuses to send a body it can't build, and had no row saying this method needs none. For the no-arg set that row is pure data, so it's served: `marshal` rows of `{method, args:"none"}`.

**Proven end to end.** `aimee brandnew thing` — absent from this client's dispatch rows, catalogue *and* no-arg list — now POSTs `/v1/brandnew/thing` and exits 0 against a stub serving only those rows.

### Deliberately only the no-arg set

Surveying all 42 marshaller functions:

| | count | |
|---|---|---|
| map argv → fields | 29 | servable once a spec carries types |
| + types / nested shapes | 5 | same |
| **read local disk / env / exec** | **8** | **should never be served** |

Those 8 (`marshal_git_cli`, `marshal_skill_request`, `marshal_index_file_request`, `marshal_identity_snapshot`…) read the **client's own** filesystem and environment. That isn't knowledge of what the server can do — a thin client reading local state in order to send it is precisely its job. Serving them would be *wrong*, not merely incomplete.

What remains coupled is **34 argv→field mappings**, which need a spec language — a design, not a move. I stopped at the data-only subset rather than invent a schema unsupervised.

Only an explicit `"args":"none"` makes the client send an empty body; an unknown or richer shape leaves the decision to its own marshaller rather than guessing `{}` for a command that needed arguments. `test_served_marshal` pins that, including absent-field and unknown-shape cases.

### Note on the guards

`check-cli-v1-routes.py` scrapes `MARSHAL_NO_ARGS` by path and failed loudly when the rows moved — correctly, and for the **fourth** time in this line of work. Every one of these checkers finds its data by file path and regex, so moving rows breaks them by construction. Repointed. (My first repair inserted a line mid-loop and left the script unparseable; lint caught it immediately, which is the argument for the gate.)

`make all` · `make lint` 60/60 · `test_cli.sh` 51/57 (baseline) · transport tests incl. the new one. Merged `testing` in and re-verified after.

JBailes:
Fourth commit — **the co-location ban had broken the integration harness, and nothing noticed.**

While checking for leftovers from the autostart removal I ran `make integration-tests`, which is a make target with **no CI job**. It failed 10 assertions:

```
FAIL: api status reports the listener   (got 'aimee: aimee-server unavailable')
FAIL: api enable turns the listener on  (got 'aimee: aimee-server unavailable')
FAIL: mcp git_status tool call          (got 'aimee-server unavailable after retries')
...
```

…against a server the harness had started and which was running perfectly well two lines above them. The harness reached it by **well-known-socket discovery**, which the co-location ban removed. It was fixed the same way `test_init_migrate_service.sh` already was — an explicit `AIMEE_API_ENDPOINT`, because a harness deliberately running both halves on one box now has to *say* so.

**63/73 → 73/73.**

### One of the ten was stale, not broken

```
"no server: DB command fails without autostart"  expecting  "server unavailable"
```

The command still fails, which is the property that matters — a pure client runs no handler locally. It now fails one step *earlier*: `cannot reach aimee-server to ask what commands it has`. The client can't learn that `memory list` exists without the server. That's the command table living server-side, i.e. exactly what this PR does, so the assertion now pins the new message. Its section header was `1. Auto-start (no server running)`; there is no autostart to be without.

### Gated it

An ungated harness is a harness that will be wrong when you next need it. Added an `integration-tests` CI job.

### Dead names swept

`AIMEE_NO_AUTOSTART` / `AIMEE_DISABLE_AUTOSTART` — nothing reads either, since the client starts no server to suppress — yet two harnesses still exported them and the reference-doc table still described one. (The doc output was already correct: env vars are scraped from `getenv`, and there is no `getenv`.)

`build-integrity` checked `windows/cli_client.c` **by path**, which is the very argument its own comment makes against naming one platform — a guard bound to a path enforces the rule at that path. Now tree-wide over non-test C sources, and **plant-tested**: a planted `getenv("AIMEE_NO_AUTOSTART")` in `cli_main.c` fails it.

Verified: build-integrity · lint 60/60 · integration-tests 73/73 incl. `thin-client-remote-exclusive` · init-migrate-service 4/4 · docs-gen in sync.

PR #2793, opened 2026-08-18
fix: let a real account close a stranded bootstrap login

JBailes:
"Secure your account" kept coming back on an appliance whose accounts had long since been created. It isn't a stale flag — **the step has no reachable completion at all.**

## What's actually on the appliance

`/var/lib/aimee/webchat/` on `.210`:

| file | contents |
|---|---|
| `bootstrap-user` | `generated:aimee-c78febb6b371` |
| `identities` | `aimee-c78febb6b371`, `admin`, `laura` |
| `bootstrap-replaced` | **missing** |
| `bootstrap-credentials` | **missing** |

Only `POST /api/setup/account` writes the marker that completes the step, and it refuses anyone not signed in *as* the generated login:

```go
if currentUser(r) != bootstrapUsername {
    writeJSONError(w, http.StatusForbidden, "sign in with the bootstrap account to replace it")
```

The operator's accounts were created by another route, so no marker was written. They can't sign in as the generated account to write one either: its password lived in `bootstrap-credentials`, which this endpoint deletes on success and which newer images never write. **The step asks forever for something nobody can do.**

## The nagging is the mild symptom

`adminUsername()` falls back to the pending bootstrap user. So the appliance administrator on `.210` is `aimee-c78febb6b371` — an account nobody can authenticate as. Every roundtable and workflow policy mutation 403s for `admin` and `laura` both.

That is the same lockout `adminUsername`'s own comment describes — *"locked out of every roundtable and workflow policy mutation on their own appliance"* — reached from the other direction.

## The fix

An established managed account may now **retire** the generated login instead of replacing it. It creates nothing — the caller already has a way in, and this only removes one. Same marker, same session invalidation, same lock as the replacement flow, minus the account creation.

Deliberate limits:

- **Opt-in.** Only `{"retire_bootstrap": true}` does it, so no other POST can retire a login as a side effect of some other intent.
- **The caller must already exist.** Otherwise the gate is "anyone `requireAuth` let through".
- **Not repeatable.** Once the marker names an operator, a second caller gets 409 and cannot take the administrator slot from them.
- **The bootstrap session keeps the replacement flow** unchanged.

⚠️ **Whoever runs setup claims the administrator slot.** That's the intended trade, and worth a second opinion: before this the slot was held by an account nobody could sign in as, which is strictly worse, and root can rewrite the marker to hand it elsewhere.

## The wizard asked the same impossible question

It rendered a create-account form that could only ever be refused. It now reads `retire_only` from the status and offers *"Close the temporary login"* instead, so it never presents an action the server will reject.

## Verification

Tests reproduce the appliance's exact state (`newStrandedApplianceServer`). **Plant-tested** — forcing `retireOnly` to return false fails them, so they pin the fix and not the scaffolding:

```
--- FAIL: TestEstablishedAccountRetiresStrandedBootstrapLogin
    a stranded appliance must offer retirement to laura:
    map["complete":false, "required":true, "retire_only":false, ...]
--- FAIL: TestRetirementIsNotRepeatable
```

`go test ./...` green · frontend build + 183 tests green.

`.210` is deliberately untouched — this lands first.


JBailes:
CI fixed — three failures, one of which explains why the job printed nothing.

**1. Link.** `unit-test-server-http` links `server_http_config_routes.o`, which now serves the CLI manifest:

```
undefined reference to `cli_marshal_defs_to_json'
```

Added `cli_marshal_defs.o` beside `cli_command_defs.o`/`cli_dispatch_defs.o` in **both** test targets that link that route — the second would have failed on the next shard boundary for the same reason. (`unit-tests`, the third failing job, is only the shards' aggregate.)

**2. Sessions.** `session.create` now answers `failed to create session` here. Not a regression: the db1/db2 conversion moved the sessions stage out of this process, `db1_client/sessions.c` checks `obs_bus_module_available()` first, and this harness starts `aimee-server` **alone** — module processes are launched by the container's module-supervisor, not by the server. The stage is absent by construction.

That composition is covered where it belongs: `unit-test-db1-module-bus` execs the real module binary against a live bus (*"a failure to start the module is a failure of this test, never a skip"*), and the Docker E2E matrix runs the supervised stack. So probe once and **skip**, the same idiom this file already uses for aimee-kb. The probe is the switch, not a hardcoded expectation — when the stage answers here these run as ordinary checks with no edit.

**62/62 pass, 20 skipped.**

**3. And why CI showed nothing.** The harness counts failures but runs under `set -e`, so one unguarded `VAR=$(cmd)` aborts the entire run at the assignment — every later check *and* the summary. That is exactly what happened: `SID=$(... json.load ...)` failed on the error response, and a job that built for four minutes printed no FAIL line, no count, nothing to read.

Fixed as the pattern it is rather than the one line: **40** capture-for-assertion sites now end `|| true`, so an empty value fails its own assertion visibly. The four that establish the fixture itself still abort, because continuing past them would test nothing.

And a floor under it — an EXIT trap:

```
integration: ABORTED after N checks (exit 1).
  The harness exited before its summary — under 'set -e' an unguarded
  command failed, so the checks below this point never ran.
```

Plant-tested with an unguarded `false` after the trap. Silence is no longer a possible outcome.

Verified: integration-tests 62/62 + `thin-client-remote-exclusive` · unit-tests green · lint 60/60 · build-integrity.

(`unit-test-vault-bootstrap` fails locally only under an environment carrying `CLAUDE_CODE_MESSAGING_TOKEN` — passes with it unset, pre-existing, and green in CI.)

PR #2794, opened 2026-08-18
feat: serve the no-arg request shapes, and gate the integration harness

JBailes:
Two commits that missed #2789 — it merged at `e932a08` before these were pushed, so the comments describing them landed on a closed PR and never got a CI run.

## 1. Serve the no-argument request shapes — the chain is closed

With routes, catalogue and dispatch served, a new server command was findable, documented and addressable, and still wouldn't run:

```
aimee: 'brandnew.thing' — arguments are missing or invalid, so no request was sent.
```

The client refuses to send a body it can't build, and had no row saying this method needs none. For the no-arg set that row is pure data, so it's served: `marshal` rows of `{method, args:"none"}` in the CLI manifest.

**Proven end to end.** `aimee brandnew thing` — absent from this client's dispatch rows, catalogue *and* no-arg list — POSTs `/v1/brandnew/thing` and exits 0 against a stub serving only those rows.

### Deliberately only the no-arg set

Surveying all 42 marshaller functions:

| | count | |
|---|---|---|
| map argv → fields | 29 | servable once a spec carries types |
| + types / nested shapes | 5 | same |
| **read local disk / env / exec** | **8** | **should never be served** |

Those 8 (`marshal_git_cli`, `marshal_skill_request`, `marshal_index_file_request`, `marshal_identity_snapshot`…) read the **client's own** filesystem and environment. That isn't knowledge of what the server can do — a thin client reading local state in order to send it is precisely its job. Serving them would be *wrong*, not merely incomplete.

What remains coupled is **34 argv→field mappings**, which need a spec language — a design, not a move.

Only an explicit `"args":"none"` makes the client send an empty body; an unknown or richer shape leaves the decision to its own marshaller rather than guessing `{}` for a command that needed arguments. `test_served_marshal` pins that.

## 2. The co-location ban had broken the integration harness

`make integration-tests` is a make target with **no CI job**. Running it:

```
FAIL: api status reports the listener   (got 'aimee: aimee-server unavailable')
FAIL: api enable turns the listener on  (got 'aimee: aimee-server unavailable')
FAIL: mcp git_status tool call          (got 'aimee-server unavailable after retries')
...
```

…against a server the harness had started and which was running perfectly well two lines above them. It reached it by **well-known-socket discovery**, which the co-location ban removed. Fixed the same way `test_init_migrate_service.sh` already was — an explicit `AIMEE_API_ENDPOINT`, because a harness deliberately running both halves on one box now has to *say* so.

**63/73 → 73/73**, and it's gated now: an ungated harness is one that will be wrong when you next need it.

One of the ten was stale rather than broken:

```
"no server: DB command fails without autostart"  expecting  "server unavailable"
```

The command still fails, which is the property that matters. It now fails one step *earlier* — `cannot reach aimee-server to ask what commands it has` — because the client can't learn `memory list` exists without the server. That's the command table living server-side, i.e. exactly what commit 1 does.

### Dead names swept

`AIMEE_NO_AUTOSTART` / `AIMEE_DISABLE_AUTOSTART` — nothing reads either, since the client starts no server to suppress — yet two harnesses still exported them and the reference-doc table still described one.

`build-integrity` checked `windows/cli_client.c` **by path**, which is the very argument its own comment makes against naming one platform. Now tree-wide over non-test C sources, and **plant-tested**: a planted `getenv("AIMEE_NO_AUTOSTART")` in `cli_main.c` fails it.

## Verification

`make all` · `make lint` 60/60 · `test_cli.sh` 51/57 (baseline) · `integration-tests` 73/73 incl. `thin-client-remote-exclusive` · `init-migrate-service` 4/4 · build-integrity · docs-gen in sync.


JBailes:
Fixed the two failures here — they were the same two I'd already fixed on #2793, which is stacked on this branch, so the fix existed but not on this side of the stack. Cherry-picked `9001b0e` down and merged `testing` in.

Both were caused by this PR's own commits:

- **`unit-test-server-http` link error.** This PR makes `server_http_config_routes.o` serve the marshal rows, so the test target needs `cli_marshal_defs.o` — added to **both** targets that link that route.
- **`integration-tests`.** The job this PR adds was failing, and printing *nothing at all* while doing it. Two causes: `session.create` now needs the DB1 sessions stage, which the db1/db2 conversion moved out of process (this harness starts `aimee-server` alone, so it's absent by construction — probe-and-skip, the idiom this file already uses for aimee-kb); and the harness runs under `set -e`, so one unguarded `VAR=$(cmd)` aborted the whole run at the assignment, summary included. 40 capture sites guarded, plus an EXIT trap so an early abort says so instead of producing silence.

**62/62 pass, 20 skipped.** Verified locally: `make all` · `make lint` 60/60 · integration-tests · build-integrity · the previously failing link target.

Full detail in the commit message and on #2793.

PR #2799, opened 2026-08-18
feat: serve the argument shapes, so a new command's flags need no new client

JBailes:
Stacks on #2794.

The manifest could already say a command exists, how it routes, what it's called, and which methods take no body. It could **not** say how the words after the command become fields — so a new command that took arguments still needed a new client. This is that last piece.

```
$ aimee widget tune w-42 --torque 7 --dry-run
```

`widget`, `tune`, `--torque` and `--dry-run` appear **nowhere** in the client binary. It POSTs `/v1/widget/tune` with exactly

```json
{"method":"widget.tune","protocol_version":1,
 "widget_id":"w-42","torque":7,"dry_run":true}
```

and exits 0, from a served spec alone. With no id it prints the **spec's** usage line and sends nothing — a served command has to misuse as loudly as a compiled one, or an operator can't tell a typo from an outage. `scripts/test-served-argspec-e2e.py`, wired into `integration-tests`.

## The spec

A bool-flag list for the parser, then ordered fields, each naming where its value comes from and what JSON it becomes:

```json
{"json":"open_weights_only","from":"flag","flag":"open-weights","type":"bool"}
{"json":"name","from":"positional","index":0}
{"json":"id","from":"positional_or_flag","index":0,"flag":"id","required":true}
```

Four sources, four types. The vocabulary was read off the marshallers that exist rather than designed for ones that might, which is why it's this small.

**What it deliberately cannot express:** reading the client's filesystem or environment, composing prompts, or cross-field rules like *"either `--task` or `--proposal`"*. The first two are the thin client's own job and must not move server-side at all. The third is a judgement about what the user meant, and a spec language that grew conditionals would become a program transmitted over the wire. Those methods keep their compiled marshaller.

An unrecognised anything — source, type, field shape — refuses the **whole** spec and falls back to the compiled path. A partially honoured spec builds a request that is wrong rather than incomplete, and the server would answer it as though the operator had asked for that. Sixteen such specs are pinned as refused, including `"from":"read_file"` and `"from":"env"`, which are the ones that would matter.

## Why this is safe

A served spec **wins** over the compiled marshaller. That's the point, and it's also the risk: every one of these methods works today, and a merely plausible spec would change what the CLI sends for commands people already run — silently, because both bodies are valid JSON and the server answers either.

So the test is **differential, not descriptive**. For each shipped spec it runs the *real* marshaller and the interpreter over the same argv and compares rendered JSON; a spec ships only once it's indistinguishable. Samples are deliberately awkward — flag absent, flag present, positional instead of flag, empty string, unknown extra flag — because the cases that differ are the edges.

Two integrity properties, both plant-tested:

- The test **includes the same data file the server serves from**. A test with its own copy would prove specs that are not the ones shipped. Renaming one output field (`available_only` → `available`) fails it.
- **Every shipped spec must have samples**, or the test fails naming it. A spec added tomorrow cannot ship with no evidence behind it.

Six methods converted (`provider.list`, `provider.models`, `catalog.list`, `trigger.list`/`status`/`cancel`) — enough to exercise every source and type. The rest follow one differential sample at a time; the mechanism, not the tranche, is what this PR is for.

## And a guard — which I got wrong twice

Every object the served-manifest work added broke the test link lists identically:

```
undefined reference to `cli_dispatch_defs_to_json'
undefined reference to `cli_marshal_defs_to_json'
undefined reference to `cli_argspec_defs_to_json'
undefined reference to `cli_argspec_build'
```

Four times, always found minutes after a push, because `tests/Rules.mk` keeps its prerequisites by hand and nothing relates them to what the code calls. This one needed edits to **twelve** targets.

`check-manifest-emitter-linkage` derives it: read the functions each small provider object exports, find which sources call them, require every target depending on a caller to link the provider.

It took three attempts, and both failures are the same lesson:

1. **v1** understood only the manifest route's emitters. Passed → next run failed on `cli_argspec_build` one file over.
2. **v2** generalised to caller→provider edges. Passed → next run failed on the *same symbol again*: two tests `#include` `cli_v1_routes_b.c` to reach a static marshaller, so they never link its object, and a check looking only for the object couldn't see them.

Both times it said ok while the build broke — the definition of decoration. It now treats "`#include`s the caller's source" as a dependency too, and the plant test exercises that path specifically rather than only the linking one, because a plant that only proves the case you already thought of is the same mistake a third time.

## Verification

`make all` · `make lint` **61/61** · `make unit-tests` · `integration-tests` **62/62** · the differential suite · the end-to-end proof.

(`unit-test-vault-bootstrap` fails locally only under an environment carrying `CLAUDE_CODE_MESSAGING_TOKEN` — passes with it unset, pre-existing, green in CI.)

### DIFF
--- a/.github/workflows/module-inventory.yml
+++ b/.github/workflows/module-inventory.yml
@@ -131,8 +131,18 @@ jobs:
       # and fails until the baseline is refrozen, and THAT diff is the useful
       # artifact: one review of everything a release changes on the public
       # surface, instead of nineteen refreezes read by nobody.
+      # Integration bases, not just `testing`: agent/<topic> and aimee/feat/<id>
+      # are stacked onto the same way (see ci.yml's pull_request branch list), so
+      # a sub-PR based on one hit this release gate while its eventual base does
+      # not. Both escapes were wrong -- re-freeze, which the refusal below then
+      # blocks on the way to `testing`, or leave the PR red.
       - name: Enforce refactor surface baselines
-        if: github.base_ref != 'testing' && github.ref != 'refs/heads/testing'
+        if: >-
+          !startsWith(github.base_ref, 'agent/')
+          && !startsWith(github.base_ref, 'aimee/feat/')
+          && github.base_ref != 'testing'
+          && github.ref != 'refs/heads/testing'
+          && !startsWith(github.ref, 'refs/heads/agent/')
         run: python3 -I -S scripts/refactor_baselines.py
       # The surface baseline is a RELEASE gate, enforced on `main` by the step
       # above and deliberately not here. That makes re-freezing it on a `testing`
@@ -149,7 +159,10 @@ jobs:
       #
       # Adding a public header therefore does not require touching this file.
       - name: Refuse a baseline re-freeze on the integration branch
-        if: github.base_ref == 'testing'
+        if: >-
+          github.base_ref == 'testing'
+          || startsWith(github.base_ref, 'agent/')
+          || startsWith(github.base_ref, 'aimee/feat/')
         run: |
           set -eu
           git fetch --no-tags --depth=1 origin "$GITHUB_BASE_REF"
@@ -158,7 +171,8 @@ jobs:
             exit 0
           fi
           cat >&2 <<'MSG'
-          ::error::This PR changes tests/baselines/refactor/index.json, and it targets `testing`.
+          ::error::This PR changes tests/baselines/refactor/index.json, and it targets the
+          integration branch `${{ github.base_ref }}`.
 
           The public-surface baseline is a release question. It gates `main`, not this
           branch, so re-freezing it here achieves nothing and collides with every other
@@ -166,10 +180,10 @@ jobs:
 
           Revert the file and push again:
 
-              git checkout origin/testing -- tests/baselines/refactor/index.json
+              git checkout "origin/${{ github.base_ref }}" -- tests/baselines/refactor/index.json
 
           You do NOT need to re-freeze after adding or changing a public header. The
-          baseline is expected to be stale on `testing`; the promotion PR into `main`
+          baseline is expected to be stale on an integration branch; the promotion PR into `main`
           re-freezes it once, and that single diff is the review of what the release
           changed on the public surface.
           MSG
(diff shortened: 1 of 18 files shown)


JBailes:
Second commit: two more specs — `model.episodes` and `graph.sync_code`. **8 shipped specs, 29 differential samples.**

`graph.sync_code` is sampled with `--project aimee` on purpose: it's a *plain* positional, so that flag must **not** become the field. `positional_or_flag` deliberately falls back and `positional` deliberately doesn't, and nothing else in the suite would have caught the two being confused.

### Why I stopped at eight

I read every remaining marshaller expecting to convert them in batches, and didn't. **The blocker isn't the spec language — the marshallers disagree with each other**, and a spec reproducing each disagreement would stop being data and start being a program:

| | examples | why it can't just be described |
|---|---|---|
| **Lenient numbers** | `aux.test`, `dogfood.tag`, `api.enable` | `atoi()`/`cli_args_get_int()` turn `"12x"` into 12 and `"abc"` into 0. The spec's `number` *refuses* rather than coerces — as `kb.grant` already does, because rounding a team id administers the wrong team. Serving these means **fixing** the coercion. |
| **Missing envelope** | `eval.results`, `dogfood.tag` | Hand-built bodies that omit `protocol_version`, which every other request carries. Found by the differential test on the first attempt to convert one — exactly what it's for. |
| **Derived fields** | `catalog.show`, `workspace.add` | Splits `provider:model` in two; inverts `--no-scan` into `"scan": false` and nests part of the body under `args`. |
| **Cross-field rules** | `cron.*`, `trigger.fire`, `dogfood.tag` | `--all` valid for two of five cron methods; `--task` OR `--proposal`; `--surprise`/`--no-surprise` exclusive. Judgements about what the operator meant, not shapes. |
| **Local state** | `eval.run`, `identity.snapshot`, +6 | `getcwd()` and the filesystem. Must never be served whatever the spec can express. |

Recorded in the data file's header, where the next person converting one will look.

So the remaining work is **normalising the marshallers**, and it now has a safety net: convert one, add its samples, and the differential test proves the request body didn't change. Widening the spec vocabulary instead would make every future client interpret one.

`make all` · lint 61/61 · integration-tests 62/62 · differential suite · end-to-end proof.

PR #2812, opened 2026-08-18
ci: give jobs a hang bound, sized to what each workflow does

JBailes:
Found while waiting on #2807: five jobs sat `in_progress` for **three hours**, with `lint` stuck the whole time on

```
running step: Install clang-format  since 15:56:42
```

— an apt-mirror hang with nothing to do with the change under test. **No job in `ci.yml` had a timeout.** None of the 24. So GitHub's 6-hour default applied, and the PR would have been held for three more hours by a package mirror.

Only `module-inventory.yml` capped anything, which is how `descriptor-envelope` got noticed at all.

## Measured

Slowest job across three green CI runs on `testing`:

```
 6m33s  e2e-docker (T2)      2m59s  build-integrity
 6m01s  e2e-docker (T1)      2m25s  lint
 5m16s  unit-tests (1/2)     0m04s  change-scope
```

30 minutes is ~4.5x that. **Uniform and generous on purpose:** tuning each job to its own runtime is exactly how `descriptor-envelope` ended up cancelled at 5m16s against a 5m cap, passing or failing on runner weather. This bounds a *hang*; it is not a performance budget.

## The measurement changed the answer

Fourteen other workflows had no caps either — and a uniform 30 would have **broken two of them outright**:

| workflow | longest successful run |
|---|---|
| auto-release | **1257m** — waits on a human gate |
| publish-llm | **34m** |
| publish-embedder | 13m |
| publish-testing | 12m |
| main-merge-approval | 12m — waits on a human gate |
| publish-images | 8m |
| everything else | <5m |

So publish workflows get **90** (~2.6x `publish-llm`), checks get **30**, and three are deliberately left alone:

- **`auto-release`** and **`main-merge-approval`** *wait*, on a human. Their long runs are the gate working, and a timeout would fail a release for being reviewed slowly.
- **`benchmark`** has no successful run recorded, so there's nothing to size a cap against — and a guess there is the mistake this whole change is about.

**46 of 53 jobs now bounded**, and the seven that aren't are named with a reason rather than left to be discovered.

Every workflow still parses — the generator refuses to write one that doesn't — and the caps are asserted present per job.


JBailes:
Re-checked the cap against a wider sample, because I've been burned twice by a number chosen from too few runs (descriptor-envelope was cancelled at 5m16s against a 5m cap sized the same way).

14 runs this time, **including failed ones** — a slow job that then failed is still evidence of how long the work can take:

```
  7m46s  success    e2e-docker (T2)
  6m55s  success    authz-residual-live
  6m33s  success    unit-tests (1/2)
  6m10s  success    lint
  5m50s  success    unit-tests (2/2)
```

Longest is **7m46s**, so 30 minutes is **3.9x** — still comfortable.

Worth noting the variance the wider sample exposed: `lint` came in at 6m10s here versus 2m25s in the original three-run sample. That's a 2.5x spread on the same job, which is exactly why a cap sized to "a bit above typical" fails on runner weather. 30m sits well clear of the whole spread.

PR #2814, opened 2026-08-18
feat(db2): add orphaned-L0 retention sweep event operation

JBailes:
feat(db2): add orphaned-L0 retention sweep event operation

Move db2_memory_prune_orphaned_l0 onto the typed event bus as
memory/prune_orphaned_l0 (operation 23), completing the pair with
memory.orphaned_l0_count: that operation counts the unlinked short-term rows,
this one deletes the ones that have aged out.

The tier and the seven-day window are fixed policy and carry no payload, so the
request is an empty envelope. That is deliberate. The backend statement is a
bare DELETE with the window interpolated, and an operator-supplied age or tier
on the wire would let a caller widen the sweep to the whole table. Nothing about
this operation needs to be caller-controlled, so nothing is.

The reply is the affected-row count. The backend returns -1 on a connection or
statement failure and the count otherwise, so the handler treats a negative
value as INTERNAL rather than reporting an empty sweep -- a failed delete and a
delete that matched nothing are the same number to a caller that only checks for
zero, and they are not the same event.

Classified single-statement and safe. The safe claim is the one worth proving
rather than asserting, so the packaged replay calls it twice against the real
process and requires both to return zero: the window is relative to now, and a
sweep that deleted on replay would not be observationally neutral. The contract
test pins the empty sweep as OK-with-zero, the -1 fault as INTERNAL, and the
declared bound as the encode limit, so a count past it is a fault instead of a
truncated reply.

Reuses db2-envelope-u32-v1 rather than introducing a wire format; the Go side
gained the matching codec pair and a check that both halves of the compiled-in
policy agree with the C header, since a drift there would have the two
implementations sweeping different rows.

Ledger: 664 externally consumed declarations pending review, now 663.


JBailes:
Consolidating this work onto the long-lived feature branch agent/db2-s3-bus-migration. This operation's commit is carried there unchanged; nothing merges to testing until that branch is complete.

PR #2815, opened 2026-08-18
fix(server): give the NDJSON surface an access line, with its duration

JBailes:
## Why

Every HTTP request gets an access line. The NDJSON socket that carries `mcp.call` and every other RPC logged **nothing at all**.

A call that stalls there is invisible twice over: the client reports only that it gave up, and the server log shows a healthy, idle process because it never recorded being asked. It reads as "the server is down" and sends the reader to the wrong subsystem.

That is not hypothetical. The integration harness has been failing its MCP git_status check with `aimee-server unavailable after retries`, diagnosed by inference repeatedly, while the server log showed a clean run with a 33-second hole in it and no request to explain the hole.

## What it showed on the first run

The two slowest calls were finally visible: an mcp.call taking 17727ms, and a hooks.pre taking 8581ms.

### DIFF
--- a/src/server/server.c
+++ b/src/server/server.c
@@ -1708,9 +1708,7 @@ static size_t method_size_limit(const char *method)
    return LIMIT_DEFAULT;
 }
 
-/* An RPC slower than this is worth an access line on its own. Below it the
- * line is DEBUG: the ordinary trace stays available without every healthy
- * call competing with the one that stalled. */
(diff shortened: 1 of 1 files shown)


JBailes:
Superseded by #2818, the integration branch for this work. The commits are carried over unchanged and verified together there: server build clean, full unit suite passing, lint 61 of 61, integration harness 62 of 62. Closing here so review happens in one place.

PR #2816, opened 2026-08-18
feat(db2): add pending time-to-live archival event operation

JBailes:
feat(db2): add pending time-to-live archival event operation

Move db2_memory_lifecycle_sweep_expired onto the typed event bus as
memory/lifecycle_sweep_expired (operation 24). It archives the pending memory
rows whose time-to-live has passed, which is the lifecycle counterpart to the
retention delete added as memory.prune_orphaned_l0.

Both lifecycle states and the archive reason are fixed policy and carry no
payload, for the same reason the prune carries none: the backend statement is a
bare UPDATE, and letting a caller name the source state, the target state, or
the reason would let it relabel or archive rows the reviewed operation never
covered.

WHAT THIS OPERATION CANNOT REPORT. The backend returns zero for a connection or
statement failure and zero for an empty sweep, so across the bus a fault is
indistinguishable from nothing to do. That is the existing contract of the
symbol and of the already-migrated promote_stable and demote operations, so it
is preserved here rather than quietly changed under cover of a bus migration --
tightening it is a real behaviour change and belongs in its own reviewable diff.
The handler still refuses a negative count, so the wire side is already correct
if that contract is later tightened, and the contract test pins zero as a
success specifically so the limitation stays visible to the next reader instead
of being mistaken for a fault path that works.

Classified single-statement and safe. The packaged replay calls it twice and
requires both to return zero, since the comparison is against now and a sweep
that archived on replay would not be observationally neutral.

The adapter gained c/memory_lifecycle.h, which is where the symbol is declared;
the source boundary is unchanged at 994 inbound includes because DB2's own
adapter including DB2's own header is not an outside consumer.

Also fixes the generated client emitter to produce clang-format-clean output for
this call signature. The template wrapped the return type onto its own line,
which the formatter rejects, and a generated file cannot be fixed in place.

Ledger: 663 externally consumed declarations pending review, now 662.


JBailes:
Consolidating this work onto the long-lived feature branch agent/db2-s3-bus-migration. This operation's commit is carried there unchanged; nothing merges to testing until that branch is complete.

PR #2817, opened 2026-08-18
feat(ir): measure what a turn actually did, from the IR

JBailes:
## Why

Token counts say what a session cost, never what it was doing. A turn that read the same file four times and changed nothing bills the same as one that found the defect and fixed it -- so the transcript that separates them is discarded at exactly the moment it could be summarised.

## What

`aimee_ir_session_measure` reads the parsed request and derives four things from the ordinary transcript:

- how many tool calls the turn made
- how many repeated an earlier call verbatim (same name AND arguments)
- which unproductive shape it fell into, if any
- whether its tools arrived over CLI, MCP or both

Those land on the ingress request, so the token-audit row that already records cost now records behaviour beside it.

Deriving this from the IR rather than a wire format is the point: one implementation covers OpenAI, Anthropic and Responses, and every client behind them, instead of one per ingress.

The `session_assist` stage runs last, after memory and the shell block, so it judges the transcript that will actually be sent rather than the one that arrived.

## Tests

Thresholds are the product behaviour here, so each is pinned at its boundary rather than somewhere comfortably past it:

- four reads with no edit
- an edit followed by three non-test calls
- a repeat of an identical name and arguments

It also pins the healthy shape the last of those exists to contrast with: an edit that IS followed by a test must not be flagged. I verified the test fails when a threshold is disabled.

## Scope

This is the client-neutral **measurement** half only. The persona placement that shares this module upstream needs a durable session field, and the storage boundary it was written against has since moved to `db1_client` with that migration still in flight. It stays out until that settles.

## Verification

- server build clean, clang-format clean
- full unit suite: All tests passed
- lint: all 61 checks passed


JBailes:
Superseded by #2818, the integration branch for this work. The commits are carried over unchanged and verified together there: server build clean, full unit suite passing, lint 61 of 61, integration harness 62 of 62. Closing here so review happens in one place.

PR #2818, opened 2026-08-18
Model-neutral session quality: RPC visibility and per-turn measurement

JBailes:
Integration branch for the model-neutral session-quality work. Individual PRs are closed in favour of this one; each commit here is the reviewed content from those branches.

## NDJSON access line (was #2815)

Every HTTP request had an access line; the socket carrying mcp.call and every other RPC had none. A call that stalled there was invisible: the client reported only that it gave up, and the log showed an idle server because it never recorded being asked. Ordinary calls log at DEBUG, slow or failed ones at INFO.
## Session measurement from the IR (was #2817)

Token counts say what a session cost, never what it was doing. A turn that read the same file four times and changed nothing bills the same as one that found the defect and fixed it. This derives tool calls, verbatim repeats, the unproductive shape if any, and whether tools arrived over CLI or MCP, and puts them on the ingress request beside cost.

## Session binding before the client starts (was #2806)

Already merged upstream, included here for completeness.

## Verification

Server build clean, full unit suite passing, lint 61 of 61, integration harness 62 of 62.


JBailes:
## Validation status — read before merging

Earlier comments in this session described this branch as ready on the strength of
local tests plus CI. That was premature: no deployment validation had been done at
that point. This records what is and is not actually verified.

### Verified
- Local: server + client build, full unit suite, 61/61 lint, integration 78/78,
  thin-client remote-exclusivity.
- CI: 30/30 checks success on the recent commits (build, unit, integration,
  descriptor-envelope, e2e-docker T1-T3, windows/macos).
- Live deployment on a Debian 13 LXC with PostgreSQL 17 + pgvector, running this
  branch's binaries:
  - `index find` / `index callers` / `index structure` return real hits against a
    freshly indexed project.
  - `index deps` with no project argument resolves the project from cwd (the
    self-scoping fix), which previously failed with "missing project".
  - `index investigate` reports `retrieval_path: [context, hybrid]` with
    `fallback_used: true`, and distinguishes `result_status: empty` from the typed
    outage verdict (`dependency: kb`, `retryable: true`) — both observed live.

### Found and fixed only by deploying
`aimee-kb --bootstrap-db2` — the documented first-run provisioning command — failed
with "embedding dimension 0 is unusable". `bootstrap_db2_try_url()` called
`db2_init()` without supplying the deployment's embedding width; every other entry
point does. Pre-existing on `testing`, not introduced here, and not reachable by any
unit test because it needs a live Postgres. All six `db2_init` call sites were swept;
this was the only one.

### NOT verified
`index investigate` returning non-empty evidence. Two environment dependencies block
it on the test host, neither in this branch's code:
- code-unit extraction requires `SYNTHESIS_ENDPOINT` (an LLM);
- vectors require a real embedder (`embedder_command: builtin` is an internal
  fallback, not a valid config value — it is exec'd and fails with 127).

Consequently `investigate_attach_code` is the one landed path never exercised against
real data. It has unit coverage; it has no live confirmation.

JBailes:
## Correction: the success path is now verified

My previous comment listed `index investigate` returning non-empty evidence as NOT
verified. That gap is now closed. Retested on the same LXC after standing up a
deterministic hashing embedder speaking the gateway contract (`/embed`,
`/embed_batch`, `/health`).

A symbol-shaped query returns, live:

- `result_status: "ok"` at both row and envelope level — the third verdict,
  completing empty / outage / ok.
- `retrieval_path: ["context"]` with `fallback_used: false` — the fallback correctly
  does not fire when context answers. The complementary branch
  (`["context","hybrid"]`, `fallback_used: true`) was observed earlier, so both are
  confirmed.
- `code` populated with real file content plus `code_line_start` / `code_line_end`
  (`investigate_attach_code`).
- `systemic_scope` populated with real graph data: `blast_radius` showing
  `dates.py` depended on by `billing.py`, `direct_callers` naming
  `invoice_window:5`, plus `likely_same_fix` and `analogous_implementations`
  (`investigate_attach_scope`).
- `answerability.decision: "answerable"`, read correctly by
  `investigate_result_answerable()`.

Every behaviour this branch changes in the investigate path is now confirmed against
a live deployment, not only unit tests and CI.

Still out of scope on that host, and unrelated to this branch: the curator LLM
extraction lane (`SYNTHESIS_ENDPOINT`) and therefore semantic vector retrieval over
code units. Structural, exact and graph retrieval are what answered above.

JBailes:
## Validation complete

Follow-up to the two earlier status comments. Every area this branch changes now has
either live deployment evidence or plant-verified deterministic coverage.

### Verified on a live deployment
Debian 13 LXC, PostgreSQL 17 + pgvector, aimee-kb, aimee-server, the DB1 module under
the module supervisor, a stub embedder speaking the gateway contract, and a logging
provider backend.

- `index investigate`: all three verdicts observed live (ok / empty / typed outage
  with `dependency: kb`), `retrieval_path` on both branches, `fallback_used` true and
  false, real source attached via `include_code`, `systemic_scope` with a populated
  `blast_radius` and `direct_callers`.
- `index find` / `callers` / `structure` return real hits; `index deps` resolves its
  project from cwd with no project argument.
- The same investigation over **MCP** returns the identical typed packet, which is the
  parity fix: that surface previously had no verdict at all.
- MCP schema: `delegate` advertises via / scope / handoff_json with the correct scope
  enum; `spans` carries its item schema.
- Persona delivered **exactly once per session**: turn 1 carries the block, turns 2
  and 3 do not, and the durable row settles at state 1.
- `persona_delivery_state` present in a real DB1 created by the running server.
- A session-suffixed bearer authenticates in both `x-api-key` and `Authorization`; a
  wrong bearer still gets 401.
- Generation controls forwarded to the provider (`max_tokens` 321, `temperature` 0.42).
- `workspace prepare` over HTTP registers the workspace.
- Responses shorthand string input reaches the provider on the Responses wire.

### Verified by deterministic tests, each plant-verified
- SSE keepalive: a new test stalls a generation past the interval and asserts the
  comment frame; disabling the keepalive thread fails it.
- `default_delegate` routing: existing assertions cover the preference winning, being
  skipped when disabled, and being skipped when it cannot satisfy the capability gate.
  Disabling both preference blocks fails them.

### Bug found only by deploying
`aimee-kb --bootstrap-db2` failed with "embedding dimension 0 is unusable" because
`bootstrap_db2_try_url()` called `db2_init()` without supplying the deployment width,
which every other entry point does. Pre-existing on `testing`, unreachable by unit
tests. Fixed here; bootstrap now reports `knowledge_ready: true`.

PR #2819, opened 2026-08-18
feat: served argument specs (18 methods), envelope normalisation, and CI hang bounds

JBailes:
Consolidates #2809, #2811 and #2812, and continues the thin-client work on top. #2807 was closed as superseded (the same `descriptor-envelope` fix landed on `testing` independently, with a better cap and a sharper observation — the cancellation reports as "Cancelled after 5m", which reads like a concurrency cancel rather than the timeout it is).

## Served argument specs: 9 → 18

The client can now be told how a command's *arguments* become fields, not just that the command exists. **18 shipped specs, 69 differential samples**, every one proven byte-identical to its compiled marshaller.

The interesting part is how the vocabulary grew, because I got it wrong first and the correction is the same shape twice:

| convention | sites | what the spec described |
|---|---|---|
| positional emits empty value | **81** | ✗ |
| positional drops empty value | 2 | ✓ |
| `atoi` / `cli_args_get_int` | **53** | ✗ |
| strict, refuses trailing garbage | 3 | ✓ |

I refused to add the first flag, arguing it would *"enshrine a convention that looks like a bug."* Then I counted. A vocabulary that can express 2 of 83 sites isn't taking a principled stand — it's incomplete. Describing the code faithfully and reforming it are different jobs, and **only the second is a decision for you**.

That unblocked nine methods: `eval.results`, `cert.revoke`, `vault.capability`, `vault.delete`, `vault.set`, `vault.set_server`, `graph.explain`, `aux.test`, `dogfood.review`.

Every one is sampled with the input that *is* the difference — `""`, `abc`, `12x` — because a spec agreeing only on well-formed input diverges silently on the first typo.

### Where I stopped

- **Conditionals.** `api.enable`'s `if (port > 0)`, `dogfood.tag`'s exclusive `--surprise`/`--no-surprise`. A spec language that grows these stops being data and becomes a program shipped over the wire.
- **Local state.** `eval.run`, `identity.snapshot`, `identity.diff`, `vault.unlock`, `delegate.launch` read the client's own disk, environment or key material. These must **never** be served.
- **`provider.set`** reads raw `argv[0]`, so `provider set --json openai` sends `name="--json"`. One method — the 53-vs-3 argument doesn't apply to a sample of one, and describing it would enshrine what looks like a bug.

## Envelope normalisation

Seven marshallers hand-built their body and omitted `protocol_version`. Inert on the wire (the server only ever *emits* it), but it's what made them undescribable — a body missing a field every other body has can't match one built from a spec.

## The parity report was blind by construction

`command_registry` exists so a name resolves one way. Its work-list report did `out[method] = (group, verb)` — last writer wins — so aliasing was structurally unrepresentable. It now surfaces **12 methods reachable under two group names** and two identical group pairs (`agent`≡`model`, `server`≡`status`), and reports 190 spellings against 166 methods. The gap between those numbers *is* the aliasing.

## CI hang bounds

No job in `ci.yml` had a timeout — five sat `in_progress` for three hours, `lint` stuck on `Install clang-format`. **46 of 53 jobs now bounded**, measured: 30m for checks (~4x the slowest of 7m46s across 14 runs), 90m for publish. Three left uncapped deliberately — `auto-release` and `main-merge-approval` *wait on a human*, and `benchmark` has no successful run to size against.

## Verification

`make all` · lint 61/61 · integration-tests 62/62 · unit-tests · differential suite · the end-to-end proof that a command with arguments absent from the client still runs.

(`unit-test-vault-bootstrap` fails locally only under an environment carrying `CLAUDE_CODE_MESSAGING_TOKEN` — `rc=0` with it unset, pre-existing, green in CI.)


JBailes:
Green: **37 pass, 0 fail**, `mergeable=CLEAN`.

One addition since the description was written, and it came out of this PR's own first CI run.

Two jobs were killed at exactly the 30-minute cap this PR adds:

```
unit-tests (1/2)   pass    3m30s
unit-tests (2/2)   KILLED 30m20s   ← in "Install dependencies"
pam-login-live     KILLED 30m17s   ← in "Install dependencies"
```

I'd claimed the cap "bounds a hang, not a performance budget," so I checked rather than assumed. Same job, same packages, same content — one finished in three and a half minutes, its sibling hung for thirty. **The cap worked as designed**: what cost three hours earlier tonight cost thirty minutes here.

But thirty minutes is still thirty minutes for an install that takes seconds, so the last commit fixes the cause. The existing mitigation retries a mirror that *fails*:

```sh
i=0; until sudo apt-get update -q; do ...retry...; done
```

and is powerless against one that *hangs* — `until` never retries a command that doesn't return. `timeout` converts the hang into a failure the loop already handles. Every apt call across every workflow is now bounded: 16 update loops, 3 single-line installs (now retried too), 16 multi-line installs.

Both previously-killed jobs pass on the re-run.

Verification note: I checked the YAML parses **and** ran `bash -n` over every apt-bearing `run:` block — 22 steps, 0 syntax errors. Valid YAML wrapping a broken shell fragment would have passed the first check and failed in CI.

JBailes:
Closing — this shouldn't have been a PR.

`agent/thin-client-and-ci-hardening` is a feature branch to keep merging into while the work continues, not something to merge out mid-flight. I opened this out of habit after consolidating #2809/#2811/#2812 into it, which contradicts what the branch is for.

The branch stays and work continues on it. It'll get a PR when the work is actually finished — which by my own account it isn't: the served-spec effort stops at a deliberate line (no conditionals, nothing reading local state, no vocabulary for a sample of one), and three decisions are outstanding:

- whether the 53 lenient `atoi` parses should refuse instead of coerce (`kb.grant` is the argument that they should)
- the three alias pairs blocking `command_registry` — `agent`≡`model`, `server`≡`status`, `delegate`/`ensemble`
- `.210`'s LLM routing

Nothing is lost by closing: the branch is green (37/37 when last run), and every commit is on it.

PR #2820, opened 2026-08-19
feat(db2): move nineteen memory operations onto the typed event bus

JBailes:
Moves nineteen DB2 memory-family symbols onto the typed event bus and closes ten
more by dispositioning them as private. The catalog goes 32 -> 51 operations and
the declaration ledger goes 664 -> 635 externally consumed declarations pending
review.

Every commit is one slice: one operation, or one reviewed disposition. Each was
built and verified locally before the next started.

## What is here

**Operations 23-41.** Retention sweeps (`prune_orphaned_l0`,
`lifecycle_sweep_expired`), single-row mutations (`demote_id`, `delete_row`,
`touch`, `link_delete`, `reject`, `decay_confidence`, `update_content`),
probes (`has_workspace_tag`, `has_scope_type`, `valid_at`), setters
(`workspace_tag_insert`, `set_cognified_kind`, `set_source_session`,
`negation_tokens_update`), and string read-backs (`get_content`,
`get_source_session`, `pick_first_temporal_ref`).

**Five reusable wire formats**, which is why the later operations are far
smaller diffs than the first: `u64-u32`, `u64-ack`, `u64-string-u32`,
`u64-string-ack`, `u64-string-reply`.

**Ten cascade deletes kept off the bus.** Every
`db2_memory_*_delete_for_memory` is the opening statement of a `memory_refresh_*`
rebuild. Publishing them individually would have added ten permanent wire
capabilities nothing wants, and split each rebuild across the process boundary
so a failure between the delete and the reinserts leaves a memory stripped of
its derived rows. Dispositioned private with that reasoning recorded, so the
next pass reads it instead of reaching the opposite conclusion.

## Judgment calls worth reviewing

These are the places a mechanical migration would have produced something that
compiles, passes every envelope test, and is wrong.

**`valid_at` carries `invalid_state`.** Its backend distinguishes "not in force"
from "could not evaluate" and says so in its own comment. Folding those together
would answer a question that was never evaluated. Both codecs refuse to encode a
verdict alongside `invalid_state`.

**`has_scope_type` deliberately does not**, despite sharing that exact wire
format. Its backend folds a fault into a miss, so claiming an unevaluated state
it cannot detect would be worse than carrying the limitation honestly.

**`get_content` distinguishes empty content from a missing memory**; a stored row
may legitimately hold `""`. `not_found` carries no payload at all, so it cannot
be read as an empty string. `get_source_session` and `pick_first_temporal_ref`
share the format and deliberately have no empty-ok, because their backends
cannot make that distinction.

**`reject` does not carry its reason.** The backend `(void)`-casts the parameter
and nothing above it persists one. A field that travels between processes reads
as a field that is stored; the adapter shim passes NULL and carries a comment
saying why, so it is not later "fixed" by reconnecting a dead parameter.

**Two setters accept an empty string and two do not**, on one shared format.
Empty clears the column for `set_source_session` and `negation_tokens_update`;
the other two backends refuse it. Copying one minimum onto the others would have
silently removed the ability to unset a field while telling the caller the write
succeeded. Each minimum is pinned in the gate with its reason.

**Bounds match the owning record, not the wire.** Content is bounded to
`memory_t.content`'s width, workspace to DB2's own workspace width, and so on. A
value that survives the wire but not the read-back is not an error -- it is
different data, silently.

**Three operations move confidence with different constants** (0.9, -0.1, 0.7).
All are pinned as binary64 bit patterns and compared as bits on both sides, and
the tests assert they differ from each other, so a constant copied between
operations fails a test rather than decaying memories at the wrong rate.

**Inherited limitations preserved, not silently changed.** Several backends
report a fault and an empty result identically. That is kept as-is and pinned in
the contract tests so it stays visible; tightening an error contract is a real
behaviour change and belongs in its own reviewable diff.

## Verification

Run locally after every operation, all green: 61 lint checks, both DB2 bus
tests, 81 contract failure-mode tests, 14 ledger tests, 59 link-closure tests,
20 boundary tests, the Go suite, DB3 portability, activation, module-test
registration, module docs, and `check-docs`. Plus a full `make -C src all` --
exit 0, zero errors.

**One gap.** The `db2-replay` job needs Docker and a live Postgres, neither
available on this machine, so its assertions were built and reasoned about but
only CI actually runs them. Where a replay assertion depended on schema
behaviour I verified the schema rather than guessing -- for example confirming
`pg_now_text(modifier)` is created during bootstrap before asserting a sweep
returns zero.

## Not in this branch

635 declarations remain pending. They need struct-returning and list-returning
operations -- `memory_t` reads, relation lists, bounded row scans -- which is a
different class of design work per operation than anything here, and the
composite `memory_refresh_*` rebuilds that own the ten cascade deletes.

## Caveat

The mandatory roundtable review has not run on any of this work, including the
already-merged #2803 that precedes it. Flagging that explicitly: it is a process
gap, not an oversight I can close from here.


JBailes:
Closing this: I opened it before the work was actually complete. 635 declarations remain pending and the feature branch agent/db2-s3-bus-migration continues to accumulate commits. Will reopen when the migration is finished, per the standing instruction that nothing merges to testing until the branch is complete.

PR #2871, opened 2026-08-27
Agent/temporal assertion learning loop

JBailes:


JBailes:
Closing in favour of two focused PRs against current `testing`.

This branch is 31 commits behind a base that moved 4462 files (+641k/-161k) since the merge base, including the `src/db2` to `src/modules/db2/c` relocation. Merging `testing` in produces **107 conflicts**, almost all modify/delete artifacts of that move rather than real disagreements.

The decisive problem is direction of travel: `testing` is now **ahead** of this branch on the memory authority model. `memory_authority_t` appears in 14 files on `testing` and in **none** here; likewise `MEMORY_PROVENANCE_FOR`. Resolving 107 conflicts from a branch that predates that model risks silently reverting it, and most of what this branch carries has already landed (`FACT_GATE_REJECT_SENSITIVE`, `memory_workspaces`).

What was genuinely unmerged has been re-landed unchanged in substance:

- #2872 — fix: keep retracted facts retracted across re-extraction (the `authority_rank` carry, 3 sites + 2 tests)
- #2874 — feat: reconcile code index snapshots (the phased begin/stage/seal/abort protocol and `index verify`)

#2865 was based on this branch and has been retargeted to `testing`, carrying the RLS proposal text with it so it stands alone.

Both re-lands pass `make build-integrity` and the full `make unit-tests` locally.

PR #2872, opened 2026-08-27
fix: keep retracted facts retracted across re-extraction

JBailes:
A user's deletion of a fact was undone by the next ingest.

## The defect

`db2_fact_mutation_assert` already consults the invalidated twin before re-asserting, and its reactivate gate compares the incoming actor against the row's `authority_rank`. On the retract path the gate was reading the wrong number.

`db2_fact_mutation_invalidate` never wrote the *retracting* actor's authority onto the row, so the row kept the original *asserter's* rank:

| step | actor | rank | result |
|---|---|---|---|
| assert `ivan works_for acme` | MODEL | 10 | row rank 10 |
| user retracts it | USER | 30 | invalidated, row **still rank 10** |
| offline drain re-derives the same triple | SYSTEM | 20 | 20 >= 10 passes, `invalidated_at` cleared, **fact resurrected** |

## The fix

Stamp the deciding authority on the transition to invalidated, so the gate compares against the retractor rather than the asserter.

This is deliberately narrow. Both rollback paths already raise the rank this way, and the operator review-reject path stamps `FACT_ACTOR_OPERATOR` — which is why the defect hid. Retract was the one remaining transition that did not record who decided.

`expire_candidates` is left alone: it is a SYSTEM TTL sweep over candidates, where recurrence is meant to revive them.

## Test

The assertion is the observable resurrection, not just the internal field: `ivan works_for acme` is MODEL-asserted, USER-retracted, then re-derived by SYSTEM from new evidence (a distinct `source_id`/`evidence_hash`, so the evidence-replay guard does not apply), and must stay invalidated.

## Provenance

Re-landed from `agent/temporal-assertion-learning-loop` (#2871), which is not merging: its conflicts against `testing` are mostly artifacts of the `src/db2` to `src/modules/db2/c` relocation, and it predates the `memory_authority_t` model `testing` now carries.

Scoped to what `testing` still lacks. The rollback half of the original change and its regression test are **already present** there, so this PR carries only the `invalidate` path and its test.

### DIFF
--- a/src/modules/db2/c/fact_mutation.c
+++ b/src/modules/db2/c/fact_mutation.c
@@ -1181,17 +1181,28 @@ int db2_fact_mutation_invalidate(const fact_actor_t *actor, const char *source,
       fm_state_t after = rows[i];
       fm_copy(after.lifecycle, sizeof(after.lifecycle), FACT_LIFECYCLE_INVALIDATED);
       fm_copy(after.invalidated_at, sizeof(after.invalidated_at), now);
+      /* Carry the retracting authority onto the row.  The reactivate gate in
+       * db2_fact_mutation_assert compares an incoming actor against the row's
+       * authority_rank, so leaving the original asserter's rank here would let
+       * the next extraction of the same text re-establish a retracted triple at
+       * the authority that first asserted it.  The two rollback paths already
+       * raise the rank this way; the retract path was the one that did not.
+       * The selector loop above already refused any row outranking this actor,
+       * so this only ever raises. */
+      after.authority_rank = (int)actor->rank;
       after.version++;
       aimee_pg_stmt_t *u = aimee_pg_prepare(
           conn,
           "UPDATE entity_edges SET lifecycle_state='invalidated',invalidated_at=?2,"
-          " version=version+1,commit_id=?3 WHERE id=?1",
+          " authority_rank=?4,actor_principal=?5,version=version+1,commit_id=?3 WHERE id=?1",
           err, sizeof(err));
       if (!u)
          return fm_end(conn, 0);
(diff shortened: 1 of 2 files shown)


JBailes:
Closing: this change is **not needed on current testing**, and the failing test is what proved it.

I ported it forward from #2871 on the strength of its commit rationale (a USER retraction leaving the row at the MODEL asserter's rank, so a later SYSTEM re-extraction satisfies the reactivate gate and resurrects the fact). That was true on the branch it was written against. It is no longer true here.

`db2_fact_mutation_invalidate` now writes a rejection tombstone (`fm_tombstone_add_assertion(..., "explicit fact invalidation")`, `fact_mutation.c:1211`), and `db2_fact_mutation_assert` checks `fm_tombstone_blocks` **first** (`:748`), returning `FACT_MUTATION_TOMBSTONED` (-3) before the authority_rank gate is ever consulted. A USER retraction already blocks SYSTEM re-extraction on the retract path. My test failed on `db2_fact_mutation_assert(...) == 0` for exactly that reason: the call correctly returns -3, not 0.

This also explains the asymmetry I mistook for an omission. `testing` carries the **rollback** half of the original change and not the retract half, and that is deliberate: rollback writes no tombstone, as the comment on the existing `rb_reextract_ev` test says, so there the rank gate is the only thing standing between a rolled-back insertion and its return. Retract has the tombstone. The two paths need different defences and already have them.

Keeping the change would not be harmless, either. Raising `authority_rank` on invalidate interacts with `fm_tombstone_restore_assertion` on `FACT_REVIEW_UNDO`, which exists to restore a rejected assertion; a raised rank could block the undo it is supposed to permit.

#2874 (code index reconciliation) is unaffected and stands on its own.

PR #2888, opened 2026-08-28
Reduce idle observability bus CPU

JBailes:
## Summary

- exponentially back off the native observability/event-bus consumer from 200 microseconds to a 5 millisecond ceiling when fully idle
- reset the backoff immediately when either observability rows or other routed module traffic is detected
- preserve burst throughput, capture flushing, shutdown draining, and deadline behavior

## Why

Post-merge exploratory testing of the 0.4.0 testing-tag stack found the native observability consumer polling every 200 microseconds even with no traffic. The same loop runs in both the server and KB processes and remained a measurable component of idle CPU after the Go module polling fix.

## Verification

- all 75 native lint gates
- full native unit registry: all tests passed
- full native unit registry under ASan/UBSan with leak detection: all tests passed
- focused bus audit durability, shutdown-race, and module-concurrency tests
- exact merged testing baseline ec42178ca3: healthy split stack, 100/100 concurrent KB status calls, cleartext KB port refused, no-cert mTLS rejected

Live exact-head CPU and latency validation is continuing in CT 112 and will be added to this PR.

### DIFF
--- a/src/modules/audit/obs_bus.c
+++ b/src/modules/audit/obs_bus.c
@@ -877,7 +877,9 @@ static void *consumer_main(void *arg)
    /* Nap only when idle. During a burst the consumer must keep pace with the
     * producer or the ring backs up and the producer is forced to wait, so it
     * loops without napping as long as it is moving rows. */
-   const struct timespec nap = {.tv_sec = 0, .tv_nsec = 200 * 1000}; /* 200 us */
+   const long nap_min_ns = 200 * 1000;      /* 200 us */
+   const long nap_max_ns = 5 * 1000 * 1000; /* 5 ms */
+   long nap_ns = nap_min_ns;
    while (!atomic_load_explicit(&g.stop, memory_order_acquire))
    {
       uint64_t now = bus_runtime_monotonic_ns();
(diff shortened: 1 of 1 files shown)


JBailes:
Live release-candidate validation against the merged testing images plus exact PR commit 5424b53b39:

- Built aimee-server and aimee-kb at the exact PR commit inside the published Debian 12 testing runtime (glibc 2.36), then layered only those binaries over the latest testing images.
- Persistent services healthy, zero restarts; identity bootstrap exited 0.
- Readiness, KB status, rules, and HTTPS login all returned 200.
- 100 concurrent KB status requests: 100/100 HTTP 200.
- KB cleartext port refused; mTLS endpoint without a client certificate returned 401; KB publishes no host ports.
- Fresh memory canary passed store, list, search, destructive delete, and empty post-delete search.
- 12 status calls after one-second idle intervals: min 11.766 ms, mean 12.205 ms, max 13.555 ms.
- Idle aggregate CPU, 5 samples: server 20.81-22.71% (mean 21.55%) versus merged testing baseline 23.41-24.09%; KB 7.96-8.67% in 4/5 samples versus baseline 10.89-11.46%, with one periodic 13.05% spike.
- Five-second per-process interval: native server 1.8% and native KB 0.6%; prior stock attribution was approximately 3.0% and 3.6% respectively.
- No crash, allocator-corruption, sanitizer, panic, deadlock, dropped-event, or OOM signatures in live logs.

Local verification also passed lint (75 gates), full clean unit tests, ASAN/UBSAN, bus audit durability, shutdown-race accounting (27,502 written, zero dropped), and obs-bus concurrency (400 calls, 8 callers). The superseding CI run has passed all completed jobs including ASAN/UBSAN; static analysis is the final job still running at time of comment.

PR #2892, opened 2026-08-28
fix(memory): treat scope all as complete context

JBailes:
## Summary

- treat an explicit `scope: all` memory search as complete caller context
- avoid reporting `active_context_missing: true` when the real failure is KB transport availability
- document the existing memory-search scope/context request fields in OpenAPI
- add integration coverage for the response envelope

## Release-testing reproduction

On the current `:testing` images, a `memory.search` request with `scope: all` and no cwd/project/workspace still returned `active_context_missing: true`. During a deliberate KB outage that caused the response message to blame missing project context instead of the unavailable knowledge service, even though all-scope was explicitly requested.

## Validation

- `make -C src lint` — all 76 checks passed
- `make -C src integration-tests` — 70/70 passed
- `git diff --check`

The local integration run did not have a KB service, so its KB-backed block was skipped; the added assertion is exercised by CI environments that provision the full stack.


JBailes:
Updated onto current testing at b7b821bcf2 and resolved the OpenAPI conflict by retaining both the new search bounds and the scope/cwd/project/workspace fields.

Fresh validation on the merged branch:
- make lint: all 76 checks passed
- integration harness: 83/83 passed (69 expected service skips)
- thin-client-remote-exclusive: passed after removing a clean stale test worktree from an earlier local run
- combined live candidate: scope=all returned HTTP 200 with active_context_missing=false at limits 1 and 32; invalid keyword/limit bounds returned typed physical HTTP 400

PR #2899, opened 2026-08-29
fix(kb): use explicit identity for shared DB probe

JBailes:
## Finding

A clean managed Compose bootstrap runs the `aimee-server-identity` helper as root against the KB shared volume. Its `pg_isready` probe inherited libpq defaults, so embedded PostgreSQL logged `FATAL: role "root" does not exist` during every otherwise-successful enrollment.

## Fix

Probe the known maintenance database as the image-owned `aimee` role. Extend the shared-volume entrypoint test to assert both explicit arguments while preserving the one-shot/no-duplicate-modules path.

## Verification

- Reproduced against immutable `:testing` (`c4fb95f`) in a clean Debian 13 CT
- Focused shared-volume entrypoint test passes
- `bash -n` passes for entrypoint and test
- `git diff --check` passes


JBailes:

PR #2901, opened 2026-08-29
fix(ci): derive testing image inputs from build graph

JBailes:
## Summary

- derive C-source image ownership from the transitive Make link graph
- derive checked-in header ownership from shipping sources and their include closure
- keep unknown and unresolved inputs fail-closed while reusing images that do not consume a change
- retain explicit Dockerfile/container and API ownership contracts

## Release incidents

The `testing` publication for `c4fb95f` changed only server-owned inputs but rebuilt all six images, including Bekko and Nomic. More directly, `f3699ab` changed only `src/cli_argspec.c` plus its unit test; that C source is thin-client-only, yet the old planner rebuilt server, authority, weightless KB, Bekko, and Nomic.

The graph-derived planner classifies that exact `f3699ab` change as no runtime-image rebuilds. It classifies the live fixes precisely as well: PR #2906 rebuilds only the three KB variants, while PR #2907 rebuilds only `aimee-server`.

## Validation

- `python3 scripts/tests/test_publish_testing_plan.py`
- `python3 -m py_compile scripts/publish_testing_plan.py scripts/tests/test_publish_testing_plan.py`
- `git diff --check`
- replayed the `7774926..f3699ab` publication diff: all six runtime images reuse
- replayed PR #2906: KB variants only
- replayed PR #2907: server only
- independent expanded-Make check confirms every direct KB-linked `src/server/*.c` leaf is classified for KB

### DIFF
--- a/scripts/publish_testing_plan.py
+++ b/scripts/publish_testing_plan.py
@@ -13,7 +13,11 @@ That makes adding a new build input expensive until it is classified, never sile
 from __future__ import annotations
 
 import argparse
+import functools
 import json
+import pathlib
+import re
+import subprocess
 import sys
 from dataclasses import dataclass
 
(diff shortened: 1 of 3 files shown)


JBailes:
Latest live evidence from testing publish run 33242219411: PR #2906 changed shared KB search code, so the three KB variants correctly rebuild, but the current planner also rebuilds aimee-server and aimee-authority-bootstrap unnecessarily. This branch is now merged with testing at 1a2d14f; its replay classifies the same change as the three KB variants only. The 26-case planner regression suite and py_compile pass.

PR #2904, opened 2026-08-29
fix(http): reject trailing JSON request data

JBailes:
## Finding

The immutable `:testing` build at `c4fb95f` accepted a valid JSON prefix followed by arbitrary data. For example, `POST /v1/memory/store` with `{"key":"x","content":"y"} {}` returned 200 and persisted the first object instead of rejecting the malformed request. The same behavior affected generic sync routes, async queued operations, bespoke control routes, completions, and streaming chat ingress.

## Fix

Enforce one complete JSON document at the HTTP routing boundary, allowing only JSON whitespace after the value. Validate streaming/mirrored completion ingress before execution or shadow copying. Empty bodies remain supported for no-argument routes.

## Verification

- Target HTTP unit suite passes under `-Werror`
- Full `make -C src build-integrity` passes
- Full production server image builds successfully
- Candidate deployed into the clean Debian 13 Compose E2E environment
- 46/46 API regression/validation checks pass
- Live malformed-body checks pass across memory store, session create, skills, async embed, completions, chat stream, workflow, and trigger paths
- Candidate remains healthy with zero restarts/OOM
- `git diff --check` passes


JBailes:

PR #2905, opened 2026-08-29
fix(memory): stop clipping stored content at 2047 bytes

JBailes:
## What breaks today

`aimee memory store` silently loses everything past 2047 bytes. Found by exploratory testing against a live `:testing` managed stack; confirmed in DB2:

```
 id |  key  | content_len
 29 | w5000 |        2047
 28 | w3000 |        2047
 27 | w2048 |        2047
 26 | w2000 |        2000
```

The command returns success and logs nothing, so an operator storing a long note has no signal that most of it is gone.

## Cause

Two fixed buffers in `memory_insert_epistemic_ex`.

**`char safe_content[2048]`** is the writable copy the content scanner needs (`memory_scan_content` redacts in place), and that copy is what gets stored, so the `snprintf` into it *was* the truncation. It also meant the scanner classified a long note from its first 2047 bytes and stored a sensitivity label describing less than the caller asked us to keep. That is not a leak, since the bytes it never read were not stored either, but the label is wrong. The copy now grows to fit, thread-local so the function's many return paths need no cleanup.

**`char preserved_content[2048]`** holds the existing row's content on the exact-key merge path, and can be handed straight back to `db2_memory_merge_update_ex` as `merged_content`. A long memory that was merely re-stored under the same key was therefore rewritten shortened. `db2_memory_lookup_merge_fields` now allocates the whole value rather than copying into a caller buffer; it has one production caller and no test caller, so the signature change is contained.

## Test

`test_long_content_survives_store_and_merge` reads `length(content)` out of DB2 rather than through `memory_t`, because `memory_t.content` is itself a fixed `char[2048]`: reading back through the struct caps at 2047 whatever the row holds, and would hide the defect under test. It covers the store and merge paths at 2047, 2048, 4096 and 40000 bytes, so a future fixed buffer of any size fails rather than moving the cliff.

Verified against real PostgreSQL: the test passes on this branch and fails on the same tree with the two fixes reverted (`Assertion (size_t)stored_content_len(key) == n failed`).

## Deliberately not fixed here

`memory_t.content[2048]` is a real read-side cap, and it is why `aimee memory get` prints 2047 bytes of a memory that `aimee memory search` returns in full (measured live: 2053 vs the row's 20000). Widening that struct reaches every reader in the tree, so it does not belong in a truncation bugfix. Flagging it rather than folding it in.

## Live verification

Built the patched KB, deployed it to the managed stack, and re-ran the store:

```
fx20000=20000
fx5000=5000
fx2048=2048
fx2047=2047
```

Lengths still exact after a full `docker restart` of the KB and server, with `kb status` reporting ok.

## Baselines

`tests/baselines/db2/declarations-v1.json` is regenerated for the changed signature; the line-number churn is the 6 added header-comment lines. `source-boundary-v2.json` and the refactor surface were already drifting on `testing` before this branch, so they are left alone here.


JBailes:
Pushed `d8b06f5`: the sanitizer lane caught a third bug, and it is a real one.

The new long-content test is the first test in the tree to store a memory over about a kilobyte, and UBSan immediately reported:

```
modules/memory/memory_core_scope_embed.c:292:13: runtime error: index 1024 out of bounds for type 'char [1024]'
```

`memory_auto_tag_workspace` lowercases key+content into a fixed `char lower_buf[1024]`. Only the copy loop respected the bound; the separator append and the terminator did not. The inner loop stops at `li == 1023`, the unguarded `lower_buf[li++] = ' '` takes `li` to 1024, and `lower_buf[li] = '\\0'` then writes one byte past the end of the array.

This is pre-existing and live in the shipped build, not something this branch introduced: any key and content that together exceed the buffer reach it, and 1024 is well below the old 2047 clip. It had simply never been exercised, because nothing stored content that long.

Fixed by keeping `li` strictly below the last index and appending the separator only when it still fits. Verified by running the long-content test under `-fsanitize=address,undefined`: clean, no diagnostics.

I checked the three other `[j++] = ' '`-shaped sites in the tree (`css_oracle.c`, `css_render_oracle.c`, `web_search.c`); all bound every write correctly, so this was the only instance.

JBailes:
Exploratory follow-up found a separate read-side cap after this PR preserves long writes: `memory get` still routed content through `memory_t.content[2048]`. Fixed losslessly in stacked PR #2910 with 5,000-byte DB2/client regressions and ASan/UBSan coverage. Keeping it separate avoids broadening this write-side fix.

JBailes:
Live release evidence from published `testing-1a2d14f` on `.252`: a 5,000-byte `/v1/actions/memory.store` followed by `memory.get` returned 2,047 bytes. #2905 fixes the write/scan side; stacked #2910 preserves the full value through the read/presentation path.

JBailes:
Heads-up on the red `Static analysis and secret scanning` check here: it is not from this branch. `testing` itself is failing that gate (run 33242219418, from #2906), and every open PR inherits it through the merge commit:

```
nullPointerRedundantCheck:kb/http/kb_http_search.c count=1 baseline=0
```

Filed #2913 to fix it. This PR should go green once that lands and this branch picks it up.

The sanitizer lane, which was the real failure on the previous push, now passes with the buffer-overflow fix in `d8b06f5`.

JBailes:
Rebased onto current `testing` and re-verified: `make lint` passes all 76 checks locally, including `db2-declaration-ledger-check` (the ledger needed regenerating because #2910 added `kb_service_backend_memory.c` as a consumer of several DB2 memory declarations).

**This needs someone with merge rights to click merge.** My token can create PRs and push branches but cannot call the merge API — both `mergePullRequest` and `PUT /pulls/:n/merge` return 403, so the account's admin rights are not reachable through it.

I tried the direct-push equivalent (a `--no-ff` merge commit pushed to `testing`, which is how #2913 landed). That does not work here: `testing` is protected and requires status checks on the exact commit being pushed, so a hand-made merge commit needs its own full CI run. I got one green twice, but `testing` takes a merge every 10-25 minutes and CI needs ~25, so the commit went stale before it could be pushed both times. #2913 only succeeded because it happened to land in a quiet gap.

Merging this PR through the UI avoids the race entirely, since GitHub does it server-side and the required checks are satisfied by this PR's own run.

PR #2907, opened 2026-08-29
fix(skill): route executable eval through server

JBailes:
## Summary
- expose `skill eval-fixtures` as the documented alias for fixture evaluation
- route `skill eval-exec` from the shipped thin client through the server
- gate executable evaluation with `CAP_DELEGATE` because it spends provider budget
- publish the command through the served argspec for older clients

## Release blocker
The published `aimee-testing-7774926` artifact reports `skill eval-exec` as an unknown subcommand even though the command and release documentation ship in the same commit. The implementation existed only in the monolithic local CLI and had no server dispatch, HTTP route, thin-client route, or served argspec entry.

## Verification
- `make -C src -j4 build/obj/tests/unit-test-cli-argspec build/obj/tests/unit-test-cli-v1-delegate build/obj/tests/unit-test-server-dispatch build/obj/tests/unit-test-server-http ../aimee-server`
- `src/build/obj/tests/unit-test-cli-argspec` (127 specs, 1250 differential samples)
- `src/build/obj/tests/unit-test-cli-v1-delegate`
- `src/build/obj/tests/unit-test-server-dispatch`
- `src/build/obj/tests/unit-test-server-http`
- `python3 scripts/check_argspec_numeric_parity.py`
- `git diff --check`


JBailes:
Superseded by #2908, which carries the same shipped-client routing fix plus the complete server implementation, candidate-image validation, and route documentation. Keeping one release-blocker PR avoids competing implementations.

PR #2908, opened 2026-08-29
fix(skill): route eval commands through shipping API

JBailes:
## Summary
- route `skill eval`, `eval-fixtures`, and `eval-exec` through the shipping thin-client/server API surfaces
- add the executable-evaluation server handler, bounded numeric validation, manifest/HTTP exposure, CLI marshalling, rendering, and verdict exit codes
- remove the unreachable implementation from legacy `cmd_skill.c`, which is linked into no shipping binary

## Published reproduction
On the published `:testing` image from `7774926`, `aimee help --all` omitted the new commands and both `aimee skill eval-fixtures ...` and `aimee skill eval-exec ...` failed as unknown commands, even though the source feature existed. The implementation had landed only in the legacy command layer.

## Validation
- focused CLI routing/marshalling/printer tests pass
- focused server dispatch tests pass
- full `build-integrity` passes after rebasing onto current `testing`
- shipping `aimee` and `aimee-server` binaries build
- production-candidate Docker image built and was exercised on isolated CT121
- live candidate help and manifest expose all three names; aliases reach their handlers
- malformed type, non-finite/fractional/out-of-range, and cost-bound requests are rejected before any model call
- valid requests safely reach agent configuration; no billable provider call was made during validation


JBailes:

PR #2910, opened 2026-08-29
fix(memory): preserve long content on get

JBailes:
## Summary

- return full stored memory content from the KB read-by-id JSON path without widening the bounded `memory_t` ABI
- preserve that lossless JSON object through the KB client and server `memory.get` response
- keep the direct content query under the identical active-memory scope predicate

## Regression coverage

- 5,000-byte DB2 value is returned byte-for-byte by `db2_kb_service_memory_get_json`
- 5,000-byte KB response remains intact through `kb_client_memory_get_json_as_of`
- focused normal and ASan/UBSan KB/client suites pass
- `aimee-server` builds and `aimee-home-check` passes

## Dependency

Stacked on #2905, which fixes the write-side truncation and long-content auto-tag scan. This PR fixes the independently discovered read-side truncation. Retarget to `testing` after #2905 merges.


JBailes:
Live reproduction on published `testing-1a2d14f` via the owned `.252` environment: stored a 5,000-byte memory through `/v1/actions/memory.store`, then `/v1/actions/memory.get` returned exactly 2,047 bytes. This independently confirms the read-path cap fixed here; #2905 fixes the write/scan side. The isolated test environment will be destroyed after validation.

PR #2913, opened 2026-08-29
fix(kb): validate search "scope" before reading it

JBailes:
## `testing` is red and this unblocks it

The static-analysis gate is failing on `testing` itself (run 33242219418, from #2906), so every open PR inherits the failure through its merge commit:

```
cppcheck-ratchet: new or increased diagnostics
  nullPointerRedundantCheck:kb/http/kb_http_search.c count=1 baseline=0
    kb/http/kb_http_search.c:37: Either the condition 'js' is redundant
    or there is possible null pointer dereference: js.
```

## Cause

`kb_http_search_project_scope` read the value before establishing the pointer was non-NULL:

```c
const cJSON *js = cJSON_GetObjectItemCaseSensitive(root, "scope");
const char *scope = cJSON_IsString(js) ? js->valuestring : "current";
if (js && !cJSON_IsString(js))   /* only now is js tested */
```

`cJSON_IsString` returns false for NULL, so this was safe in practice. cppcheck cannot see inside it, and the ordering is genuinely harder to follow regardless: the reader meets a dereference first and the null case several lines later.

## Fix

Settle the shape first, in the order it actually happens. After the guard `js` is either NULL or a string, so the default needs no type test.

Behaviour is unchanged in all three cases: absent stays `"current"`, a valid keyword passes through, and a present non-string is still a 400.

## Why you can believe that

`test_search_scope_absent_valid_and_wrong_type` pins the three shapes, including two that are easy to get wrong:

- a JSON `null` is a real cJSON item rather than a missing key, so it is rejected as malformed rather than defaulted;
- with no verified credential and no project, `"current"` resolves to no active project and the route returns 409 `scope_required` — that is the success-shaped path, proving the value parsed as `"current"` rather than being rejected.

**The test passes against both the old and the new ordering.** That is what makes "behaviour-preserving" a checked claim rather than an assertion.

I could not reproduce the diagnostic with standalone cppcheck using the ratchet's own flags (it needs the whole-source-set analysis CI runs), so CI is the authority on the ratchet clearing. The reorder removes the exact pattern named: there is no longer a read of `js->valuestring` ahead of the `js` null test.

### DIFF
--- a/src/kb/http/kb_http_search.c
+++ b/src/kb/http/kb_http_search.c
@@ -34,12 +34,18 @@ int kb_http_search_project_scope(const char *body, char *project, size_t project
       return kbhs_error(out_buf, out_cap, 400, "invalid json");
    const cJSON *jp = cJSON_GetObjectItemCaseSensitive(root, "project");
    const cJSON *js = cJSON_GetObjectItemCaseSensitive(root, "scope");
-   const char *scope = cJSON_IsString(js) ? js->valuestring : "current";
+   /* Settle the shape of "scope" before reading it: absent defaults to
+    * "current", present-but-not-a-string is a client error. Reading
+    * js->valuestring first and only then testing `js` for NULL read as a
+    * null-deref to cppcheck (nullPointerRedundantCheck), which is what the
+    * ratchet caught -- and it is genuinely harder to follow. After this guard
+    * js is either NULL or a string, so the default below needs no type test. */
    if (js && !cJSON_IsString(js))
    {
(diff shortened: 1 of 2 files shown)


JBailes:
Independent parity validation on the owned `.252` environment passed with Ubuntu 24.04 and the pinned `Cppcheck 2.13.0`: targeted `kb_http_search.c` analysis reports no `nullPointerRedundantCheck`. Consolidating here because this PR also adds explicit absent/string/wrong-type route regressions.

PR #2914, opened 2026-08-29
fix(kb): validate search scope before dereference

JBailes:
## What

Validate an optional search scope value before dereferencing it, then select either the validated string or the existing current-scope default. Runtime behavior is unchanged.

## Why

The testing head introduced a cppcheck nullPointerRedundantCheck diagnostic in kb_http_search.c. The static-analysis ratchet fails every PR rebased onto testing until the base warning is resolved.

## Verification

- Exact pinned Cppcheck 2.13.0 ratchet on kb/http/kb_http_search.c: 0 known diagnostics, no increase
- KB shipping object compiled with warnings as errors
- clang-format-19 dry-run passed
- git diff --check passed

Reproduced from failed CI job 99075530556 after testing commit 1a2d14ff3cb6640db22e0b876ffbd61d7cb656fd.


JBailes:
Superseded by merged #2913, which applies the same validation-before-dereference fix and includes the route regression coverage. Closing this duplicate so the release queue has one authoritative fix.

PR #2918, opened 2026-08-29
fix(server): classify read route errors

JBailes:
## Summary

- classify missing agent configuration as `not_found` and malformed configuration as `unavailable`
- classify agent/provider/catalog input failures as `invalid_argument`, absent resources as `not_found`, and provider fetch failures as `unavailable`
- preserve those typed errors through the command-dispatch HTTP bridge instead of emitting retryable 502 Bad Gateway responses
- add focused handler and full integration regression coverage

## Release-test discovery

A read-only sweep of all 100 exact GET routes on the published `:testing` container found eight 502 responses. Five were misclassified caller/configuration outcomes (`agent.list`, `model.list`, `model.local`, `provider.models`, and `catalog.show`); the dependency-backed sandbox and agents routes remain unchanged.

## Validation

- `make -C src all -j2`
- `src/build/obj/tests/unit-test-agent-list-handler`
- integration harness: 88/88 passed (69 topology-dependent skips)
- thin-client remote-exclusive post-check: passed in an isolated clone
- `make -C src aimee-home-check`
- `clang-format-19 --dry-run --Werror` on changed C files
- `bash -n src/tests/test_integration.sh`
- `git diff --check`

### DIFF
--- a/src/server/server_agent.c
+++ b/src/server/server_agent.c
@@ -560,10 +560,11 @@ int handle_agent_list(server_ctx_t *ctx, server_conn_t *conn, cJSON *req)
        * (unreadable, malformed, truncated) is a real fault and names the file. */
       const char *path = agent_config_path();
       if (path && access(path, F_OK) != 0)
-         return server_send_error(conn,
-                                  "no agents are configured yet: choose a provider in the setup "
-                                  "wizard, or run `aimee provider list --available`",
-                                  NULL);
+         return server_send_error_kind(
+             conn, SERVER_ERR_NOT_FOUND,
+             "no agents are configured yet: choose a provider in the setup "
+             "wizard, or run `aimee provider list --available`",
+             NULL);
       /* The file is there and the load still failed: unreadable, malformed, or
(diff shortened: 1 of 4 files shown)


JBailes:
Live CT121 candidate validation passed against a full compose topology:

- `/v1/model/local` and `/v1/agent/local` missing endpoint: HTTP 400, `kind=invalid_argument`
- `/v1/provider/models` missing name: HTTP 400, `kind=invalid_argument`
- `/v1/provider/models` unknown provider: HTTP 404, `kind=not_found`
- `/v1/catalog/show` missing model: HTTP 400, `kind=invalid_argument`
- `/v1/agent/list` and `/v1/model/list` with absent agent config: HTTP 404, `kind=not_found`
- negative controls remained distinct: sandbox Docker backend HTTP 502, agents backend HTTP 502, deploy-disabled HTTP 503
- `/v1/ready`: all dependencies `ok`; container healthy with 0 restarts

Candidate image was built from `4e63bc04bb` and deployed only for this test. The environment will now be restored to the official published `:testing` image.

PR #2921, opened 2026-08-29
fix(http): require Host in HTTP/1.1 requests

JBailes:
## Summary

- require exactly one non-empty `Host` header in every data-plane HTTP/1.1 request
- reject missing, empty, and duplicate Host fields before authentication or dispatch
- add raw-framing regression coverage

## Validation

- `make -C src all -j2`
- `make -C src lint` (all 76 checks)
- `src/build/obj/tests/unit-test-server-http`
- integration harness: 88/88 passed (service-dependent checks skipped locally and covered by full-stack E2E)
- isolated `test-thin-client-remote-exclusive.py`: pass

Live CT121 raw-protocol candidate evidence will be added as a PR comment.


JBailes:
Live CT121 candidate validation on commit `530d3f2ba98758d9a31467b20ca23a3edad0781a`:

- built `Dockerfile.server` successfully and deployed only the server service
- raw UDS baseline with Host: 200
- missing Host: 400
- empty Host: 400
- whitespace-only Host: 400
- duplicate Host: 400
- case-variant duplicate (`Host` + `hOsT`): 400
- existing duplicate Content-Length, TE+CL, and pipelining controls remain 400
- native TLS path: missing Host -> 400 before auth; valid Host without credentials -> expected 401
- supported concurrency: 100/100 valid Host requests returned 200; 100/100 missing Host requests returned 400
- readiness remained green; persistence sentinel resolves; server/KB/PostgreSQL healthy with zero restarts
- no panic, segfault, sanitizer, or fatal log entries

Restoring the exact published `:testing` server image now.

PR #2922, opened 2026-08-29
Merge #2905 (memory truncation fixes) into testing

JBailes:
Integration branch carrying the exact merge commit for #2905, so branch protection sees status checks on the SHA that lands on `testing`.

#2905 is green on its own (34/34) but the merge commit created for it is a new SHA with no checks of its own, and `testing` requires checks on the commit being pushed. This PR runs them.

Contents are exactly #2905 merged into current `testing`: the store/merge truncation fix and its regression test, plus the auto-tag and read-side fixes that landed on that branch from #2910 and the auto-tag scan commit, and the regenerated DB2 declaration ledger.

`make lint` passes all 76 checks locally on this merge.


JBailes:
Closing: this was scaffolding to get status checks onto a hand-made merge commit, because my token cannot call the merge API. It went green twice, but `testing` takes a merge every 10-25 minutes and CI needs ~25, so the merge commit is stale before it can be pushed. #2905 itself is green and mergeable and should be merged normally.

PR #2925, opened 2026-08-29
Fix KB index failures crossing HTTP as success

JBailes:
## Summary
- classify typed KB index proxy failures with authoritative server error kinds
- map authorization refusals to 403 and dependency failures to 503 while preserving KB diagnostics
- regress index list against unauthorized and unavailable results

## Validation
- `make -C src build/obj/tests/unit-test-index-investigate-status -j2`
- `src/build/obj/tests/unit-test-index-investigate-status`
- `make -C src all -j2`
- `make -C src lint` (all 76 checks)


JBailes:
Live fault-injection proof on disposable CT122 using amended commit `ad44544c`, built inside the official `ghcr.io/rakuensoftware/aimee-server:testing` Debian baseline and bind-mounted into the otherwise official stack:

- Deleted only the disposable root KB membership: raw `/v1/index/list` returned HTTP 403 with `status=unauthorized`, `kind=permission_denied`, `http_status=403`.
- Official `:testing` thin client returned the same typed JSON and exit 1.
- Stopped the KB container: raw route returned HTTP 503 with `status=unavailable`, `kind=unavailable`, `http_status=503`, plus retry metadata.
- Official thin client returned the same unavailable JSON and exit 1.
- Restarted KB: both KB and server returned healthy with zero restarts.

The live test initially caught that an absent optional runtime-web classifier left HTTP at 200; the amended commit now uses the authoritative KB result enum as the fallback status source. Focused regression, full build, and all 76 lint checks pass.

JBailes:
CI integration follow-up: the red client/server harness still asserted that `index overview` exits 0 in its deliberately KB-less topology—the behavior this PR corrects. Final head `ea42a5754f` updates that assertion to require exit 1, typed `status=unavailable`, and `http_status=503`. The local client/server integration harness now passes 90/90. The server code at final head is identical to the live-tested `ad44544c`; the only subsequent change is this regression-harness expectation.

PR #2926, opened 2026-08-29
fix(compose): select bundled A25M embedder by default

JBailes:
## What

Pair the default `aimee-kb-a25m` image with `EMBEDDER_MODEL=bekko-a25m` in both root Compose entrypoints, and enforce that invariant for every shipped Compose file.

## Release-blocking reproduction

On isolated release CT121, recreating the official split stack from published commit `6b8bbdc2` and `ghcr.io/rakuensoftware/aimee-kb-a25m:testing@sha256:0df2c059...` with no ambient embedder variable produced a restart loop:

```
aimee-kb: no embedder selected, and there is no fallback. Retrieval needs one.
aimee-kb: then re-run Deploy. Refusing to start.
```

The default image already carries A25M weights; the entrypoint deliberately requires an explicit vector-space selection. The shipped root Compose files selected the image but passed an empty model, unlike `deploy/compose/aimee.yaml`, which already keeps the pair together.

## Verification

- `python3 scripts/check-kb-container-packaging.py --root .`
- `python3 scripts/check-kb-container-packaging.py --root . --plant-test`
- Live official-image recreation proof will be posted from CT121.

### DIFF
--- a/compose.server.yaml
+++ b/compose.server.yaml
@@ -45,7 +45,10 @@ services:
       # MUST receive both: `aimee config deploy-env` emits them, and without them
       # reaching the container the entrypoint sees no selection, starts nothing, and
       # the builtin serves forever — the wizard's embedder silently never runs.
-      EMBEDDER_MODEL: ${EMBEDDER_MODEL:-}
+      # Keep the selection paired with the default bundled-image variant. The
+      # entrypoint intentionally refuses to guess a vector space, so an empty
+      # value makes the default image restart forever despite carrying A25M.
+      EMBEDDER_MODEL: ${EMBEDDER_MODEL:-bekko-a25m}
       EMBEDDER_URL: ${EMBEDDER_URL:-}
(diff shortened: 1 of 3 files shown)


JBailes:
Exact live proof on isolated CT121 against the same official artifact that failed before the change (`aimee-kb-a25m:testing@sha256:0df2c059f3a5...`):

- Copied this PR’s `compose.server.yaml` and invoked it with `EMBEDDER_MODEL` and `EMBEDDER_URL` explicitly unset.
- Container resolved `EMBEDDER_MODEL=bekko-a25m`; log: `starting bundled embedder (bekko-a25m) on :8760`.
- KB became healthy with zero restarts; `/v1/health` reported `status=ok`, `db2_ok=true`, `db2_kb_tables_ok=true`, `embed_ok=true`.
- Server became healthy with zero restarts and returned `status=ok`.
- Existing persistent DB rows and webchat database/WAL remained present after recreation.

Before this exact compose change, the same digest with the same volumes restarted 8 times and refused startup with `no embedder selected, and there is no fallback`.

PR #2927, opened 2026-08-29
fix(kb): stop the audit bus consumer pinning a DB2 pool connection

JBailes:
## The leak

The KB permanently loses one of its sixteen DB2 pool connections, from the first audit row it writes until the process exits. The pool's own reaper says so every thirty seconds, with the hold time tracking uptime:

```
db2.pool: member 0 leased 779902ms (> 300000ms ceiling)
  by unattributed — missed lease_end?
```

Reproduced on a clean container running the stock `:testing` image, so it is not workload-specific.

## Cause

`db2_conn()` leases lazily and caches the connection in a thread-local, so a thread that acquires once outside a `db2_lease_begin`/`_end` scope holds a pool member until it exits or calls `db2_lease_release_idle`. Every other long-lived KB worker calls one of those at a job boundary.

The `obs_bus` consumer thread does not, because it has no notion of a job — it pumps, drains, and naps forever. Its durable sink is aimee-kb's WORM append, which reaches `db2_kb_audit_append` and takes the lease on first use.

## Fix

`obs_bus` cannot release it itself: it is storage-neutral by design and the sink is injected per daemon (SQLite in the server, PostgreSQL in the KB). So the bus now exposes the one thing only it knows — when the flow goes idle — and the KB adapter decides what that costs.

The hook fires only on the idle edge, so a burst still runs on a single lease and pays nothing extra. The server adapter needs no hook: its WORM append is SQLite and its guardrail sink is DB1, neither of which uses the DB2 pool.

## How it was found

By instrumenting a live KB, not by reading. A temporary trace of every lazy acquire and every release, keyed by thread, showed **exactly one thread with an unmatched acquire**, and its backtrace named the path:

```
consumer_main → bus_host_pump → governance_tap
  → persist_or_queue_durable → persist_durable → db2_kb_audit_append
```

I had previously guessed a different thread and was wrong; the trace is what disproved it, and I did not ship that guess.

## Verification

On the same container, with the fix:

- **zero** stuck-lease reports after eight minutes, where the same image logged one every thirty seconds from startup;
- the trace shows that thread completing **three acquire/release cycles**, so it releases and then re-acquires correctly for later rows rather than losing the connection. Release resolves to `consumer_main` (the new hook), acquire to `db2_kb_audit_append`.

`make lint` passes all 76 checks. The DB2 declaration ledger is regenerated for the one new consumer entry this adds.

## Test

`test_bus_audit_durability` now asserts the invariant that prevents the leak: once the backlog drains, the consumer reaches its idle edge and keeps telling the sink so. It also pins that reconfiguration while running is refused, matching the sinks it accompanies.

It **fails with the hook call removed** (`Assertion 'first_idle > 0' failed`), and the file's existing durability invariants still hold alongside it (5000 rows, each exactly once, zero drops).

### DIFF
--- a/src/kb/kb_ingest_workers.c
+++ b/src/kb/kb_ingest_workers.c
@@ -347,6 +347,11 @@ static void *kbiw_timer_thread(void *arg)
       if (interval_secs <= 0)
          interval_secs = 6 * 3600;
 
+      /* Same reason as the watch thread, over a much longer wait: the enqueue
+       * above leases lazily, and this loop then sleeps out the whole ingest
+       * interval (six hours by default) with nothing to end the unit of work. */
+      db2_lease_release_idle();
+
       int slept = 0;
       while (slept < interval_secs)
       {
@@ -431,6 +436,13 @@ static void *kbiw_watch_thread(void *arg)
    char evbuf[4096] __attribute__((aligned(__alignof__(struct inotify_event))));
    while (!ctx->ingest_stop)
    {
+      /* Let go before waiting for the next filesystem event. The enqueue below
+       * leases a pooled DB2 connection lazily and this thread has no unit of
+       * work to end, so without this it kept the connection across an idle
+       * watch -- which on a quiet tree is the whole life of the process, and
+       * the pool reaper reports it as a stuck lease. Releasing here costs one
+       * re-acquire per burst of edits, not per event. */
+      db2_lease_release_idle();
       struct pollfd pfd = {.fd = ifd, .events = POLLIN};
       if (poll(&pfd, 1, 1000) <= 0)
          continue;
--- a/src/kb/kb_obs_bus_adapter.c
+++ b/src/kb/kb_obs_bus_adapter.c
@@ -4,6 +4,7 @@
 #include <aimee/audit/obs_bus.h>
 
 #include "modules/db2/c/kb_audit_worm.h"
+#include "modules/db2/c/db2.h" /* db2_lease_release_idle */
 
 static int persist_durable(const char *actor_role, const char *actor_principal, const char *action,
                            const char *subject, const char *verdict, const char *detail, void *ctx)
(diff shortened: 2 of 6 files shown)


JBailes:
Pushed `b7d91fd`: two more instances of the same defect, found by auditing every long-lived KB thread against the shape this PR's first commit identified.

**`kbiw_watch_thread`** enqueues an ingest job when inotify reports an edit (which leases), then goes back to polling for the next event. On a tree nobody is editing, that wait is the rest of the process's life.

**`kbiw_timer_thread`** is the same over a longer wait: it enqueues on startup and each interval, then sleeps out the whole ingest interval, six hours by default, still holding the connection.

Both now release where they start waiting, matching every other long-lived KB worker.

**On verification, to be precise about what is and is not proven:** the bus consumer fix is verified end to end (zero stuck-lease reports over eight minutes where the same image logged one every thirty seconds, plus a trace showing three clean acquire/release cycles). These two are fixed on the reading. Neither path fires with background ingest off, which is the default, and enabling it needs configuration the borrowed test container would not take. The remedy is the identical one-liner at the same kind of boundary, so I am confident in it, but I am not claiming the same evidence.

The full stack runs healthy for nine minutes with all three fixes and zero stuck-lease reports, and `make lint` passes all 76 checks.

I also audited the paths I did **not** change: the four ingest workers use balanced `db2_lease_begin`/`_end`, the maintenance and blob-reconcile timers release every second, mining is balanced, HTTP connection workers release per connection, the mTLS workers bracket with begin/end, and the curator lanes release each cycle. The server's bus adapter needs no hook at all: its WORM sink is SQLite and its guardrail sink is DB1, neither of which touches the DB2 pool.

JBailes:
**Correction: the two ingest fixes are now verified, and the leak was worse than this PR first described.**

I previously said these two were fixed on the reading because I could not enable background ingest in a borrowed test container. I built a dedicated one instead (standalone KB, a git workspace with a resolvable project identity, `kb.bg_ingest.enabled: true`) and got a clean A/B on identical config:

| build | stuck-lease reports after 9 min |
| --- | --- |
| stock `:testing` | **24** — pool members 0, 1 and 2 all pinned |
| this branch | **0** |

Three pinned connections, not one: the bus consumer plus both ingest threads, exactly the three this PR fixes. My earlier statement that only the bus consumer leaked was an artifact of testing with background ingest off, which is the default and hides two thirds of it.

Both builds reached the same code path (`kb.ingest.timer: enqueued 1 project(s)`, worker picked it up and completed), so the difference is the fix, not a path that went unexercised.

So the impact is larger than stated: on a deployment with background ingest enabled, the KB permanently loses **three of sixteen** pool connections, from startup, until restart.

